PR #7, opened 2018-05-24
Typed config

srid:
- [x] Add `obelisk-executable-config-core`
- [x] obelisk-run: make it use the above API
- [x] obelisk-command:
  - Add `ob deploy verify` to verify config (route only)
  - Make `ob deploy push` verify before pushing
- [x] Undo using of common/route for determining backend port when deploying
- [x] Modify `lookup` and `inject` to use `ObeliskConfig` typeclass.


srid:
I expanded on the prototype, and integrated the changes with the rest of the obelisk. @ryantrinkle I'll stop at this point. Please review again when you get a chance (checkout the exports of the `Types` module), and let me know if I should be working on this further.

srid:
> ... let me know if I should be working on this further.

Next steps I envison are: 
- [ ] Modify `lookup` and `inject` to use `ObeliskConfig` typeclass.
- [ ] Add constraints, eg: executable-config: Ensure contents from config/backend are not injectable
- [ ] Add example use case to the skeleton

srid:
I actually ended up porting `inject` and `lookup` (including on mobile) to use the new typeclass. `inject` constraints the kind of config (i.e., not backend) you can inject. Skeleton is updated with an example use. Next, I'll test it on mobile and cleanup.

srid:
@ryantrinkle I have finished the PR. This is now ready for final review.

lib/executable-config/inject/src/Obelisk/ExecutableConfig/Inject.hs
```diff
@@ -9,9 +17,28 @@ import qualified Data.Text.Encoding as T
...
+      text $ encodeConfig cfg
+  where
+    key = "injected-" <> T.pack (getConfigPath @config)
+
+-- | Check if we are allowed to inject the given config into the DOM.
+-- TODO: Can we perform this check at the compile time?
+canInject :: ObeliskConfig config => ConfigLocation config -> Bool
+canInject = \case
+  ConfigLocation CabalProject_Backend _ -> False
+  _ -> True
```
srid: If the user tries to inject a backend config, this would throw an error at runtime. I was wondering if we can make that a compile error (so it gets caught early). @ryantrinkle Is this worth doing? It would, as one solution, involve a new type class `ObeliskInjectableConfig` (derivative of `ObeliskConfig`).

ali-abrar:
It looks like this has drifted quite a bit. I do think we need to handle configs a bit differently - especially now that we've got hydration merged, configs should be something that can be looked-up and used both pre- and post-hydration using the same interface.

Rather than keep this here, I'm planning on replacing this with a new issue and referring to this there. Please reopen if you'd like to resuscitate this.

PR #9, opened 2018-05-24
Add custom routes to backend config

srid:
Example:

```haskell
backend :: IO ()
backend = Ob.backend $ Ob.def
  { Ob._backendConfig_head = fst frontend
  , Ob._backendConfig_routes =
      [ ("/api", someApi)
      ]
  }

someApi :: Snap ()
someApi = do
  writeBS $ BSC.pack commonStuff
```

### DIFF
--- a/skeleton/backend/src/Backend.hs
+++ b/skeleton/backend/src/Backend.hs
@@ -1,10 +1,22 @@
+{-# LANGUAGE OverloadedStrings #-}
 module Backend where
 
-import Common.Api
-import Frontend
+import qualified Data.ByteString.Char8 as BSC
+
 import qualified Obelisk.Backend as Ob
+import Snap (Snap, writeBS)
+
+import Common.Api (commonStuff)
+import Frontend (frontend)
 
 backend :: IO ()
-backend = Ob.backend Ob.def
+backend = Ob.backend $ Ob.def
   { Ob._backendConfig_head = fst frontend
+  , Ob._backendConfig_routes =
+      [ ("/api", someApi)
+      ]
   }
+
+someApi :: Snap ()
+someApi = do
+  writeBS $ BSC.pack commonStuff
--- a/lib/backend/src/Obelisk/Backend.hs
+++ b/lib/backend/src/Obelisk/Backend.hs
@@ -7,13 +7,16 @@ module Obelisk.Backend
   ) where
 
 import Control.Lens
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BSC8
 import Data.Default (Default (..))
+import Data.Semigroup ((<>))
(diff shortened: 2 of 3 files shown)


ali-abrar:
What if the custom handlers have the same route as a builtin handler?

srid:
@ali-abrar Hmm, I think that would end up overriding the builtin one. I've pushed a commit to make that not happen.

3noch:
@srid I'm not sure that's actually better. Either way you end up with a runtime bug that could be hard to track down. Can we think of a way of doing this that doesn't have that problem?

srid:
@3noch Not sure how this would be a problem in practice. Do you know of ways to improve this? I can't see anything in stela (for example) either.

ryantrinkle:
This shouldn't be a breaking change - it should be a new function or a new
field in the existing config.

mightybyte:
It would be really cool if Obelisk's functionality could be expressed as a snaplet because ideally I'd like to be able to build an Obelisk app with other snaplets like the built-in auth snaplet, snaplet,postgresql-simple, etc.

srid:
@ryantrinkle I've pushed 77e4062 that makes it backward compatible. Users would specify their routes in the config; it should look like this:

```haskell
backend :: IO ()
backend = Ob.backend $ Ob.def
  { Ob._backendConfig_head = fst frontend
  , Ob._backendConfig_routes =
      [ ("/api", someApi)
      ]
  }
```

skeleton/backend/src/Backend.hs
```diff
@@ -1,10 +1,22 @@
...
+backend = Ob.backend $ Ob.def
   { Ob._backendConfig_head = fst frontend
+  , Ob._backendConfig_routes =
+      [ ("/api", someApi)
+      ]
   }
+
+someApi :: Snap ()
+someApi = do
+  writeBS $ BSC.pack commonStuff
```
srid: @ryantrinkle if I understand [your comment](https://github.com/obsidiansystems/obelisk/pull/32#issuecomment-393898442) correctly, this will need changing so that the user does not have to directly import/use `Snap` right? If so, I wonder how that would look ...

ali-abrar:
Right now the division of responsibility between the obelisk's backend library, the backend config, and what the user does on their own, is unclear. With this change, the config potentially clashes with the hardcoded routes in the library and the resolution of that clash is bound to infuriate some set of users.

We might want to start by breaking up the functionality as @srid suggested above so that users can easily construct the current backend from components. Instead of overriding things via config, they should be able to just swap pieces out. Even if we go this route, we still need to decide what is hardcoded, what is configurable, and what is detachable/replaceable.

srid:
Now that we have routes based backend handling, this can be closed.

PR #10, opened 2018-05-25
Add devel to improve ob run error messages

tomsmalley:
This is the work of @luigy in https://github.com/obsidiansystems/obelisk/commit/3caa96376b9f9ed2a6180bf8c266a52f45082e85 but based on current master


luigy:
Thanks!

ryantrinkle:
Is there a way we can get this out of the skeleton? We shouldn't let the
skeleton get any more complex than absolutely necessary.

PR #11, opened 2018-05-25
Cleanup, support static <body> content, switch back to mainWidget

tomsmalley:
Based on #10 

- Makes it less likely to serve the wrong static content (or none at all): share static rendering between ob run and the backend.
- Allow static body content
- Switches back to mainWidget: mainWidgetWithHead has problems with external libraries which manipulate the <head> contents, and should be kept as static as possible.
- Pull obelisk-snap into obelisk-backend


lib/backend/src/Obelisk/Backend.hs
```diff
@@ -1,49 +1,139 @@
...
 -- | Start an Obelisk backend
-backend :: BackendConfig -> IO ()
-backend cfg = do
+runObeliskBackend
+  :: ReflexStaticContent -- ^ The static content
+  -> BackendConfig -- ^ Other configuration
+  -> IO ()
+runObeliskBackend static cfg = do
   -- Make output more legible by decreasing the likelihood of output from
   -- multiple threads being interleaved
```
3noch: Totally unrelated. Is this comment even true? I doubt it.

3noch: Actually I can think of how this would be true. Ignore my comment.

lib/frontend/src/Obelisk/Frontend.hs
```diff
@@ -0,0 +1,16 @@
...
+
+module Obelisk.Frontend where
+
+import Data.ByteString (ByteString)
+import Reflex.Dom
+
+-- | Specify a reflex frontend.
+data ObeliskFrontend = ObeliskFrontend
+  { _obeliskFrontend_body :: forall x. Widget x ()
+  -- ^ The content of <body> after the JavaScript is loaded
```
3noch: Do we want to support mounting to a dom element as well?

tomsmalley: I think we should, but I refrained from adding that to this PR

3noch: Ok sounds good.

tomsmalley:
See #32

PR #12, opened 2018-05-25
Add 'shell.nix' files

alexfmpe:


kmicklas:
Nice. I should have done it this way initially.

srid:
@alexfmpe What is the new `ghcid` command for lib/command with this change?

I am using `nix-shell --run "cd lib/command && ghcid -c 'cabal new-repl' -o ghcid-output.txt"` and it seems to be building the documentation and thus taking long.

srid:
... it was building Go from source as well.

Ah, never mind. I just need to explicitly specify `default.nix` to retain the current behaviour, viz.:

```
nix-shell default.nix -A command.env --run "cd lib/command && ghcid -c 'cabal new-repl' -o ghcid-output.txt"
```

ryantrinkle:
Could we make the one in skeleton also refer to ./. instead of the obelisk
impl? That seems more robust to customizations

PR #19, opened 2018-05-25
Simplified Obelisk.CLI API, take 2

srid:
Improves on #8 by dropping the dependency on `unliftio`.


lib/command/src/Obelisk/App.hs
```diff
@@ -1,31 +1,50 @@
...
 newtype ObeliskT m a = ObeliskT
-  { unObeliskT :: ReaderT Obelisk (LoggingT Output m) a
+  { unObeliskT :: ReaderT Obelisk (CliT m) a
   }
-  deriving (Functor, Applicative, Monad, MonadIO, MonadThrow, MonadCatch, MonadMask, MonadReader Obelisk, MonadLog Output)
+  deriving
+    ( Functor, Applicative, Monad, MonadIO, MonadThrow, MonadCatch, MonadMask
+    , HasObelisk, HasCliConfig)
+
+deriving instance Monad m => Cli (ObeliskT m)
```
matthewbauer: @srid On ghc 8.4 I am getting this very weird error on this line:

```
src/Obelisk/App.hs:28:19: error:
    • Illegal instance for a type synonym
      A class instance must be for a class
    • In the stand-alone deriving instance for
        ‘Monad m => Cli (ObeliskT m)’
```

Do you know if there is a way to fix this?

3noch: `Cli` really is a type synonym (https://github.com/obsidiansystems/obelisk/pull/19/files/b9dcdb059d34584a07548e91a035b23fe0296c69#diff-aca605a5857369b608180dbdf6023b13R31). Hm that's probably not the best idea actually. For now you could just inline the type synonym into this instead. I.e. replace `Cli` with `MonadLog ...`.

matthewbauer: Thanks! should be fixed in 6619da042a88cdd97bd9a5cd79c2ac8ad99f0d4e.

srid: @matthewbauer Sorry, I don't have much of a clue here. If @3noch's suggestion works, then you may have to export `MonadLog` and `Output` outside of `obelisk-cliapp` (or, better, derive the instance inside of cliapp) ...

Also there was no particular reason I chose a type synonym (more like a class synonym?) here. @3noch what's the background on type synonyms? Where, or where not, would you say they be used?

3noch: @srid You usually don't want to write instances for a type synonym because it's somewhat confusing. The new instance will exist on the underlying type, not just for things that use the type synonym. So it ends up creating instances for things you're not mentioning. Usually you only want to add instances for `newtypes` or the type that is actually carrying the instance.

PR #21, opened 2018-05-25
Use random port for ob run tests

tomsmalley:


luigy:
Thanks!

ryantrinkle:
At some point we should accomplish the same thing using unshare. See the GC
test in reflex-dom - it's actually quite easy to do.

alexfmpe:
Didn't we have a similar situation with the benchmarks in reflex-dom due to `unshare` not being available on mac os? 
Had to resort to de-hardcoding the port upstream.

PR #30, opened 2018-05-29
Add VM builder support using Docker

3noch:
I would appreciate various testing attempts on both Mac and Linux. You should be able to just run `ob deploy push --enable-vm-builder` on any OS and Obelisk will tell you what to do each step of the way. I've tried to think of useful automated tests we could add for this feature, but it's not easy to write tests that require a VM running and possibly modify root's SSH config.

Note that we do *not* actually modify root's SSH config. We just tell the user how to do it themselves. We could add this automation if desired but @luigy and I thought best not to.

Also, I've been chatting with @cleverca22 about his linux-build-slave generator (https://github.com/cleverca22/not-os/blob/master/linux-build-slave.nix) and he's made some changes such that we could likely drop the need for Docker entirely. However, there are some unknowns that I'd want to hash out first (e.g. performance, reliability, bootstrapping the build [which requires Linux], etc.). Because of this I've made the CLI API abstract enough that we could swap this out underneath without needing to change any user-facing APIs. I designed it in such a way that I think we can move to this model without breaking users who used Docker in the past.

### DIFF
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -153,7 +153,7 @@ deployCommand cfg = hsubparser $ mconcat
         enabledByDefault = _argsConfig_enableVmBuilderByDefault cfg
         enabled = Just RemoteBuilder_ObeliskVM
         flagBase = "vm-builder"
-        flagDesc = "managed Linux virtual machine and allow builds to use it (requires Docker)"
+        flagDesc = "managed Linux virtual machine as a Nix remote builder (requires Docker)"
 
 
 deployInitOpts :: Parser DeployInitOpts
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -23,6 +23,7 @@ import Options.Applicative
 import System.Directory
 import System.Environment
 import System.FilePath
+import qualified System.Info
 import System.IO (hIsTerminalDevice, stdout)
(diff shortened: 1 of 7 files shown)


3noch:
There is still something going on with the actual deploy stage. It seems the ssh agent isn't supplying the key correctly. This might be a problem only on Mac.

3noch:
It looks like we don't even need `ssh-agent`. We can pass everything via `NIX_SSHOPTS`.

3noch:
This was deployed from a mac: http://ec2-54-89-23-75.compute-1.amazonaws.com/

3noch:
@luigy Rebased on `master`.

lib/command/src/Obelisk/Command.hs
```diff
@@ -45,8 +47,12 @@ data Args = Args
   }
   deriving Show
 
-args :: Parser Args
-args = Args <$> noHandoff <*> verbose <*> obCommand
+newtype ArgsConfig = ArgsConfig
+  { _argsConfig_enableVmBuilderByDefault :: Bool
+  }
```
srid: Can you explain the need for this? I wonder how `--enable-vm-builder` is essentially different from other flags (like `-v`).

3noch: This changes the default behavior of the flag parsing. I.e. on Darwin you don't need to pass this flag and it will still be enabled.

PR #32, opened 2018-05-29
Support static body content, make app/backend config clearer

tomsmalley:
This is a part of breaking up #11 
Other main part (not in this PR) is pulling obelisk-snap into obelisk-backend, since there's no clear motivation for having them as separate packages.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -110,7 +110,9 @@ rec {
   inherit reflex-platform;
   inherit (reflex-platform) nixpkgs pinBuildInputs;
   path = reflex-platform.filterGit ./.;
+  backend = ghcObelisk.obelisk-backend;
   command = ghcObelisk.obelisk-command;
+  run = ghcObelisk.obelisk-run;
   shell = pinBuildInputs "obelisk-shell" ([
     command
     pkgs.openssh
--- a/lib/backend/src/Obelisk/Backend.hs
+++ b/lib/backend/src/Obelisk/Backend.hs
@@ -2,11 +2,14 @@
 module Obelisk.Backend
   ( backend
   , BackendConfig (..)
+  , StaticContent (..)
+  , ReflexStaticContent
   -- * Re-exports
   , Default (def)
   ) where
 
 import Control.Lens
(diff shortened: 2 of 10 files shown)


srid:
> Other main part (not in this PR) is pulling obelisk-snap into obelisk-backend, since there's no clear motivation for having them as separate packages.

I think we should discuss about the role of `obelisk-snap` first. Ryan indicated to me that it would be better if `obelisk-backend` did not enforce an requirement of Snap, and that it would be better if we allowed the user to use whatever web framework--snap, yesod, servant, wai, etc.--they want to use. To that end, `obelisk-backend` could simply provide individual functions that can be composed to achieve this. 

Which basically means providing a way to extract,
- `ReflexStaticContent` as ByteStrings
- static assets paths and URLs (`static.assets`, `frontend.jsexe.assets`, ..) 
- http host and port for the server to bind

(there might be more)

... from an Obelisk project, and have the user pass them to their web framework functions/ configuration.

So `obelisk-backend` could simply be a small individual library which the user can glue together on their own. (Or, the likes of `obelisk-snap` can provide the said gluing). 

Essentially `obelisk-snap` (and `obelisk-yesod`, etc.) would depend on `obelisk-backend` instead of it being the other way.

ryantrinkle:
@srid The change you're describing is much larger than I was envisioning.  The goal right now is not to let the user use whatever web framework they want to use (although that would be great); the immediate goal is to hide the choice of web framework so that we can change it in the future.  We don't want most users to need to know or care about the web framework we've chosen.

For right now, we just need to limit the ways that we expose our choice of web framework, while still making it possible for users to do whatever they need to do.

ryantrinkle:
I believe this change is more or less obsoleted by the routes branch; also, I'd like to avoid having static content specified separately from dynamic content - instead, we should distinguish them using combinators like `prerender` and possibly type-level distinctions.

ryantrinkle:
@tomsmalley If I've missed anything, please let me know!

tomsmalley:
@ryantrinkle indeed, this is no longer relevant :+1:

Issue #6, opened 2018-05-24
Ob deploy push fails with exit code 1


SebastianCallh:
After having installed Obelisk on Ubuntu 16.04 by following the instructions in the readme (installing Obelisk in the nix-env, and using hub) and trying to deploy the example app(also following the readme) the command `ob deploy push` fails with the following error message.

    ✔ Running nix-build on ./.obelisk/impl [command]
    RawCommand "nix-build" ["./src","-A","server","--argstr","hostName","ec2-xx-xxxx-xx-xxxx.eu-central-1.compute.amazonaws.com","--no-out-link"] failed with exit code: 1
    ✖ Running nix-build on ./src [server]

The content of the ./src folder in the myapp-deploy have not been edited by hand and consists of the default.nix file

    # DO NOT HAND-EDIT THIS FILE
    import ((import <nixpkgs> {}).fetchFromGitHub (
      let json = builtins.fromJSON (builtins.readFile ./github.json);
      in { inherit (json) owner repo rev sha256;
           private = json.private or false;
         }
    ))

the github.json file

    {
      "owner": "...",
      "repo": "...",
      "branch": "...",
      "private": ...,
      "rev": "...",
      "sha256": "..."
    }

and the .attr-cache folder which is empty.

srid:
@SebastianCallh Could you show the output of `ob deploy push -v`?

SebastianCallh:
Hi @srid! Great idea, apparently it tries to fetch from an url that gives 404.

Sorry about the formatting, the output is below

Starting Obelisk </nix/store/s200msjpipi4jpyv2xkas35dsb101s9m-obelisk-command-0.1/bin/ob> args=Args {_args_noHandOffPassed = False, _args_verbose = True, _args_command = ObCommand_Deploy DeployCommand_Push} noSpinner=False logging-level=Debug
   Creating process: CreateProcess {cmdspec = RawCommand "nix-build" ["./.obelisk/impl","-A","command","--no-out-link"], cwd = Nothing, env = Nothing, std_in = Inherit, std_out = CreatePipe, std_err = CreatePipe, close_fds = False, create_group = False, delegate_ctlc = False, detach_console = False, create_new_console = False, new_session = False, child_group = Nothing, child_user = Nothing}
    ✔ Running nix-build on ./.obelisk/impl [command]
    Handing off to /nix/store/s200msjpipi4jpyv2xkas35dsb101s9m-obelisk-command-0.1/bin/ob
[...]

PR #33, opened 2018-05-29
Run all nix-builds with --sandbox

### LINKED ISSUE #28: Warn the user if sandboxing is not enabled
See https://github.com/obsidiansystems/obelisk/issues/6, in which a derivation failed to build correctly due to not being sandboxed.

We should warn the user anytime they are trying to nix-build something with obelisk and sandboxing is not enabled.

3noch:
Fixes #28


3noch:
selftest is failing for some reason on this. I'll fix it.

3noch:
I think we'll need to run `--sandbox` on only certain builds. I haven't looked deeply into it yet but some things fail to build with this flag.

ryantrinkle:
Interesting; I'm surprised some things fail.  I did add `sandbox = true` to the installation instructions, but having this as a fallback also seems good.

ali-abrar:
How does this relate to #55?

alexfmpe:
I think the idea here is enforcing sandbox whenever possible.
#55 is more about the kind of sandbox that should be used (be it recommendation or enforcing).

ryantrinkle:
@3noch Looks like this needs to be rebased/merged.

ryantrinkle:
Also: should we be using 'relaxed' somehow?

alexfmpe:
Hard to say at this point - it may very well be that #6 also happens with 'relaxed'. 
Also, neither setting seems to completely accept obelisk currently.
`true`: https://github.com/obsidiansystems/obelisk/pull/55#issuecomment-395419688 
`relaxed` : https://github.com/obsidiansystems/obelisk/issues/52#issuecomment-394168617

3noch:
Should we close this then? I don't think it's correct.

PR #35, opened 2018-05-29
Fix route config for files ending in newline

tomsmalley:


lib/run/src/Obelisk/Run.hs
```diff
@@ -48,9 +48,13 @@ run port backend frontend = do
   runWidget conf frontend `finally` killThread backendTid
 
 getConfigRoute :: IO (Maybe URI)
-getConfigRoute = do
-  mroute <- get "common/route"
-  return $ URI.mkURI =<< mroute
+getConfigRoute = get "common/route" >>= \case
+  Just r -> case URI.mkURI $ fromMaybe r $ T.stripSuffix "\n" r of
```
luigy: let's use T.strip it should be safe to remove trailing/leading whitespace

tomsmalley: Forgot about that function, done :+1:

luigy:
Thanks!

PR #36, opened 2018-05-29
Don't rerun closure-compiler if frontend hasn't changed

tomsmalley:
Moves the JS into its own derivation


3noch:
@tomsmalley @luigy Just tested changing the backend and the frontend does not rebuild.

PR #38, opened 2018-05-29
Decouple `Obelisk.CLI` from obelisk-command

srid:
Move `Obelisk.CLI` to `lib/cliapp`

Instead of importing:

```haskell
import Obelisk.CLI
```

import:

```haskell
import Obelisk.CliApp
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -93,7 +93,7 @@ let #TODO: Upstream
     addLibs = self: super: {
       obelisk-asset-manifest = self.callCabal2nix "obelisk-asset-manifest" (hackGet ./lib/asset + "/manifest") {};
       obelisk-asset-serve-snap = self.callCabal2nix "obelisk-asset-serve-snap" (hackGet ./lib/asset + "/serve-snap") {};
-      cliapp = self.callCabal2nix "cliapp" (cleanSource ./lib/cliapp) {};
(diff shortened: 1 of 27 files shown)


default.nix
```diff
@@ -93,6 +93,7 @@ let #TODO: Upstream
     addLibs = self: super: {
       obelisk-asset-manifest = self.callCabal2nix "obelisk-asset-manifest" (hackGet ./lib/asset + "/manifest") {};
       obelisk-asset-serve-snap = self.callCabal2nix "obelisk-asset-serve-snap" (hackGet ./lib/asset + "/serve-snap") {};
+      cliapp = self.callCabal2nix "cliapp" (cleanSource ./lib/cliapp) {};
```
ryantrinkle: This should be named `obelisk-cliapp`.  That naming convention should apply to everything in the `lib` directory.

Let's also add a README.md in that directory explaining this convention.

lib/cliapp/src/CliApp.hs
```diff
@@ -3,7 +3,7 @@
 -- See Demo.hs for an example
 --
 -- This package should eventually be made its own library.
-module Obelisk.CLI
+module CliApp
```
ryantrinkle: This should all still be under the `obelisk` namespace.

PR #39, opened 2018-05-30
Allow verbose output for tests

alexfmpe:


lib/selftest/src/Obelisk/SelfTest.hs
```diff
@@ -35,11 +36,16 @@ data ObRunState
 
 main :: IO ()
 main = do
+  isVerbose <- getArgs >>= \case
```
alexfmpe: Strangely, `./result anythingAtAll` makes `getArgs` return an empty list.

lib/selftest/src/Obelisk/SelfTest.hs
```diff
@@ -35,11 +36,16 @@ data ObRunState
...
+    ["-v"] -> pure True
+    []     -> putStrLn "Tests may take longer to run if there are unbuilt derivations: use -v for verbose output" *> pure False
+    _      -> putStrLn "Usage: selftest [-v]" *> exitWith (ExitFailure 2)
+  let verbosity = bool silently verbosely isVerbose
```
alexfmpe: This is probably not needed.
I tried directly using `silently/verbosely` instead of True/False above, but I get type errors e.g.
>      Expected type: Sh a -> IO ()
>      Actual type: Sh Text -> IO ()

I don't see what makes it different

Issue #52, opened 2018-06-03
Takes a really long time to install


cah6:
I'm fairly new to nix, so apologies if this is just how it is. 

I just finished installing with `nix-env -f https://github.com/obsidiansystems/obelisk/archive/master.tar.gz -iA command` and it probably took around ~8-12 hours. Is this normal? I did make sure to edit my nix.conf as required, and if it makes a difference I'm on OSX 10.13.4 tracking the nixpkgs-unstable channel. 

From looking at the output as it was going on, it looks like for a lot of stuff it missed the caches and was compiling C / Haskell files.

alexfmpe:
You probably need to restart the daemon after editing: I've updated the instructions in #53
This will probably be automated soon, since https://github.com/reflex-frp/reflex-platform already handles setting up the caches

cah6:
Ah yep I believe that's it, I didn't know before that it wouldn't automatically pick up changes to the conf file. 

Btw even smaller thing (don't want to make a new "issue" for it) that I just ran into is that for me I had to turn the sandbox parameter off when installing hub as it was giving:
`error: derivation '/nix/store/rkpli4272qmz5nsfw0a0wq5nbxjjhj9r-user-environment.drv' specifies a sandbox profile, but this is only allowed when 'sandbox' is 'relaxed'`
Might be best to mention this since the instructions before specifically say to turn sandboxing on.

cah6:
Actually, this is impacting my initial setup more than I thought. For me `ob init` and `ob run` also required `sandbox = false` (same error as above). But when I run `ob run` with this, it takes a very long time since, again, it's manually compiling a lot of haskell dependencies. I don't know if this is normal, i.e. if it takes a long time first time but then gets significantly better. Thoughts?

alexfmpe:
I've also run into that sandbox issue on mac - we might want to recommend `sandbox = relaxed` instead of `sandbox = true`.

I'm not sure but I think sandboxing doesn't change any hashes, in which case it wouldn't affect whether caches work. I'll do some experiments. 

> I don't know if this is normal, i.e. if it takes a long time first time but then gets significantly better.

It shouldn't take that long even on the first time (expected behaviour is minutes of downloads from cache), but the second time should take only a few seconds.

AFAIK lack of sandboxing has so far only been a problem when trying `ob deploy push` in Ubuntu: https://github.com/obsidiansystems/obelisk/issues/6#issuecomment-391871025.

matthewbauer:
[...]

PR #55, opened 2018-06-03
Recommend 'sandbox=relaxed'

alexfmpe:
`sandbox = true ` [is failing](https://github.com/obsidiansystems/obelisk/issues/52#issuecomment-394168617) for [mac os](https://github.com/NixOS/nix/blob/master/src/libstore/build.cc#L1809) nix builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
         ```
     1. If you are using another operating system or linux distribution, ensure that these lines are present in `/etc/nix/nix.conf`:
         ```
-        sandbox = true
+        sandbox = relaxed
         substituters = https://cache.nixos.org https://nixcache.reflex-frp.org
         trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ryantrinkle.com-1:JJiAKaRv9mWgpVAz8dwewnZe0AzzEAzPkagE9SP5NWI=
         ```


ryantrinkle:
@alexfmpe Has it been tested that this doesn't break #6?

alexfmpe:
Not yet

ryantrinkle:
Perhaps we could add something to the impure tests that test that and
this?  Also: i guess we need to set up some kind of CI for running the
impure tests on both macOS and Linux

ali-abrar:
What does "relaxed" actually do? Sorry if I missed the explanation somewhere.

alexfmpe:
From the manual:
> If this option is set to relaxed, then fixed-output derivations and derivations that have the __noChroot attribute set to true do not run in sandboxes.

We might have to change some things to allow running with `relaxed` though.
On macOS `sandbox=false` works while I'm getting

```nix-build -A ios.frontend -o result-ios --check
checking outputs of '/nix/store/xk4xhclhj6hf0z5nkfr5m32h3xwjb0hj-frontend-app.drv'...
++ mkdir -p /nix/store/6x98v2381hgdi8nwl8rf72k9hjmkbz3v-frontend-app/frontend.app
++ ln -s /nix/store/3g3557wcabdwafpkz4k3gcpfj86wr17n-Info.plist /nix/store/6x98v2381hgdi8nwl8rf72k9hjmkbz3v-frontend-app/frontend.app/Info.plist
ln: failed to create symbolic link '/nix/store/6x98v2381hgdi8nwl8rf72k9hjmkbz3v-frontend-app/frontend.app/Info.plist': File exists```

alexfmpe:
#83 dealt with this

PR #56, opened 2018-06-03
Disable spinner when not interactive

srid:
eg: when redirecting, as in hydra.


lib/cliapp/src/Obelisk/CliApp/Spinner.hs
```diff
@@ -6,39 +6,43 @@ module Obelisk.CliApp.Spinner (withSpinner) where
...
 
 -- | Run an action with a CLI spinner.
 withSpinner
   :: (MonadIO m, MonadMask m, Cli m, HasCliConfig m)
   => Text -> m a -> m a
-withSpinner s = bracket' start cleanup . const
+withSpinner s f = do
+  noSpinner <- _cliConfig_noSpinner <$> getCliConfig
+  if noSpinner
+    then bracket_ runFake (pure ()) f
```
3noch: Do you need bracket at all if there is nothing to cleanup?

Issue #58, opened 2018-06-04
Is there a way to create a deployment locally?


dalaing:
I'm playing around with Obelisk and wanted to check the size of the output js.  I couldn't find anything in the ob thunk or ob deploy options that let me get hold of the output of a project built using obelisk, unless I actually set up a server to deploy the project to.  

I don't have a pressing business need for this, I mostly just want it to be able to locally debug a few steps in my workflow.

Am I missing something? Is Obelisk missing something? Or is there magic going on under the hood that should be hidden from the users?  Hiding magic from users is a good thing, so I'm fine if that's the case :)

srid:
Looking at the code I _think_ we can do it by exposing [`serverExe`](https://github.com/obsidiansystems/obelisk/blob/6f856f4/default.nix#L161-L171) from default.nix to Obelisk projects somehow.

3noch:
I noticed this as well. Nixops has some options for `deploy` like `--copy-only` and `--build-only`. `--build-only` builds all derivations and then stops. `--copy-only` also copies them to the server but does not switch the configuration. I think we need to add options like this, especially since they can help make time-sensitive deployments go better (e.g. you can ensure everything is built and copied before you officially start a scheduled down time).

srid:
I gave this a stab in #61. 

Something fundamentally different is happening when you run the server _locally_; it seems that `getAssetPath` is returning a Nothing, which means that we cannot serve `all.js` successfully. Need to see what's going on ...

ali-abrar:
This is covered in the readme now: https://github.com/obsidiansystems/obelisk#locally

PR #61, opened 2018-06-04
Local deployment build

srid:
Run `nix-build -A release` to create the Nix package containing a runnable backend, along with frontend assets, config etc. I am yet to figure out how to make the hashed out compressedJs path work (it currently doesn't).

- [x] Fix broken access to compressedJs path to all.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
 - [Installing Obelisk](#installing-obelisk)
 - [Developing an Obelisk project](#developing-an-obelisk-project)
 - [Deploying](#deploying)
+  - [Locally](#locally)
+  - [EC2](#ec2)
 - [Mobile](#mobile)
   - [iOS](#ios)
   - [Android](#android)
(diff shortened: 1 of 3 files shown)


srid:
The suppressed exception in `getAssertTarget` is: 

```
/all.js/target: openBinaryFile: does not exist (No such file or directory)
```

srid:
@luigy fixed the aforementioned bug in 056c81e.

@ryantrinkle This is now ready for merge.

3noch:
@luigy Wouldn't it be better to not create the directory first and then copy with `-r`?

ryantrinkle:
I will need to review this.  Please don't merge yet.

srid:
@ryantrinkle wrote in #58:
> Right now, let's make sure there's a simple `nix-build` command that builds everything, and let's also make `ob deploy prepare` execute that command.  Later, we'll add more stuff to it.

Note that this PR just adds that `nix-build` command. 

It doesn't add a `ob deploy prepare` yet. I thought a bit about it and came up with this open question: _on Mac should we build a native mac backend executable, or a linux executable (using docker)?_ The later makes sense if it will be deployed to a remote server, but former makes sense if you are planning on just running it locally (which is the intention in #58). This PR assumes the former, and I've tested that it works fine on Mac (and it doesn't depend on docker).

3noch:
@srid I think it makes sense to support both in some fashion. This is a lower-level approach but it does allow you to run the server natively on mac. If you want to run a deployment and ensure that everything is already built, however, you'd want the right binaries for your target server.

ali-abrar:
Given that `nix-build -A release` is being run from the project directory, I assume it'll be using the development config found in that directory. This makes the use of `release` here a bit confusing.

srid:
@ali-abrar Good point; renamed to `nix-build -A localserver`.

3noch:
@ryantrinkle I've rebased this and added a migration. I need this for something else so would be nice to get it in.

ryantrinkle:
@3noch Thanks!

Issue #50, opened 2018-06-02
`ob run` not using main files in frontend & backend


SebastianCallh:
Replicated by running `ob init` to start a new project and then breaking/removing `Frontend/src/main.hs` or `Backend/src/main.hs` followed by `ob run` which happily launches the dev server. The main files do not seem to be used since changes are not reflected.

Is this intended? I feels off.

3noch:
That is intended behavior (c.f. https://github.com/obsidiansystems/obelisk/blob/68f0725a29eadb8583b1a7db05ce65835a172887/skeleton/devel/Devel.hs#L9). *However* this is a cause for confusion. Perhaps the skeleton could document the `main.hs` modules better with a clear explanation?

The `main` function on both sides should do as little as possible. In fact, we could probably generate them during the production build (but that might be confusing in a different way).

SebastianCallh:
Oh, of course. Can't believe I missed that. Either that or adding something to the readme under [Developing an Obelisk project](https://github.com/obsidiansystems/obelisk/blob/master/README.md#developing-an-obelisk-project)?  I could take a stab at it.

ohri-anurag:
@dfordivam Is this still an issue? I don't think new Obelisk projects have a `main.hs` file in the `frontend/src` or `backend/src` directory anymore.

alexfmpe:
@ohri-anurag they're in `src-bin` nowadays

PR #63, opened 2018-06-05
Isolate main.hs

### LINKED ISSUE #51: Move the Main.hs out of src directory
The executable should depend upon the library in cabal, but if the Main is in same folder as other files, it will try to compile it as part of executable creation (instead of using via library)

srid:
So it doesn't end up building the library sources along with it. Consequently we do not have to repeat the library cabal dependencies in the executable stanza.

Fixes #51 
Clarifies #50 a bit

### DIFF
--- a/skeleton/backend/backend.cabal
+++ b/skeleton/backend/backend.cabal
@@ -16,17 +16,13 @@ library
   ghc-options: -Wall
 
 executable backend
-  main-is: main.hs
-  hs-source-dirs: src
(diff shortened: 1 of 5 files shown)


ryantrinkle:
Instead of this, let's take out main.hs completely from the skeleton.  See https://github.com/obsidiansystems/obelisk/pull/57

ryantrinkle:
See https://github.com/obsidiansystems/obelisk/pull/57

PR #65, opened 2018-06-05
https support

srid:
Https is enabled by default. Consequently this adds two new arguments to `ob deploy init`: 

```
  --ssl-hostname SSLHOSTNAME
                           SSL hostname
  --ssl-email SSLEMAIL     SSL admin email
```

- [x] Initial development
- [x] Verify, after subdomain is activated

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -168,11 +168,11 @@ rec {
       ln -s ${compressedJs frontend} $out/frontend.jsexe
     ''; #TODO: run frontend.jsexe through the asset processing pipeline
 
-  server = { exe, hostName, adminEmail, sslHost }:
+  server = { exe, hostName, adminEmail, sslHost ? null }:
     let system = "x86_64-linux";
         nixos = import (pkgs.path + /nixos);
         # https is enabled unless sslHost is empty string
-        extraConfig = if sslHost == "" then {} else {
+        extraConfig = if sslHost == null then {} else {
           sslConfig = {
             hostName = sslHost;
             adminEmail = adminEmail;
@@ -281,7 +281,7 @@ rec {
       linuxserver = serverOn "x86_64-linux";
     in projectOut system // {
       inherit linuxserver;
-      server = args@{ hostName, adminEmail, sslHost }:
+      server = args@{ hostName, adminEmail, sslHost ? null}:
         server (args // { exe = linuxserver;});
       obelisk = import (base + "/.obelisk/impl") {};
     };
--- a/default.nix
+++ b/default.nix
@@ -171,7 +171,7 @@ rec {
   server = { exe, hostName, adminEmail, sslHost ? null }:
     let system = "x86_64-linux";
         nixos = import (pkgs.path + /nixos);
-        # https is enabled unless sslHost is empty string
+        # https is enabled unless sslHost is null
         extraConfig = if sslHost == null then {} else {
           sslConfig = {
             hostName = sslHost;
--- a/default.nix
+++ b/default.nix
@@ -273,10 +273,12 @@ rec {
                 };
               };
           in mkProject (projectDefinition args));
+      serverOn = sys: serverExe (projectOut sys).ghc.backend (projectOut system).ghcjs.frontend assets.symlinked configPath;
+      linuxserver = serverOn "x86_64-linux";
     in projectOut system // {
+      inherit linuxserver;
       server = { hostName, sslHost, adminEmail }:
-        let exe = serverExe (projectOut "x86_64-linux").ghc.backend (projectOut system).ghcjs.frontend assets.symlinked configPath;
-        in server exe hostName sslHost adminEmail;
+        server linuxserver hostName sslHost adminEmail;
       obelisk = import (base + "/.obelisk/impl") {};
     };
   haskellPackageSets = {
--- a/default.nix
+++ b/default.nix
@@ -168,16 +168,16 @@ rec {
       ln -s ${compressedJs frontend} $out/frontend.jsexe
     ''; #TODO: run frontend.jsexe through the asset processing pipeline
 
-  server = exe: hostName:
+  server = exe: hostName: sslHost: sslEmail:
     let system = "x86_64-linux";
         nixos = import (pkgs.path + /nixos);
         https = (import lib/https {}).module {
           backendPort = 8000; # TODO read from config
-          # sslConfig = {
-          #   hostName = "example.com";
-          #   adminEmail = "webmaster@example.com";
+          sslConfig = {
+            hostName = sslHost;
+            adminEmail = sslEmail;
           #   subdomains = [ "www" ];
-          # };
+          };
         };
     in nixos {
       inherit system;
@@ -274,9 +274,9 @@ rec {
               };
           in mkProject (projectDefinition args));
     in projectOut system // {
-      server = { hostName }:
+      server = { hostName, sslHost, sslEmail }:
         let exe = serverExe (projectOut "x86_64-linux").ghc.backend (projectOut system).ghcjs.frontend assets.symlinked configPath;
-        in server exe hostName;
+        in server exe hostName sslHost sslEmail;
       obelisk = import (base + "/.obelisk/impl") {};
     };
   haskellPackageSets = {
--- a/default.nix
+++ b/default.nix
@@ -168,17 +168,22 @@ rec {
       ln -s ${compressedJs frontend} $out/frontend.jsexe
     ''; #TODO: run frontend.jsexe through the asset processing pipeline
 
-  server = exe: hostName: sslHost: adminEmail:
+  server = { exe, hostName, adminEmail, sslHost }:
     let system = "x86_64-linux";
         nixos = import (pkgs.path + /nixos);
-        https = (import lib/https {}).module {
-          backendPort = 8000; # TODO read from config
+        # https is enabled unless sslHost is empty string
+        extraConfig = if sslHost == "" then {} else {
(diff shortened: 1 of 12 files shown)


srid:
We might want to look into revamping the deployment configuration mechanism. At the moment the deploy directory is growing to hold much more small config files:

```
backend_hosts        config  ssh_key    ssl_host
backend_known_hosts  src     ssl_email
```

srid:
Verified :tada: https://obelisk-skeleton.srid.ca/

ryantrinkle:
@srid Awesome, thanks!  One thing to note: i think we should have a single email address for all alerts, not one specific to SSL.

srid:
@ryantrinkle I've renamed `--ssl-email` to `--admin-email` - so the later can be reused in other places.

lib/command/src/Obelisk/Command.hs
```diff
@@ -161,6 +161,8 @@ deployInitOpts = DeployInitOpts
   <$> strArgument (action "directory" <> metavar "DEPLOYDIR" <> help "Path to a directory that it will create")
   <*> strOption (long "ssh-key" <> action "file" <> metavar "SSHKEY" <> help "Path to an ssh key that it will symlink to")
   <*> some (strOption (long "hostname" <> metavar "HOSTNAME" <> help "hostname of the deployment target"))
+  <*> strOption (long "ssl-hostname" <> metavar "SSLHOSTNAME" <> help "SSL hostname")
+  <*> strOption (long "admin-email" <> metavar "ADMINEMAIL" <> help "Site admin email")
```
ryantrinkle: We need a better description here.  How about "Email address where administrative alerts will be sent"?

lib/command/src/Obelisk/Command.hs
```diff
@@ -161,6 +161,8 @@ deployInitOpts = DeployInitOpts
   <$> strArgument (action "directory" <> metavar "DEPLOYDIR" <> help "Path to a directory that it will create")
   <*> strOption (long "ssh-key" <> action "file" <> metavar "SSHKEY" <> help "Path to an ssh key that it will symlink to")
   <*> some (strOption (long "hostname" <> metavar "HOSTNAME" <> help "hostname of the deployment target"))
+  <*> strOption (long "ssl-hostname" <> metavar "SSLHOSTNAME" <> help "SSL hostname")
```
ryantrinkle: How about "The hostname that end-users will use to access the site"

Also, come to think of it, this should just be taken from the route, shouldn't it?

lib/command/src/Obelisk/Command.hs
```diff
@@ -181,6 +183,8 @@ data DeployInitOpts = DeployInitOpts
   { _deployInitOpts_outputDir :: FilePath
   , _deployInitOpts_sshKey :: FilePath
   , _deployInitOpts_hostname :: [String]
+  , _deployInitOpts_sslHostname :: String
+  , _deployInitOpts_adminEmail :: String
```
ryantrinkle: What is with all the use of `String` here?  Why not `Text`?

3noch: This whole code base could use a revamp to use `Text`. `process` uses `String` for all its arguments and I think that bled all the way up to our types.

srid: Yup, I used `String` only to be consistent. I can open a separate PR that converts the whole code base to use `Text`.

ryantrinkle: @3noch I understand how that can happen, but it's a symptom of using the wrong approach.  People should always keep the core of their code in the best types available, and always do conversions to worse types at the edge of the system.  In other words, convert to "better" types at the earliest possible time, and convert to "worse" types at the last possible moment.  So, `String` should never pollute datastructures like this that we are writing.

@srid That sounds excellent; thanks!

ryantrinkle: @srid And I 100% agree about keeping consistent with the surrounding code.

3noch: @ryantrinkle I very much agree. I was tempted to make this transition multiple times but realized that using both `Text` and `String` in unpredictable places would be worse than just being consistent and doing a big switch at some point.

lib/command/src/Obelisk/Command.hs
ryantrinkle: @srid That sounds good.  Let's make sure we cover all our bases, e.g. if someone passes an invalid URL or a URL with a path, etc.

default.nix
```diff
@@ -168,16 +168,16 @@ rec {
       ln -s ${compressedJs frontend} $out/frontend.jsexe
     ''; #TODO: run frontend.jsexe through the asset processing pipeline
 
-  server = exe: hostName:
+  server = exe: hostName: sslHost: adminEmail:
```
ryantrinkle: I don't think we should just add more params here.  Let's make it use a record or something instead.

default.nix
```diff
@@ -168,16 +168,16 @@ rec {
...
-          # sslConfig = {
-          #   hostName = "example.com";
-          #   adminEmail = "webmaster@example.com";
-          #   subdomains = [ "www" ];
-          # };
+          sslConfig = {
+            hostName = sslHost;
+            adminEmail = adminEmail;
+            subdomains = [ ];
+          };
```
ryantrinkle: It looks like this makes HTTPS *mandatory*.  It should not be mandatory - HTTPS termination doesn't make sense, e.g., if the server is running behind a load balancer.

srid: Pushed a change that includes `--disable-https` to disable the by-default https behaviour.

ryantrinkle: We should not use "" to denote a missing sslHost; it should be `null`.

default.nix
```diff
@@ -168,17 +168,22 @@ rec {
...
+        # https is enabled unless sslHost is empty string
+        extraConfig = if sslHost == "" then {} else {
+          sslConfig = {
+            hostName = sslHost;
+            adminEmail = adminEmail;
+            subdomains = [ ];
+          };
         };
+        httpsConfig = {
+          backendPort = 8000; # TODO: read from config
```
ryantrinkle: This todo looks wrong.

default.nix
```diff
@@ -168,17 +168,22 @@ rec {
...
+        } // extraConfig;
+        # TODO: switch from lib/https to nixos builtin ssl config
```
ryantrinkle: Please make a ticket for this one.

lib/command/src/Obelisk/Command.hs
```diff
@@ -161,6 +161,9 @@ deployInitOpts = DeployInitOpts
   <$> strArgument (action "directory" <> metavar "DEPLOYDIR" <> help "Path to a directory that it will create")
   <*> strOption (long "ssh-key" <> action "file" <> metavar "SSHKEY" <> help "Path to an ssh key that it will symlink to")
   <*> some (strOption (long "hostname" <> metavar "HOSTNAME" <> help "hostname of the deployment target"))
+  <*> strOption (long "route" <> metavar "PUBLICROUTE" <> help "Public URL to your app (comon/route)")
```
ryantrinkle: Let's take out `(common/route)` here.  Without further explanation, that won't mean anything to people.

lib/command/src/Obelisk/Command.hs
```diff
@@ -181,6 +184,9 @@ data DeployInitOpts = DeployInitOpts
   { _deployInitOpts_outputDir :: FilePath
   , _deployInitOpts_sshKey :: FilePath
   , _deployInitOpts_hostname :: [String]
+  , _deployInitOpts_route :: String
+  , _deployInitOpts_adminEmail :: String
+  , _deployInitOpts_disableHttps :: Bool
```
ryantrinkle: This should be `enableHttps` here

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -116,6 +140,9 @@ deployPush deployPath getNixBuilders = do
           ["-C", deployPath, "add", "."]
         callProcessAndLogOutput (Debug, Error) $ proc "git"
           ["-C", deployPath, "commit", "-m", "New deployment"]
+    putLog Notice $
+      (if disableHttps then "Deployed (without https) => " else "Deployed => ")
```
ryantrinkle: I don't think we need to mention https here.

lib/command/src/Obelisk/Command.hs
srid: Here's the full range of possible route validation:

```haskell
    InvalidRoute_MissingScheme uri -> route uri "must have an URI scheme"
    InvalidRoute_NotHttps uri -> route uri "must be HTTPS"
    InvalidRoute_MissingHost uri -> route uri "must contain a hostname"
    InvalidRoute_HasPort uri -> route uri "cannot specify port"
    InvalidRoute_HasPath uri -> route uri "cannot contain path"
 ```

File: default.nix (hunk 5 above)
Comment (srid) [reply default.nix]: Done.

**Comment by ryantrinkle** [comment]
@srid How can I upgrade an existing obelisk project to use this?

**Comment by srid** [comment]
> @srid How can I upgrade an existing obelisk project to use this?

You might want to test/use #77 instead of this PR.

In any case, try this from the deploy directory of that project:

```
echo True > enable_https
echo admin@reflex-frp.org > admin_email
echo https://reflex-frp.org/ > config/common/route
git add ... && git commit ...
ob deploy push
```

srid:
Superseded by #97

Issue #67, opened 2018-06-06
The `static` directory cannot be a symlink


dalaing:
I don't know if `static` _should_ be allowed to be a symlink, but I know I'd find it really handy if it could be.

At the moment I'm trying to use a symlink to a nix derivation as the `static` directory, so that I can do various odd things to prepare my static assets.

If this is out of scope for Obelisk I completely understand, but I figured I'd note it here.  If you're open to it, great.  If you're not open to it, it'll save someone else from asking or filing an issue about it.

ryantrinkle:
Would it make sense, instead, to have a way that you can configure obelisk so that your static files are represented by a nix *expression*?  That way, it could be automatically rebuilt for you whenever it changes.  We could just set the default expression to be `./static`.

dalaing:
That sounds great to me :)

ali-abrar:
Fixed via #71

PR #71, opened 2018-06-07
WIP: Allow the static files directory to be overridden

ryantrinkle:
@dalaing This PR should allow you to provide a `static` argument in your project's default.nix, which can be an arbitrary nix derivation.  Do you think this would help with #67?

@luigy Do we need to do anything with `ob run` to make it support this?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -218,20 +218,19 @@ rec {
 
   # An Obelisk project is a reflex-platform project with a predefined layout and role for each component
   project = base: projectDefinition:
-    let assets = processAssets { src = base + "/static"; };
-        configPath = base + "/config";
+    let configPath = base + "/config";
         projectOut = sys: (getReflexPlatform sys).project (args@{ nixpkgs, ... }:
           let mkProject = { android ? null #TODO: Better error when missing
                           , ios ? null #TODO: Better error when missing
                           , packages ? {}
                           , overrides ? _: _: {}
+                          , static ? base + /static
                           }:
               let frontendName = "frontend";
                   backendName = "backend";
                   commonName = "common";
                   staticName = "static";
-                  staticPath = base + "/static";
-                  assets = processAssets { src = base + "/static"; };
+                  assets = processAssets { src = static; };
                   # The packages whose names and roles are defined by this package
                   predefinedPackages = filterAttrs (_: x: x != null) {
                     ${frontendName} = nullIfAbsent (base + "/frontend");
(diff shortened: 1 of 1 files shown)


dalaing:
It looks like `obelisk-assert-manifest-generate` is still struggling to deal with the symlink:

```
building '/nix/store/ab1cqva27y4mxdf9v873swnwmknl63ms-asset-manifest.drv'...
obelisk-asset-manifest-generate: /nix/store/4xl8ky7rlh2dgf8nk72y98s4is0xdjjr-static: getDirectoryContents:openDirStream: does not exist (No such file or directory)
builder for '/nix/store/ab1cqva27y4mxdf9v873swnwmknl63ms-asset-manifest.drv' failed with exit code 1
cannot build derivation '/nix/store/8p18pfvhwamqc8kafzapxaq8l73dzckq-cabal2nix-static.drv': 1 dependencies couldn't be built
error: build of '/nix/store/8p18pfvhwamqc8kafzapxaq8l73dzckq-cabal2nix-static.drv' failed
(use '--show-trace' to show detailed location information)
```

ryantrinkle:
A symlink won't work, unfortunately, because Nix doesn't traverse or rewrite symlinks when pulling them into the store.  Instead of a symlink, can you just put the path to the real destination directly in your default.nix?

dalaing:
I've got a directory that has my css / js / fonts in it, along with a pile of markdown files, and then I have a default.nix that uses pandoc to turn those markdown files into html and marshalls all of my assets into place.

At the moment, I'm recursively copying the contents of the `result` directory into `static/`.  

- if I symlink that to `static`, things go badly as above.  
- if I rename the directory with the `default.nix` into `static`, I get: `No instance for (StaticFile "css/syntax.css")` and the like.

I'm guessing I don't have things laid out in the way the manifest generator is expecting? Is there a good way to debug what is in the static asset manifest?  Or another approach you could recommend?

ryantrinkle:
What I'm imagining is *deleting* static, and then putting something in your default.nix like `static = import your/static/builder/dir`.  I think that may work somewhat, but I do suspect it'll go wrong with `ob run`.

dalaing:
Cool, I'll try some things and get back to you.

dalaing:
Hah - I had just messed up my use of this.  It works fine :)  

If `ob run` could catch changes to this and rebuild things that might be even nicer for some folks, although my stuff isn't changing that often so I'm very much good for now.

dalaing:
While it all builds fine, I'm getting a loop of 404s when I use this with `ob run`. I'll poke around with it and see what I can find.

ryantrinkle:
@dalaing I bet somewhere in `ob run` it's hardcoded to `./static`.  We just need to make it get the location from `nix-build` instead :)

PR #72, opened 2018-06-09
Fix thunk pack error message

alexfmpe:


alexfmpe:
Made redundant by #73

PR #73, opened 2018-06-09
ob thunk pack: relax excessive checks

alexfmpe:
This makes `ob thunk pack` succeed if 
  * all branches were pushed *somewhere* 
  * branch being thunked was pushed to specified upstream 

both apply, instead of the current behavior of failing unless all remotes have all branches, which can be a real annoyance.

### DIFF
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -678,6 +678,11 @@ getThunkPtr' checkClean thunkDir upstream = do
           ] ++ T.lines stashOutput
       True -> return ()
 
+    b <- readGitProcess thunkDir ["rev-parse", "--abbrev-ref", "HEAD"]
+    isBranchPushed thunkDir b upstream >>= \case
+      False -> failWith $ mconcat ["thunk pack: current branch (", b, ") has not been pushed to requested remote: ", upstream]
+      True -> return ()
+
     --Check whether all local heads are pushed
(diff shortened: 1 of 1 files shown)


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -775,6 +778,17 @@ getThunkPtr' checkClean thunkDir upstream = do
         }
       }
 
+isBranchPushed :: MonadObelisk m => FilePath -> Text -> Text -> m Bool
+isBranchPushed thunkDir hd rm = do
+  hasBranch <- not . null . T.lines <$> readGitProcess thunkDir ["ls-remote", "--heads", T.unpack rm, T.unpack hd]
```
alexfmpe: This fixes an unfriendly error when a branch does not even exist in a remote, e.g.
```
RawCommand "nix-shell" ["-p","git","--run","'git' '-C' 'reflex-platform' 'merge-base' 'f4c17b91e6104830667ded50e950dd08f5e79bfe' 'remotes/origin/fix-ios-simulator'"] failed with exit code: 128
```

ryantrinkle:
@alexfmpe Could we add some tests for this?  Especially the negative case, i.e. when things have not been pushed.  I don't want to delete data unexpectedly.

Ericson2314:
@alexfmpe I think the status quo now subsumes this?

alexfmpe:
@Ericson2314 I agree. Closing

PR #77, opened 2018-06-11
Replace obelisk-https with nixos acme

### LINKED ISSUE #75: Remove obelisk-https
> @ryantrinkle : Note: obelisk-https itself should be deprecated and removed from obelisk. The built-in support in nixos seems to be good enough now.

srid:
Based on the branch for #65. Fixes #75 

- [x] Test http deployment
- [x] Test https deployment (waiting on domain update)


srid:
Note: this is not yet in master, that will happen in #97

PR #79, opened 2018-06-14
ob run: Handle application websocket requests

luigy:
Workaround `jsaddleOr` wanting to handle all websockets requests without
giving a chance for `run` to proxy non jsaddle websocket requests to the
backend. This assumes that websocket requests to "/" are meant for
jsaddle which is what the current version pinned does


3noch:
I've tested this and it works.

PR #80, opened 2018-06-14
Backfill migration graph

srid:
This PR backfills the two migration graphs (called `obelisk-upgrade` and `obelisk-handoff`) with default migration type value (empty string, and `True` accordingly).

The vertex hashes were generated using:
```
(for ref in `git log --abbrev-commit --pretty=oneline | awk '{print $1}'`; do .migration/obelisk-upgrade.hash.sh $ref; done) | uniq > hashes
```
(See the aforementioned hash script for the actual files included).


srid:
This will become part of the upcoming `ob-upgrade` PR.

cgibbard:
That's a larger number of True's than I was expecting...

srid:
> That's a larger number of True's than I was expecting...

Oops; I now remember we agreed upon something slightly different ... could you remind me what that is?

cgibbard:
Most of the edges should be mempty (i.e. Any False which you might choose to represent as a blank file perhaps), while the commit which introduces the upgrade command should have (Any True) as its value.

PR #83, opened 2018-06-16
Add per-platform sandboxing instructions

alexfmpe:
It seems that even with `sandbox = relaxed` there are problems for macOS, so I think we need to separate the instructions.

I'd say #28 still makes sense, as long as it takes both platforms into consideration.

IMO #33 #55 should be closed and we'll want to have a ticket for all sandbox-related problems that tracks what needs to be fixed for each platform/sandbox-mode pair


ryantrinkle:
Thanks!

PR #84, opened 2018-06-16
Use wrapProgram make ob supply its own runtime deps

### LINKED ISSUE #78: Utils.git ignores failing commands
For example,

```haskell
        git obImpl
        [ ["checkout", T.unpack gitBranch]
        , ["pull"]
        ]
```

If the `checkout` command fails, this merrily continues on without failing immediately.

3noch:
The test `can build everything` is failing and I haven't looked into why yet. If you happen to have an idea why, let me know.

Also, installing this makes obelisk show up as `ob` instead of `obelisk-command` in `nix-env -q`. I can change this if we liked `obelisk-command` as the CLI package name. I thought `ob` was a bit more `ob`vious.

For devs, you now should use `nix-shell -A commandEnv` instead of `nix-shell -A command.env` to get into dev shell for `command`.

Fixes #78

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -16,6 +16,15 @@ let
       then removeConfigureFlag drv' "--ghc-option=-optl=-dead_strip"
       else drv';
 
+  commandRuntimeDeps = pkgs: with pkgs; [
+    coreutils
+    git
+    gitAndTools.hub
+    nix-prefetch-git
+    nixStable
+    openssh
+  ];
+
   #TODO: Upstream
   # Modify a Haskell package to add completion scripts for the given
   # executable produced by it.  These completion scripts will be picked up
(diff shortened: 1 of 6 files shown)


3noch:
Because `hub` is in this list, we should probably also run `hub` ourselves to ensure that the auth is configured with the version of hub we are using. *Or* we should not bring hub with us and rely on the user already having it (but this assumes that the way hub stores auth is compatible with how we expect, which may not be true for various combinations of versions of hub and ob).

ryantrinkle:
Thanks!

srid:
@3noch How does this affect the use of `nix-shell ....env` commands (there are two of them) in the README? cf. https://github.com/obsidiansystems/obelisk#contributing-to-obelisk

srid:
Actually this PR doesn't affect it. n/m.

PR #85, opened 2018-06-16
Add `ob upgrade` and migration graph

srid:
- Add `ob upgrade` command to upgrade obelisk of a project and display migrations to be performed
- Add migration graph, backfilled for `master` and this branch

## Status

Functional, and ready for wider testing, however the branch has a trail of TODOs/FIXMEs as potential next stage of work. In order to test, install this branch of Obelisk to your $PATH, then go to your Obelisk project and run `ob upgrade ob-upgrade-pr` (as soon as this PR is merged, we would be able to test it with other open PRs).

## Developer workflow as of now

Immediately after every individual commit they make (in any branch), the developer must generate new vertices/edge for that commit by running the following commands:

```sh
./migration/backfill.sh
git add migration
git commit --amend --no-edit
```

Alternatively they may do it for multiple commits at a time (but not `amend`ing) at the expense of not being able to migrate to/from the parent commit inside.


ryantrinkle:
@srid Awesome :) Do you have a sense of how long the TODOs/FIXMEs will take?  I'd definitely like to get this merged ASAP.

Regarding backfill.sh: can they just run it prior to committing, instead of amending?

lib/command/src/Obelisk/Command/Upgrade.hs
```diff
@@ -0,0 +1,186 @@
...
+
+fromGraphName :: Text -> MigrationGraph
+fromGraphName = \case
+  "obelisk-handoff" -> MigrationGraph_ObeliskHandoff
+  "obelisk-upgrade" -> MigrationGraph_ObeliskUpgrade
+  _ -> error "Invalid graph name specified"
+
+
+ensureCleanProject :: MonadObelisk m => FilePath -> m ()
+ensureCleanProject project = ensureCleanGitRepo project False "Cannot upgrade with uncommited changes"
```
ryantrinkle: Is there a specific reason we can't upgrade when there are uncommitted changes?  And, if so, why do we not care about ignored changes?

lib/command/src/Obelisk/Command/Upgrade.hs
```diff
@@ -0,0 +1,186 @@
...
+
+-- | Decide whether we (ambient ob) should handoff to project obelisk before performing upgrade
+decideHandOffToProjectOb :: MonadObelisk m => FilePath ->  m Bool
+decideHandOffToProjectOb project = do
+  ensureCleanProject project
+  updateThunk (toImplDir project) $ \projectOb -> do
+    getMigrationGraph projectOb MigrationGraph_ObeliskHandoff >>= \case
```
ryantrinkle: I don't think we should be checking the project migration graph for anything.  What we discussed previously was using the ambient ob's migration graph; is there a reason that can't work?

lib/command/src/Obelisk/Command/Upgrade.hs
```diff
@@ -0,0 +1,186 @@
...
+        Just (m, _, ambientHash) -> do
+          unless (hasVertex ambientHash m) $
+            failWith "Ambient ob's hash is not in its own graph"
+          return (m, ambientHash)
+
+-- | Return the path to the current ('ambient') obelisk process Nix directory
+getAmbientOb :: MonadObelisk m => m FilePath
+getAmbientOb = takeDirectory . takeDirectory <$> liftIO getExecutablePath
+
+getEdge :: MonadObelisk m => Migration action -> (Hash, Hash) -> m action
```
ryantrinkle: We should move all of the non-ob-specific migration graph stuff into a separate module where it can be reused.

lib/migration/src/Obelisk/Migration.hs
```diff
@@ -0,0 +1,112 @@
...
+          return $ Just c
+        False -> return Nothing
+
+-- | Finds a path from a to b
+--
+-- Vertices a and b must already exist (as verified by hasVertex).
+findPath :: Hash -> Hash -> Migration action -> Maybe [(Hash, Hash)]
+findPath a b (Migration g' _) = fmap listInPairs $ go a b $ adjacencyMap $ getDag g'
+  where
+    -- Do a slow and dummy search. TODO: improve algorithm performance
```
ryantrinkle: Doesn't the graph lib have an algorithm for this?

lib/migration/src/Obelisk/Migration.hs
```diff
@@ -0,0 +1,112 @@
...
+
+-- | Get the directed-acylic graph of migration
+getDag :: Ord a => AdjacencyMap a -> AdjacencyMap a
+getDag g = case topSort g of
+  Just _ -> g
+  Nothing -> error "Cyclic graph detected"
+
+-- Return actions that migrate from one hash to another
+runMigration :: Migration action -> Hash -> Hash -> Maybe [(Hash, action)]
+runMigration m a b = fmap getAction <$> findPath a b m
```
ryantrinkle: All paths in the DAG from `a` to `b` must sum to the same `action` value, otherwise the migration is invalid.  It looks like that check may still need to be added.

ryantrinkle:
Getting close :)

srid:
> @srid Awesome :) Do you have a sense of how long the TODOs/FIXMEs will take? I'd definitely like to get this merged ASAP.

@ryantrinkle Great! I'd say a few hours of work, including addressing your comments above.

> Regarding backfill.sh: can they just run it prior to committing, instead of amending?

Only if we switch it away from using `git ls-tree` to something else for computing hash. AFAIK, `git ls-tree` requires a commit ref (there appears to be no option to make it work for a modified working copy).

There is also the open question of how to ensure, before merging a PR, that we can verify that the branch has a graph edge from latest master to HEAD of the branch. One way is to merge PRs always using `git merge --ff-only` (though I don't think Github UI supports that).

lib/command/src/Obelisk/Command/Upgrade.hs
srid: The only files `ob upgrade` will touch in the project repo is the `.obelisk/impl` thunk. 

At minimum we want that path to be clean, but other than that there is no reason to require the rest of the repository to be clean. I just figured it is easier to reuse that function instead of writing another that checks just for the subpath. Should we do the specific check?

lib/migration/src/Obelisk/Migration.hs
srid: No, I couldn't find one. If our implementation gets slower, we may be able to switch over to one of [these algorithms](https://en.wikipedia.org/wiki/Pathfinding#Algorithms).

lib/migration/src/Obelisk/Migration.hs
srid: Good point; that would also tell us the 'integrity' of the migration graph.

srid: Addressed this in 259b853 and 2e81b11

lib/cliapp/src/Obelisk/CliApp/Spinner.hs
```diff
@@ -23,7 +27,15 @@ import Obelisk.CliApp.Types (Cli, CliConfig (..), HasCliConfig, Output (..), get
 withSpinner
   :: (MonadIO m, MonadMask m, Cli m, HasCliConfig m)
   => Text -> m a -> m a
-withSpinner s f = do
+withSpinner s = withSpinner' (s, const $ Just s)
+
+-- | Advanced version that controls the last spinner output based on action's result.
+withSpinner'
+  :: (MonadIO m, MonadMask m, Cli m, HasCliConfig m)
+  => (Text, a -> Maybe Text) -- ^ How to (and whether to) render the final spinner message; note that this doesn't take sub-spinner behaviour into account (although it probably should)
```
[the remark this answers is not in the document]
srid: Improved the API in 6c09b99

srid:
This is superseded by #93

PR #91, opened 2018-06-19
Make obelisk-executable-config-inject more flexible

3noch:


3noch:
@luigy @ali-abrar Thoughts?

3noch:
@ryantrinkle?

ryantrinkle:
@3noch can you elaborate on the purpose / use case for this?

3noch:
I wanted to enable some configs to be passed via command line arg in addition to the config file without actually overwriting any existing config file that is there. The way this library was implemented didn't make this possible. Factoring out the "injection encoding" as its own function allows more complex use cases to decide how to get the injection content differently (maybe from a db, env var, etc.).

3noch:
Granted `injectPure` isn't a great name. `encodeAsInjection` might be better.

srid:
The typed config stuff presents an alternate way to achieve this.

PR #92, opened 2018-06-21
Reuse the same official package name so nix-env does an upgrade properly

3noch:
@srid pointed out that the official package name had changed with my #84 PR which makes `nix-env` try to install a new package that has conflicting binaries on the path (both have `ob`). This changes the package name to what it was before so that nix-env does the right thing.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -123,7 +123,7 @@ rec {
   inherit (reflex-platform) nixpkgs pinBuildInputs;
   path = reflex-platform.filterGit ./.;
   obelisk = ghcObelisk;
-  command = pkgs.runCommand "ob" { nativeBuildInputs = [pkgs.makeWrapper]; } ''
+  command = pkgs.runCommand "obelisk-command" { nativeBuildInputs = [pkgs.makeWrapper]; } ''
     mkdir -p "$out/bin"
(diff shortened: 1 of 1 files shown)


3noch:
@ryantrinkle This should be merged ASAP.

default.nix
```diff
@@ -123,7 +123,7 @@ rec {
   inherit (reflex-platform) nixpkgs pinBuildInputs;
   path = reflex-platform.filterGit ./.;
   obelisk = ghcObelisk;
-  command = pkgs.runCommand "ob" { nativeBuildInputs = [pkgs.makeWrapper]; } ''
+  command = pkgs.runCommand "obelisk-command" { nativeBuildInputs = [pkgs.makeWrapper]; } ''
```
ryantrinkle: Can't you get the name from the underlying derivation?

3noch: You probably can, but I think pinning it explicitly might be even better, for the very reason that it matters.

ryantrinkle: I think it would make sense for the wrapper to not be a name-changing thing.  If we want to ensure that the name doesn't change, I think it would be reasonable to add a test for that.

3noch:
@ryantrinkle Fixed.

ryantrinkle:
Thanks!

PR #105, opened 2018-06-29
Fix typo

alexfmpe:


ryantrinkle:
@srid How can we make it as easy as possible for this PR to merge?

srid:
@ryantrinkle Since the commit(s) was(were) made without a migration, you will need to rewrite them. An easy way is as follows,

```sh
# First, rebase from master
git fetch && git rebase origin/master  # you may use `git merge` as well
# Substitute `N` with the number of commits in the branch:
export N=1
git reset --soft HEAD~$N
ob internal create-migration 
git add migration
git commit -m "Your commit message"
ob internal verify-migration
git push -f  
```

ryantrinkle:
@srid Since verify-migration doesn't actually read the git graph, we should be able to do it without rewriting commits, right?  Let's see if we can get this down to a single, additional commit.

srid:
@ryantrinkle Okay, I see what you mean.

The `create-migration` command so far works under the assumption that all commits have a vertex in the graph. We can relax that assumption, and provide a way to create a migration from a different commit, instead of from HEAD commit. Something like:

```sh
ob internal create-migration --from=HEAD~1  # default being HEAD
```
(For this PR, `HEAD~1` should refer to the commit in master when this branch was created).

Should be trivial to implement.

ryantrinkle:
Do we even need it to inspect git at all?  What if it just assumes the migration starts from the apex of the graph - i.e. the version that has no incoming edges.

@danharaj Any thoughts?

danharaj:
Since for now Obelisk has a linear migration graph I think that's reasonable logic. I still have to work out how to deal with implicit apexes when there's branches.

srid:
That would work as long as the apex of a linear graph (non-linear graph has multiple apexes) is the same as the hash of the base of the branch.

danharaj:
@srid What do you mean by base branch, and do you mean the commit hash?

PR #109, opened 2018-07-05
Check for TERM=dumb when enabling spinner

matthewbauer:
The spinner uses a lot of characters that can break things in some
dumb terminals.


eskimor:
lgtm

ryantrinkle:
What do we need to do to get hydra to pass?

srid:
@ryantrinkle https://github.com/obsidiansystems/obelisk/pull/105#issuecomment-401377963

PR #112, opened 2018-07-08
Routes

ryantrinkle:


srid:
I was playing with this branch over the weekend, and noticed that I was unable to call a function (from within `runRouteViewT`) like this:

```haskell
-- FIXME: can't use this with runRouteViewT (constraint violations)
 fetchUrl ::
   ( PostBuild t m
   , TriggerEvent t m
   , PerformEvent t m
   , MonadJSM m
   , MonadJSM (Performable m)
   , HasJSContext (Performable m)
   )
   => Text -> m (Event t Text)
 fetchUrl url = do
   let req = xhrRequest "GET" url def
   pb <- getPostBuild
   asyncReq <- performRequestAsync (tag (constant req) pb)
   pure $ fmap (fromMaybe "fetchUrl: Unknown error" . _xhrResponse_responseText) asyncReq
```

Results in the error:

```
frontend/src/Frontend.hs:46:14: error:
    • No instance for (PerformEvent
                         (SpiderTimeline Global)
                         (RoutedT
                            (SpiderTimeline Global)
                            a
                            (EventWriterT
                               (SpiderTimeline Global)
                               (Endo (R Route))
                               (PostBuildT
                                  DomTimeline
                                  (ImmediateDomBuilderT
                                     DomTimeline
                                     (WithJSContextSingleton
                                        x (PerformEventT DomTimeline DomHost)))))))
        arising from a use of ‘fetchUrl’
 ```

Not sure if this is a bug or I'm just missing something.

**Comment by srid** [comment]
> I was playing with this branch
... i.e., the `routes` branch *before* the reunification merge (which fails to build, for different reasons).

**Comment by ryantrinkle** [comment]
Yep, RoutedT doesn't have many instances yet.  We'll have to go through and add them at some point.

**Comment by srid** [comment]
If I encode a route to a string with a slash in it (`Just "foo/bar"`) there won't be any compiler complaints, but it just won't work, as accessing that URL will end up going to the landing route.

**Comment by ryantrinkle** [comment]
@srid Yes, that's a very good point.  pathComponentEncoder should work based on `Maybe PathComponent`, not `Maybe Text`, where `PathComponent` is just like `Text` but it can't include "/" (and possibly some other characters).

**Comment by ryantrinkle** [comment]
Actually, probably it should just do URI encoding, so "/" should become "%2F"

**Comment by ryantrinkle** [comment]
Created issue #118

File: lib/run/src/Obelisk/Run.hs [hunk 1]
```diff
@@ -74,31 +116,97 @@ runWidget conf (h, b) = do
...
+      then sendResponse $ W.responseLBS H.status200 [("Content-Type", "text/html")] html
+      else backend req sendResponse
 
 indexHtml :: StaticWidget () () -> IO ByteString
 indexHtml h = do
   ((), bs) <- renderStatic $ el "html" $ do
     el "head" $ h
     el "body" $ return ()
-    elAttr "script" ("src" =: "/jsaddle.js") $ return ()
+    elAttr "script" ("src" =: "/jsaddle/jsaddle.js") $ return ()
```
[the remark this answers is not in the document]
ryantrinkle: Eventually, we'll generate it instead of hard-coding it

srid:
@ryantrinkle Do you know how to have URL paths like http://example.com/wiki/someName where `someName` is arbitrary? 

I imagine there would be a `WikiRoute` but how would you specify that it takes an 'argument' (someName, here) that is arbitrary (and thus determined at runtime)?

lib/frontend/src/Obelisk/Frontend.hs
```diff
@@ -0,0 +1,155 @@
...
+  , MonadFix (Performable m)
+  , PrimMonad m
+  , Prerender x m
+  , EventWriter t (Endo route) m
+  )
+
+data Frontend route = Frontend
+  { _frontend_head :: !(forall t m x. ObeliskWidget t x route m => RoutedT t route m ())
+  , _frontend_body :: !(forall t m x. ObeliskWidget t x route m => RoutedT t route m ())
+  , _frontend_title :: !(route -> Text)
```
srid: Does this not introduce a restriction on the page title being determined **solely** from the route? 

In my particular case I fetch content from the backend (and this fetching happens inside `_frontend_body`), which returns the title (think markdown rendered page with a title extracted out of it), but I cannot use it in `_frontend_title`.

srid: Actually changing `_frontend_title` has no effect (at least on `ob run`); bug?

ryantrinkle: Yep, it sure does!  Let's figure out what it really should be.

The initial reason I put this in here was to be able to give something to history.pushState; however, maybe there's another way we can achieve that without making this so inflexible.

ryantrinkle:
@srid Regarding the arbitrary route: the way to do it is like this:

```haskell
data Route :: * -> * where
  Route_Landing :: Route ()
  Route_Page :: Route Text
```

srid:
Just caught a bug:

```
jsaddle.js:67 Resource interpreted as Stylesheet but transferred with MIME type text/html: "http://localhost:8000/wiki/static/1csak72m6s1c9wvzqj2q2dsl43lgm0q25kzl9lb1p9gi9bfnzh37-semantic.min.css".
```

It is trying to locate static asserts under `/wiki/`.

ryantrinkle:
@srid Hm, looks like it might not have the `base` tag set properly?  Static assets should always be located at `/static`

srid:
Oh, right. `elAttr "base" ("href" =: "/") blank` resolves it.

--

In regards to `Route_Page :: Route Text` I was able to make it work (after forking `pathOnlyValidEncoder` to use `Text` instead of `[Text]`), however it does not retain the `Universe` instance's behaviour of checking that the value is a member of a given `[Text]`.

srid:
I came across another challenge with using `Route_Page :: Route Text` (instead of `Route_Page :: Text -> Route ()`): I don't have access to the `Text` from the function being passed to `subRoute` (note: it is `f c` instead of `f c r'`),

```
subRoute_ f = factorRouted $ strictDynWidget_ $ \(c :=> r') -> do
  runRoutedT (f c) r'
```

Issue #116, opened 2018-07-12
`master`'s migration graph is incorrect


srid:
When PR #91 was merged I think it introduced a non-mempty diamond (which we don't support yet). Looking at git history I see that that branch was not rebased with `master` (so as to avoid the non-mempty diamond) _prior_ to merging.

@ryantrinkle  We can manually fix this in master now, but how to ensure that this doesn't happen in the future?

```
$ ob internal verify-migration
✔ Reading migration graph
✔ Checking graph integrity
No vertex found for HEAD (b0b562b83e8a198bb858deeb19d5868c)
✖ Checking existence of HEAD vertex
```

```
$ ob internal hash
b0b562b83e8a198bb858deeb19d5868c
$ ob internal hash HEAD~1
e66afc8befac4d76b5780a960f768c31
$ ob internal hash HEAD~2
6a18939264224d6ce24c7426249edfea
$ ob internal hash HEAD~3
554c459ff2cced3e1b11241ed403890d
```

```
$ git show 7d0b41b50870bebb55d41ec41956647ce35804a2 | tail -3
diff --git a/migration/554c459ff2cced3e1b11241ed403890d-3543ea607791bc2e19419d22918823bd/obelisk-upgrade b/migration/554c459ff2cced3e1b11241ed403890d-3543ea607791bc2e19419d22918823bd/obelisk-upgrade
new file mode 100644
index 0000000..e69de29
```

3noch:
Oops. I think that PR was submitted prior to the migration changes. Perhaps we could disable non-fast-forward merges?

JBetz:
Possibly related. This is after installing the latest version of master.
```
$ ob upgrade
✖ Deciding whether to handoff
ob: MultipleTerminalVertices
```

srid:
@JBetz Yes, that would be an (expected) symptom of this issue.

JBetz:
Is this going to be fixed anytime soon? And in the meantime, what's the best way to switch an existing project to a different version of obelisk?

luigy:
Closing this since migrations have been deferred for now

PR #117, opened 2018-07-12
read files strictly in "Obelisk.Migration".

### LINKED ISSUE #115: hydra build is currently failing
Is this expected? I am already looking into it, whether I can find an obvious reason. I can reproduce the issue locally (obviously - thank you nix) and get a `too many open files` error:

```
ob: ./migration/f595b01939b4c87b7c146ea9ff331b39-4da182e21be35feeee484dc85be7a075/obelisk-upgrade: openFile: resource exhausted (Too many open files)
```

eskimor:
To avoid "too many open files" error.

This fixes #115 partly, I still get an error that is similar to the one on hydra:

```
No vertex found for HEAD (b0b562b83e8a198bb858deeb19d5868c)
✖ Checking existence of HEAD vertex
```

### DIFF
--- a/lib/migration/src/Obelisk/Migration.hs
+++ b/lib/migration/src/Obelisk/Migration.hs
@@ -17,6 +17,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Text (Text)
 import qualified Data.Text as T
+import qualified Data.Text.IO as T
 import System.Directory
 import System.FilePath
 
@@ -66,10 +67,10 @@ instance Exception MissingEdgeInActionMap
 instance Exception NonLinearGraph
 
 class Action action where
-  parseEdgeMeta :: String -> action
+  parseEdgeMeta :: Text -> action
 
 instance Action Text where
-  parseEdgeMeta = T.pack
+  parseEdgeMeta = id
 
 -- | Get the directed-acylic graph of migration
 -- TODO: Consider calling getDag only when reading the graph for efficiency.
@@ -95,7 +96,7 @@ readGraph root name = doesDirectoryExist root >>= \case
       let f = root </> (T.unpack $ v1 <> "-" <> v2) </> T.unpack graph
       doesFileExist f >>= \case
         True -> do
-          c <- readFile f >>= pure . parseEdgeMeta
+          c <- parseEdgeMeta <$> T.readFile f
           return $ Just c
         False -> do
           return Nothing


srid:
> I still get an error that is similar to the one on hydra
> No vertex found for HEAD (b0b562b83e8a198bb858deeb19d5868c)

Yea, see #116

eskimor:
thanks!

lib/migration/src/Obelisk/Migration.hs
```diff
@@ -240,3 +240,8 @@ writeEdge dir actionFiles v1 v2 = do
       forM_ actionFiles $ \fp -> liftIO $
         writeFile (edgeDir </> T.unpack fp) ""
       pure $ Just edgeDir
+
+readFileStrict :: FilePath -> IO String
+readFileStrict = makeStrict <=< readFile
```
3noch: [Text.IO.readFile](https://hackage.haskell.org/package/text-1.2.3.0/docs/Data-Text-IO.html#v:readFile) is strict and seems more appropriate here?

ryantrinkle:
Yes; we should always be using Text anyway

lib/migration/src/Obelisk/Migration.hs
eskimor: Yep. I already had it checked out. Will fix.

lib/migration/src/Obelisk/Migration.hs
```diff
@@ -95,7 +95,7 @@ readGraph root name = doesDirectoryExist root >>= \case
       let f = root </> (T.unpack $ v1 <> "-" <> v2) </> T.unpack graph
       doesFileExist f >>= \case
         True -> do
-          c <- readFile f >>= pure . parseEdgeMeta
+          c <- readFileStrict f >>= pure . parseEdgeMeta
```
eskimor: but if we use T.readFile - what to do with parseEdgeMeta - just drop it? At the moment it is just a T.pack anyway, but I assume this is intended to do more at some point? Should I preserve the `Action` class and change it to:

```haskell
class Action action where
  parseEdgeMeta :: Text -> action

instance Action Text where
  parseEdgeMeta = id
```

eskimor: @srid ?

srid: @eskimor Yes, preserve but change it to use `Text` and `id` like you've written above.

eskimor: awesome!

eskimor:
Requested improvements are incorporated now! :-)

kmicklas:
Can we just use [`readFile` from `Data.Text.IO`](http://hackage.haskell.org/package/text-1.2.3.0/docs/Data-Text-IO.html#v:readFile)? It's explicitly strict and avoids manually `pack`ing.

eskimor:
> Can we just use readFile from Data.Text.IO? It's explicitly strict and avoids manually packing.

Yeah, thanks for reviewing! The latest version of this PR already does this. Are you somehow reviewing an older version or is something messed up with my PR?

kmicklas:
@eskimor Oops, I must have clicked on the first commit by mistake.

PR #175, opened 2018-07-21
Provide nix-shells with complete dev environment.

eskimor:
The current `nix-shell` commands assumed that the user has `ghcid` and
`cabal-install` installed at the right versions. This commit provides
"dev" variants that include these tools.


eskimor:
Test suit fails because of migration stuff, are there some docs available about what those are about and what I have to do to keep them ok? Thanks in advance!

PR #177, opened 2018-07-25
Replace MonadThrow with domain specific MonadError for Obelisk.Migration

### LINKED ISSUE #99: Use MonadError instead of MonadThrow in ob upgrade code
http://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Except.html

The errors [here](https://github.com/obsidiansystems/obelisk/pull/93/files#diff-9296fe9dc90366d27ba82bcba1f180f3R60) will only be defined together in this module so it is effectively a closed type of errors. Using MonadError surfaces this static guarantee whereas MonadThrow allows the possible exceptions to form an open type, obscuring the error conditions in this code.

kmicklas:
Fixes #99.

### DIFF
--- a/lib/command/src/Obelisk/Command/Upgrade.hs
+++ b/lib/command/src/Obelisk/Command/Upgrade.hs
@@ -8,7 +8,8 @@
 module Obelisk.Command.Upgrade where
 
 import Control.Monad (forM, unless, void)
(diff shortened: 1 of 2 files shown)


ryantrinkle:
Thanks!

PR #178, opened 2018-07-25
temporarily use obsidian monoidal-containers fork as override

hSloan:


default.nix
```diff
@@ -68,6 +68,15 @@ let
         rev = "480a73137e9b38ad3f1bc2c628847953d2fb3e25";
         sha256 = "0dpwi5ffs88brl3lz51bwb004c6zm8ds8pkw1vzsg2a6aaiyhlzl";
       }) {});
+
+		monoidal-containers =
```
ryantrinkle: Looks like hard tabs crept in

hSloan: Ah, sneaky critters...

Issue #180, opened 2018-07-25
Debug messages aren't readable on consoles with light backgrounds


ali-abrar:
https://github.com/obsidiansystems/obelisk/blob/0827f9bf7ad0132f4e57b3eec2b76ac0079dffe1/lib/cliapp/src/Obelisk/CliApp/Logging.hs#L149

PR #181, opened 2018-07-26
 Logging: Don't set white foreground color

ali-abrar:
#180 

I changed it to set the intensity but refrain from setting the color. Fixes the issue where the text is completely unreadable, but probably isn't as pretty as originally intended. Improvements welcome.


3noch:
Setting *any* colors directly is probably wrong for the same reason. (Although I know no one who has a vivid yellow terminal background...).

3noch:
Or we could expose an env var for specifying colors, as other utilities do.

PR #187, opened 2018-07-30
Changed command lines in "contributing" to include build tools

eskimor:
in spawned nix-shell.

This way the developer does not need to have cabal-install and ghcid in correct versions installed.


ryantrinkle:
I think we should make these simpler before recommending them

eskimor:
Will work something out!

eskimor:
Done - commands are as simple as before.

luigy:
Thanks!

PR #196, opened 2018-08-12
Move executables to src-bin

ali-abrar:
When the executables are in the same directory as the library, we
have to set hs-source-dirs to the library's source-dir, causing
the library to recompile when you build the executable and requiring
that the library's dependencies be re-listed as build-depends of the
executable.


ali-abrar:
Thanks @madelinja for figuring this one out.

PR #198, opened 2018-08-13
Fix command runtime nix dep

luigy:
ob requires at least nix 2.0, but it was being wrapped with an older
version. This gets rid of having to manually set NIX_REMOTE for systems
running on 2.0

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,7 @@ let
     git
     gitAndTools.hub
     nix-prefetch-git
-    nixStable
+    nixUnstable
     openssh
   ];
 


default.nix
```diff
@@ -21,7 +21,7 @@ let
     git
     gitAndTools.hub
     nix-prefetch-git
-    nixStable
+    nixUnstable
```
luigy: Or maybe it should not even manually wrap any `nix` in here since whatever version the user is running would be more correct to run against the `nix-daemon`

ElvishJerricco: I agree with that. You should definitely be using the version they have installed, or else you risk miscommunication with the daemon

3noch:
Hm are you sure it's because `nixStable` is old? Opening `nix repl '<nixpkgs>'` and running `nixStable` gives me nix 2. I don't think `nixUnstable` is actually preferable since I *think* that refers to an unofficial release of nix? Also, since obelisk is pinned to a version of nixpkgs, this `nixStable` can only refer to the version of nix in the pin. If removing this fixes an issue, great, but I'm a bit suspicious of the diagnosis.

3noch:
Oh I just realized that the *pin* itself is still on nix 1.11! Yes that does seem problematic. *However* I still don't think using `nixUnstable` is the cure, but using `nixStable` from a nixpkgs that gives us nix 2.

default.nix
3noch: Hm I haven't encountered this directly. For example, `nix-shell -p nixStable2` on NixOS 17.09 (still using nix 1) works fine and even nested shells. However, I'm no expert on how the daemon works so I could see that being a problem.

ElvishJerricco: Opening a nix-shell isn't exactly a comprehensive test of all the nix-daemon's capabilities :P I've had breakage in corner cases between 2.0.x bumps, so I'm cautious to use different versions in any scenario. There's nothing stopping them from making major breaking changes to the daemon's protocol in the future, as I don't think it's considered public API.

3noch: Yeah if there's legitimate concern then I'm all for avoiding issues.

ryantrinkle:
The real fix here is to get reflex-platform's reunification branch up to date asap; it's getting pretty close.

PR #203, opened 2018-08-22
[WIP] Allow users to pass additional backend handlers

imalsogreg:
This adds a configuration field to `BackendConfig`, for users to pass arbitrary snap handlers alongside the ones that manage serving cached javascript artifacts.

My goal is to be able to serve an HTTP api from the same backend application that serves the obelisk frontend assets. Is my change to `obelisk-backend` the right approach? Or does it conflict with the cross-platform nature of `obelisk`? If this is the wrong approach, is there another preferred way of mixing obelisk apps with general API servers?


srid:
@imalsogreg Checkout #112 to see where obelisk is headed towards. That branch supports specifying your own backend handlers ([see example](https://github.com/srid/revue/blob/master/backend/src/Backend.hs#L44-L62)).

imalsogreg:
Thanks @srid, perfect. That routes branch is really something!

PR #205, opened 2018-08-24
Don't wrap nix tools for ob command

### LINKED ISSUE #204: Consider not pinning nix for ob commands
Obelisk currently uses a pinned `nix` (et al). From various reports (#52, #100, #198) it seems this may not be the best idea after all. Consider dropping the pin entirely.

luigy:
Closes #204 
Might help some cases of #52

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,6 @@ let
     git
     gitAndTools.hub
     nix-prefetch-git
-    nixUnstable
     openssh
   ];
 


default.nix
```diff
@@ -20,8 +20,6 @@ let
     coreutils
     git
     gitAndTools.hub
-    nix-prefetch-git
```
3noch: This should stay.

luigy: Oh I thought it was also bundled as port of a nix installation

ElvishJerricco: Wait till you realize Hydra ships with its own copy of nix-prefetch-git from like 2012...

Issue #193, opened 2018-08-09
Some thoughts on cutting down on Docker image sizes


dalaing:
Hi again :)

I've been playing around with Obelisk and docker, initially via a Dockerfile and more recently via this bit of nix in the same directory as a fresh `ob init`: 

```
# to use this:
# > nix-build docker.nix -o result-docker
# > docker load --input ./result-docker
# > docker run -p 8000:8000 ob-test

{ nixpkgs ? import <nixpkgs> {} }:
let
  ob-test = (import ./. {}).exe;
in
 nixpkgs.pkgs.dockerTools.buildImage {
    name = "ob-test";
    contents = ob-test;
    config = {
      Cmd = [ "${ob-test}/backend" ];
      ExposedPorts = {
        "8000/tcp" = {};
      };
    };
 }
```

I noticed that the image size of the default Obelisk project (and hence the closure size of the exe derivation) is pretty big - it weighs in at 603Mb.

I had a poke around and it looked like the bulk of that was coming from the `backend`, which uses the static renderer to create the head section of the page.  This brings in the GHC build of `reflex-dom`, and via that Webkit, and via that a big pile of X libraries.

So I played around a little [here](https://github.com/dalaing/obelisk/commit/55dae3dea4c3dea922ba5e36f014ba9dfefedb4f) and made some alterations so that the head section could be pregenerated.  It seems to work with both `ob run` (and the associated ghcid reloading) and with the `exe` derivation, which is nice.

Initially it only saved a 1Mb, until I removed `obelisk-executable-config-inject` from `skeleton/backend/backend.cabal`, which also uses the GHC build of `reflex-dom`.  If there is someway to pregenerate that stuff, then we can get the size of the output down.  At that point we're at 85.5MB.

I poked around some more, and saw that `nodejs` and a few friends was in there, so I made [this](https://github.com/dalaing/obelisk/commit/55dae3dea4c3dea922ba5e36f014ba9dfefedb4f#diff-5712e736e0de6ba170577f8472c398e9L192) ultra dodgy change, and the output is now at 46.2Mb.

This has all been done in horrendous hack mode in a short amount of time, but I thought I'd post it in here in case it seems useful / gives people ideas / is an interesting discussion point.

ryantrinkle:
Wow, nice sleuthing! I think that indicates we should definitely split
different reflex-dom backends into their own packages, so this stuff can
simply import the right one.

ElvishJerricco:
[...]

PR #208, opened 2018-08-29
obelisk-executable-config-inject: Use reflex-dom-core

3noch:
This project doesn't need all of reflex-dom, just reflex-dom-core.

This makes the dependency closure much smaller.


3noch:
cc @dalaing this helps closure size a lot, as per the discussion in #193.

3noch:
@luigy Thoughts?

luigy:
Thanks!

PR #209, opened 2018-09-02
Update skeleton with routes

ali-abrar:


ali-abrar:
Added the changes necessary for nix-build -A exe

ali-abrar:
@kmicklas yeah, it can be. I just don't know what to call it yet.

PR #210, opened 2018-09-02
Remove skeleton/devel

ali-abrar:
This relies on #209 

I don't particularly like the .ghci stuff [here](https://github.com/obsidiansystems/obelisk/compare/routes...aa-no-devel?expand=1#diff-0b3c81348165c4d2fe2d548ffa2d346aR101). Open to other ideas.


ryantrinkle:
The `.ghci` stuff doesn't seem any worse, conceptually, than before.  This definitely looks like an improvement to me.

PR #213, opened 2018-09-09
Quickfix: Unused import

sectore:
```[-Wunused-imports]
    The import of ‘Common.Api’ is redundant```


ali-abrar:
Looks like this was fixed in another commit. Thanks!

PR #214, opened 2018-09-11
Rename generated static module and pkg

luigy:
from `Static` to `Obelisk.Generated.Static`


lib/asset/manifest/src/Obelisk/Asset/Promoted.hs
```diff
@@ -32,11 +33,15 @@ data StaticConfig = StaticConfig
 
 writeStaticProject :: Map FilePath FilePath -> FilePath -> StaticConfig -> IO ()
 writeStaticProject paths target cfg = do
-  createDirectoryIfMissing True $ target </> "src"
+  createDirectoryIfMissing True target
   T.writeFile (target </> T.unpack (_staticConfig_packageName cfg) <.> "cabal") $ staticCabalFile cfg
   let modName = _staticConfig_moduleName cfg
+      (modName', moduleDirPath) = case L.uncons (reverse $ T.splitOn "." modName) of
+        Nothing -> error $ "writeStaticProject: invalid module name " ++ T.unpack modName
```
alexfmpe: `<>` ?

luigy: :+1:

PR #216, opened 2018-09-17
Expose obelisk server modules as functions

3noch:
This refactors the obelisk server as two nixos modules that are exposed as module-making-functions. This allows projects to build entirely custom servers out of obelisk's pieces and still work with `ob deploy`.


3noch:
@luigy Thoughts?

PR #219, opened 2018-09-21
Mobile config fixes

luigy:


default.nix
```diff
@@ -279,6 +279,12 @@ rec {
                     ${backendName} = addBuildDepend super.${backendName} self.obelisk-run;
                   };
                   totalOverrides = composeExtensions (composeExtensions defaultHaskellOverrides projectOverrides) overrides;
+                  inherit (nixpkgs) lib;
+                  inherit (lib.strings) hasPrefix;
+                  privateConfigDirs = ["config/backend"];
+                  injectableConfig = builtins.filterSource (path: _:
```
3noch: AFAIK `filterSource` doesn't work on nix paths. Do we really not support config paths produced by a nix derivation?

3noch: After talking to @ryantrinkle I now realize we mean for configs to be secret (in general) so we should definitely not support nix paths.

PR #223, opened 2018-09-26
Add support for hpack

kmicklas:


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -68,31 +71,46 @@ getLocalPkgs = pure ["backend", "common", "frontend"]
...
+  => FilePath -- ^ package directory
   -> m (Maybe CabalPackageInfo)
-parseCabalPackage cabalFp = do
-  exists <- liftIO $ doesFileExist cabalFp
-  if exists
+parseCabalPackage dir = do
+  let hpackFp = dir </> "package.yaml"
+      cabalFp = dir </> (takeBaseName dir <> ".cabal")
+  hasHpack <- liftIO $ doesFileExist hpackFp
+  mCabalContents <- if hasHpack
```
ryantrinkle: Are we sure that hpack should take precedence over cabal? I think nixpkgs might do the opposite. @luigy

kmicklas: @ryantrinkle Good point. I'm not sure I like that default but consistency with `nix-build` is important. Fix in https://github.com/obsidiansystems/obelisk/pull/224.

PR #225, opened 2018-09-27
Use hpack for skeleton

kmicklas:
This removes the need to keep `exposed-modules` up to date.


ElvishJerricco:
It looks like [cabal will support globbing in an upcoming version](https://github.com/haskell/cabal/pull/5284), and it already has common stanzas. I think hpack should be avoided. It's a 100% unnecessary point of failure that adds confusion to the file hierarchy of the project.

kmicklas:
@ElvishJerricco Whoa, I didn't realize Cabal finally supported that. One thing that still appears to be missing is having a common stanza included from an external file. That would be super useful, for example, for maintaining a set of `default-extensions` shared across the packages in the project.

ElvishJerricco:
@kmicklas I'd consider that minor enough that it's not worth the tremendous risks of complicating the tooling. Plus I'm not a fan of `default-extensions` to begin with ;)

kmicklas:
@ElvishJerricco We'd like to use `default-extensions` to make it easier for beginners to use, not having to worry about understanding the risks of each and maintaining them per-file. In general I prefer to use it as well, as, for example, putting `ScopedTypeVariables` in every file is just needless verbosity. I leave the per-file declarations for "dangerous" extensions (e.g. `UndecidableInstances`, `TemplateHaskell`) which now stick out much more easily. I'm curious why you don't like this feature?

I'm also not convinced the tooling complexity is really an issue here since we have pretty deep control over the build infrastructure of an Obelisk project.

On the other hand, after testing I've discovered that sharing configuration across packages doesn't actually work since it will be outside the package root for the purposes of `cabal2nix`. (There may be a workaround for this but I'm closing this PR for the time being.)

ElvishJerricco:
@kmicklas I don't like `default-extensions` because it means I can't visit a file and know what language I'm using. That said I'm not suggesting `default-extensions` should be avoided; it's just not my preference.

I would strongly suggest `hpack` be avoided though. It doesn't matter if it's the user or obelisk taking on the complexity; it's complexity nonetheless, so it adds points of failure that will be ever more opaque and surprising to the user. Using `hpack` is an easy way to buy build failures that are 100% avoidable.

PR #227, opened 2018-09-27
Require extra flag before overwriting on ob init

tomsmalley:
Assumes that checking for existence of ".obelisk" is sufficient to know that an obelisk project already exists

### DIFF
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -87,8 +87,11 @@ initSource = foldl1 (<|>)
   , InitSource_Symlink <$> strOption (long "symlink" <> action "directory" <> metavar "PATH")
   ]
 
+initForce :: Parser Bool
+initForce = switch (long "force" <> help "Allow ob init to overwrite files")
+
 data ObCommand
-   = ObCommand_Init InitSource
+   = ObCommand_Init InitSource Bool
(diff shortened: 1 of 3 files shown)


3noch:
Note that `.obelisk` may also exist in a parent directory, in which case obelisk uses that. In this case `ob init` doesn't do anything destructive, but it may be unexpected. It might be worth adding a warning in this scenario, e.g. "WARNING: Initializing a new obelisk project inside another one.".

3noch:
Sorry, accidentally closed.

tomsmalley:
Good point - now uses `findProjectRoot`. I think we should probably error out by default in those cases, I can't imagine this being a common occurrence - can still override with `--overwrite`. Will add tests for this to `selftest` tomorrow.

3noch:
Ah, nice. In that case `--force` might be better since you're not overwriting anything in the latter case.

tomsmalley:
Yes, a separate flag makes sense here.

luigy:
does this cover a folder with files that overlap obelisk's skeleton if .obelisk does not exist?

3noch:
I meant rename `--overwrite` to `--force` to work for both better.

3noch:
@luigy Oh I forgot that `ob init` writes a lot more than `.obelisk`. Yes that would be ideal to cover.

PR #228, opened 2018-09-28
Flexible backend port

3noch:


luigy:
This is better than before, but now I wonder where is the `--port` parsing actually happening? `obelisk-backend` is aiming to be backend agnostic, but not quite there yet

luigy:
this may just be a leftover that is now fully handled by `config/common/route`

3noch:
@luigy The port parsing is being done by snap. The previous behavior was to use the default port and implicitly imply that it was 8000. This forces you to pick the same port for the server and the `proxyPass`. But yes ideally we'd have a better, backend-agnostic way.

3noch:
Note that *not* specifying a port (i.e. the previous way) is also not backend agnostic. :)

luigy:
Oh yeah I just meant to note it as an improvement  :)

3noch:
Righteo!

PR #231, opened 2018-10-01
default.nix: make JS optimization level configurable

### LINKED ISSUE #229: Make closure compiler level configurable
The closure compiler's ADVANCED optimizations causes problems when using external JS libs. It's been an issue for me when trying to use [reflex-dom-ace](https://github.com/reflex-frp/reflex-dom-ace). It worked fine when using `ob run`, but not at all when I built the frontend with `ghcjs`.

vaibhavsagar:
Fixes #229. We also need this for one of our projects because of weird interactions between this setting and the ACE editor.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -158,11 +158,11 @@ in rec {
     obelisk-asset-manifest-generate "$src" "$haskellManifest" ${packageName} ${moduleName} "$symlinked"
   '';
 
-  compressedJs = frontend: pkgs.runCommand "compressedJs" { buildInputs = [ pkgs.closurecompiler ]; } ''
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -1,6 +1,7 @@
 { system ? builtins.currentSystem
 , profiling ? false
 , iosSdkVersion ? "10.2"
+, jsOptimizationLevel ? "ADVANCED"
```
ryantrinkle: Shouldn't this be a project setting, not an Obelisk setting?

vaibhavsagar:
This doesn't work yet, even though the checks pass.

ryantrinkle:
Please mark "WIP"

luigy:
Thanks!

PR #233, opened 2018-10-02
Minor routes fixes

tomsmalley:


tomsmalley:
These are things I ran into when attempting to reproduce the iOS route bug in the skeleton

ryantrinkle:
@tomsmalley Thanks!  I ended up writing the SetRoute instance, too, but the fundep looks like it would be great to add.

PR #237, opened 2018-10-10
Obelisk.CliApp.Process: reconstruct commands on display

vaibhavsagar:
This makes them easier to copy and debug when things go wrong.

### DIFF
--- a/lib/cliapp/src/Obelisk/CliApp/Process.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Process.hs
@@ -20,7 +20,7 @@ module Obelisk.CliApp.Process
 
 import Control.Applicative (liftA2)
 import Control.Monad (void, (<=<))
-import Control.Monad.Catch (Exception, MonadMask, throwM)
+import Control.Monad.Catch (Exception(..), MonadMask, throwM)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import qualified Data.ByteString.Char8 as BSC
 import Data.Function (fix)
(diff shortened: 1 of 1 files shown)


3noch:
I've wanted this too! Thanks! When a raw command gets reconstructed, it might have very different semantics to what was actually run. Better would be to quote each argument and apply escaping.

3noch:
This might help: https://github.com/obsidiansystems/obelisk/blob/495f20ccf033a7700fb8afaca40229dedd9d161b/lib/command/src/Obelisk/Command/Utils.hs#L65

ryantrinkle:
Let's escape the arguments when they need it. Also, why not keep CmdSpec in
the error datatype?

vaibhavsagar:
Thanks @3noch, I've duplicated that function because `obelisk-cliapp` is a dependency of `obelisk-command` so I couldn't import it. @ryantrinkle I did that primarily to avoid writing a `Show` instance but that's not a good reason so I've changed it back.

ryantrinkle:
The human-readable output shouldn't be produced by `show`. We should have
another function (or class)

vaibhavsagar:
Okay, I've instead defined `displayException` to provide the expected behaviour (that an error message contains a command that the user can copy and reuse as-is).

luigy:
Thanks! would it be more helpful if it also printed the absolute path of the exe?

ryantrinkle:
@luigy Hmm, is the PATH usually different from the ambient PATH?

ryantrinkle:
(@luigy regardless, I think we don't have to hold up this merge for that)

luigy:
Agreed, this is already an improvement over what we currently have

luigy:
thanks @vaibhavsagar

vaibhavsagar:
My pleasure :smile:!

PR #243, opened 2018-10-16
More MTList fixes

Ericson2314:
1. Adding more missing commuting instances, and MonadTrans instances

2. Make `failWith` impose a `MonadError` constraint rather than directly kill the app.

The latter can be used with a judicious `runExceptT` to catch failures and turn them into warnings.

### DIFF
--- a/lib/cliapp/src/Obelisk/CliApp/Logging.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Logging.hs
@@ -7,12 +7,29 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 -- | Provides a logging handler that facilitates safe ouputting to terminal using MVar based locking.
 -- | Spinner.hs and Process.hs work on this guarantee.
-module Obelisk.CliApp.Logging where
+module Obelisk.CliApp.Logging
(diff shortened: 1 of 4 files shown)


lib/cliapp/src/Obelisk/CliApp/Logging.hs
```diff
@@ -107,21 +125,13 @@ handleLog' noColor output = do
...
 putLogRaw :: Cli m => Severity -> Text -> m ()
 putLogRaw sev = logMessage . Output_LogRaw . WithSeverity sev
 
 -- | Like `putLog Alert` but also abrupts the program.
-failWith :: (MonadIO m, Cli m) => Text -> m a
-failWith s = do
-  putLog Alert s
-  liftIO $ exitWith $ ExitFailure 2
+failWith :: MonadError Text m => Text -> m a
+failWith = throwError
```
srid: This is missing a `putLog Alert`, no?

srid: nevermind - just noticed that you defined a `MonadError` instance that does it. cool.

PR #246, opened 2018-10-18
Improved error messages

Ericson2314:
Random improvements that got caught in the mac deploy dragnet.

PR #247, opened 2018-10-18
Redo much of `ob thunk`, using `git ls-remote` and not using `hub`

Ericson2314:
All hunks are now unpacked the same way. Github thunks are attempted first with `nix-prefetch-url` and then with `nix-prefetch-git`. Authentication is now the responsibility of the user. (`ssh-agent` would be recommended.)

Depends on https://github.com/obsidiansystems/obelisk/pull/246


lib/cliapp/src/Obelisk/CliApp/Process.hs
```diff
@@ -9,25 +9,29 @@
...
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BSC
 import Data.Function (fix)
 import Data.Monoid ((<>))
+import Data.Text (Text)
 import qualified Data.Text as T
-import Data.Text.Encoding (decodeUtf8)
-import qualified Data.Text.IO as T
+import qualified Data.Text.Encoding as T
+--import qualified Data.Text.IO as T
```
luigy: stray comment

PR #248, opened 2018-10-18
 Ignore query params for static routes 

tomsmalley:


alexfmpe:
`pathOnlyEncoderIgnoringQuery` sounds odd/redundant to me. It's not clear from the name which one throws and which one is safe. Maybe add `safe`/`unsafe` suffix(es) and/or make the safe one the default?

ryantrinkle:
We do want to have a path only encoder that is strict. Note that throwing
is expected behavior from the parse side of things generally. However, I
might have gotten default intuitions wrong, so we can definitely move
things around

luigy:
Thanks!

PR #253, opened 2018-10-18
Use Haskell overlays, temporarily expose __useLegacyCompilers

Ericson2314:


default.nix
```diff
@@ -1,20 +1,13 @@
...
-  pkgs = nixpkgs;
-in with pkgs.haskell.lib; with pkgs.lib;
-let
-  # TODO: Remove this after updating nixpkgs: https://github.com/NixOS/nixpkgs/issues/37750
-  justStaticExecutables' = drv: let
-      drv' = justStaticExecutables drv;
-    in if pkgs.stdenv.isDarwin
-      then removeConfigureFlag drv' "--ghc-option=-optl=-dead_strip"
-      else drv';
+  cleanSource = builtins.filterSource (name: _: let baseName = builtins.baseNameOf name; in !(
```
ElvishJerricco: Btw, it's better to use `lib.cleanSourceWith` than `builtins.filterSource`. Does the same thing, but it composes with itself, whereas builtins.filterSource cannot.

ElvishJerricco: This can actually be important if you want to filter a source provided by a user, because the user may want to pass you a source they've already filtered.

Ericson2314: Good point. I think I'll do that in a follow-up PR if you don't mind? (this is just moving it around.)

For one, I need a better store for the inner overlay (haskell) accessing the outer `self` (`pkgs`) than this nasty `callPackage` hack!

PR #260, opened 2018-10-19
Test updates

tomsmalley:
still waiting for all tests to actually finish, but this works for the most part


tomsmalley:
@luigy tests now pass

luigy:
Thanks!

PR #262, opened 2018-10-21
Fix build with no assets

matthewbauer:
If no assets exist, obelisk failing in asset-manifest for unclear
reasons. It looks like symlinked just needs to be created
automatically.


luigy:
Thanks!

PR #267, opened 2018-10-22
Fix unpacking of thunk with non-default branch

Ericson2314:
Testing needed


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -582,12 +582,11 @@ unpackThunk' noTrail thunkDir = readThunk thunkDir >>= \case
         git $ [ "clone" ]
           ++  ("--recursive" <$ guard (_gitSource_fetchSubmodules s))
           ++  [ T.unpack $ render $ _gitSource_url s ]
+          ++  do branch <- maybeToList $ _gitSource_branch s
```
luigy: 👀

PR #268, opened 2018-10-22
Fix profiling build

Ericson2314:
Hydra jobs for profiling will be added later once reflex-platform has them too.


luigy:
Thank!

PR #285, opened 2018-10-29
Preload all.js

tomsmalley:


luigy:
https://github.com/obsidiansystems/obelisk/blob/master/lib/run/src/Obelisk/Run.hs#L152 jsaddlejs should also use this

luigy:
Thanks!

PR #286, opened 2018-10-29
Add RouteToUrl

ali-abrar:
Depends on #282 

Some machinery to support `routeLink`, a link widget that, when clicked, sets the route to the provided route. In non-javascript contexts, this widget falls back to using `href`s to control navigation.

### DIFF
--- a/lib/route/src/Obelisk/Route/Frontend.hs
+++ b/lib/route/src/Obelisk/Route/Frontend.hs
@@ -41,6 +41,11 @@ module Obelisk.Route.Frontend
   , SetRoute(..)
   , runSetRouteT
   , mapSetRouteT
+  , RouteToUrl(..)
+  , RouteToUrlT(..)
+  , runRouteToUrlT
+  , mapRouteToUrlT
+  , routeLink
   ) where
 
 import Prelude hiding ((.), id)
@@ -61,6 +66,8 @@ import Data.Constraint (Dict (..))
 import Data.Dependent.Sum (DSum (..))
 import Data.GADT.Compare
 import Data.Monoid
+import Data.Proxy
+import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Functor.Compose
 import Reflex.Class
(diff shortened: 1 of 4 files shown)


ali-abrar:
`routeLink` links can also be middle-clicked and should work correctly.

srid:
@ali-abrar Can `routeToUrl` be used to encode _backend_ routes from the frontend code? The types seem to suggest that this is impossible (see below), but I do wonder if I'm missing something.

```
    • Couldn't match type ‘Route’ with ‘BackendRoute’
        arising from a functional dependency between constraints:
          ‘RouteToUrl (DSum BackendRoute Identity) m’
            arising from a use of ‘askRouteToUrl’
            at frontend/src/Frontend.hs:85:22-34
          ‘RouteToUrl (R Route) m’
            arising from a type expected by the context:
                           forall js t (m :: * -> *).
                           ObeliskWidget js t route0 m =>
                           RoutedT t route0 m ()
```

ali-abrar:
> Have you experimented with pages where you've scrolled down and then clicked one of these links? In my experience, the JS-driven "page change" does not scroll to the top and must be done manually.

If this is the case, please open an issue.

ali-abrar:
@srid This might be what you're looking for: https://github.com/obsidiansystems/obelisk/blob/develop/lib/route/src/Obelisk/Route.hs#L838-L864

lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -337,6 +423,26 @@ runRouteViewT routeEncoder a = do
...
+  => R route -- ^ Target route
+  -> m a -- ^ Child widget
+  -> m a
+routeLink r w = do
+  enc <- askRouteToUrl
+  let cfg = (def :: ElementConfig EventResult t (DomBuilderSpace m))
+        & elementConfig_eventSpec %~ addEventSpecFlags (Proxy :: Proxy (DomBuilderSpace m)) Click (\_ -> preventDefault)
+        & elementConfig_initialAttributes .~ "href" =: enc r
+  (e, a) <- element "a" cfg w
+  setRoute $ r <$ domEvent Click e
```
3noch: You may need to explicitly `scrollTo(0)` here as well. Has this been tested on links very far down the page?

srid:
> If this is the case, please open an issue.

@ali-abrar I am currently experiencing this issue on a project of mine. I'll try to create a simple repro project and open an issue. Or I'll play with Elliot's suggestion and see if I can fix it.

lib/route/src/Obelisk/Route/Frontend.hs
kmicklas: Isn't this going to be a general issue with `setRoute` and not just this code?

3noch: I'm presuming that not all `setRoute` use cases actually want to mimic a full "page load". But here it's quite obvious that you do want it.

alexfmpe: `setRouteWithOptions` ?

srid: I don't think we should _unconditionally_ `scrollTo` near `setRoute` or inside `routeLink`. It should be an option to be enabled or disabled. 

I have at least one use case where I would want my app to control where to scroll to. 

Here's [an example](https://github.com/srid/Taut/pull/24/files) where the app does automatic scrolling by processing a _top-level_ dynamic -- basically we either scroll to top (most routes) or to a specific element in the body (if a specific permalink route is active) -- and this would be messy to implement if obelisk was automagically scrolling for us.

3noch: Definitely not `setRoute` but possibly `routeLink` since it is actually designed to work without JavaScript. I.e. the behavior with and without JS is supposed to be the same.

PR #291, opened 2018-10-31
Don't put configs into the nix store

tomsmalley:
This breaks `nix-build --attr exe` because the backend expects the
config folder to be in the result

### DIFF
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -101,7 +101,6 @@ deployPush deployPath getNixBuilders = do
             , strArg "adminEmail" adminEmail
             , strArg "routeHost" routeHost
             , boolArg "enableHttps" enableHttps
-            , rawArg "config" $ deployPath </> "config"
             ]
           , _nixBuildConfig_builders = builders
           }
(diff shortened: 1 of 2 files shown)


3noch:
Would it help to just put an empty `config` folder in the result?

tomsmalley:
@3noch No, the breakage is only revealed when you run the app: no config values are injected. Perhaps I shouldn't have said it breaks `nix-build`, but rather the documented method of running local builds.

3noch:
I see. And this is only true when the app requires the config files to exist for it to startup. Perhaps the default behavior should be to fallback on CLI args, Env vars, and/or defaults.

eskimor:
How does this play together with overriding config values on deployment? (config directory in deploy/..) ?

luigy:
> How does this play together with overriding config values on deployment? (config directory in deploy/..) ?

Yes, that's correct, the `config` should had always come from the deploy's `config` directory when running `ob deploy push`

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -122,6 +121,14 @@ deployPush deployPath getNixBuilders = do
     callProcess'
       (Map.fromList [("NIX_SSHOPTS", unwords sshOpts)])
       "nix-copy-closure" ["-v", "--to", "root@" <> host, "--gzip", outputPath]
+  withSpinner "Uploading config" $ ifor_ buildOutputByHost $ \host _ -> do
+    callProcessAndLogOutput (Notice, Warning) $
+      proc "rsync"
+        [ "-e ssh " <> unwords sshOpts
+        , "-qarvz"
```
luigy: quiet & verbose? :P let's use the long option names to make it easier to read when coming back to this. This should also have `--delete` to ensure removed configs in the deploy directory get removed in the deployment targets

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -122,6 +121,14 @@ deployPush deployPath getNixBuilders = do
...
+        , deployPath </> "config"
+        , "root@" <> host <> ":/var/lib/backend"
```
luigy: is there a way to query for this path?

ryantrinkle:
Generally, the things I've put in config would not make any sense to have defaults or any kind of overriding semantics.  Are there examples of things like that?

PR #293, opened 2018-10-31
`staticFilesImpure` and not polluted assets library

Ericson2314:
The first commit exposes a `staticFilesImpure` to do things other than `toString staticFiles`. The second commit cleans things up so this incidental path doesn't pollute all the hashes (especially important with prod builds).


Ericson2314:
Not quite finished yet https://hydra.reflex-frp.org/eval/85748?compare=master#tabs-unfinished

ryantrinkle:
@Ericson2314 @luigy Even if this fully works, I'm not sure it should be merged to master.  Let's make sure the interface we're exposing is clean enough that we're willing to support it for a while before we merge anything.

Ericson2314:
So the second commit purely cleans things up and should definitely get in. The first commit exposes `staticFilesImpure`, which is the one new interface to think about.

Ericson2314:
@ryantrinkle https://github.com/obsidiansystems/obelisk/pull/294/files contains just the second commit, which solves the deploy purity problem, without exposing any new interfaces. (`staticFilesImpure` is now let bound.)

Ericson2314:
@ali-abrar This should still work just fine.

Ericson2314:
Other plans supersede this.

PR #294, opened 2018-10-31
Prevent impure path from polluting builds

Ericson2314:
`obelisk-generated-static` will no longer contain `staticRootPath`. Instead `ob run` will pass that in to the interpreter. That way no builds are polluted.


luigy:
Thanks!

PR #300, opened 2018-11-05
Manage Android signing keys correctly for app release versions

hSloan:
- .jks file is stored in deploy root directory
- json object of keytool credentials are stored within `android_keytool_config.json` of deployment directory


README.md
```diff
@@ -258,7 +258,7 @@ The previous section would have generated a debug version of the app. In order t
 First, if you do not already have a keystore, create it as follows (for more information, see the [Android documentation](https://developer.android.com/studio/publish/app-signing#signing-manually)):
 
 ```
-nix-shell -p androidenv.platformTools --run "keytool -genkey -v -keystore myandroidkey.jks -keyalg RSA -keysize 2048 -validity 10000 -alias myandroidalias"
+nix-shell -E "let pkgs = import <nixpkgs> {}; in pkgs.mkShell { buildInputs = [ pkgs.jdk9 ]; }" --run "keytool -genkey -v -keystore myandroidkey.jks -keyalg RSA -keysize 2048 -validity 10000 -alias myandroidalias"
```
luigy: better to use the project's nixpkgs here to make this doc more reproducible

dep/reflex-platform/github.json
```diff
@@ -1,7 +1,7 @@
 {
   "owner": "reflex-frp",
   "repo": "reflex-platform",
-  "branch": "develop",
-  "rev": "018576456043fe8372a0332c93245a4a620c1fb7",
-  "sha256": "0qhnyyf96iajz97pbqfsw2l1kkbr1yq3bvw3qhjk031vdrkv7b7y"
+  "branch": "android-buildApp-overridable",
```
luigy: This can point back to `develop` now

README.md
ryantrinkle: This is a really complex command to ask the user to do.  Can we give them an `ob` command or make it an automatic part of deployment somehow?

luigy: Actually this whole section is outdated now that `obelisk` generates the keystore/key for the user. The user only gets prompted for a keystore password during deploy command and everything else gets set with default values that get tracked from within config directory. Though it would still be useful to have documented how to get an environment with `keytool` in order to make tweaks from obelisk's default values like switching alias or anything else

ryantrinkle:
Why does androidKeyStore belong in config/backend?  Does the backend actually use it?

luigy:
@ryantrinkle hmm the `backend` does not make use it and it can be outside the `config` folder given that it only gets used when building the apk but not after

README.md
ryantrinkle: @hSloan @luigy Based on this discussion, should we get rid of this stuff?

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -150,13 +159,88 @@ deployPush deployPath getNixBuilders = do
...
+
+
 deployMobile :: MonadObelisk m => String -> [String] -> m ()
 deployMobile platform mobileArgs = withProjectRoot "." $ \root -> do
   let srcDir = root </> "src"
   exists <- liftIO $ doesDirectoryExist srcDir
   unless exists $ failWith "ob test should be run inside of a deploy directory"
-  result <- nixBuildAttrWithCache srcDir $ platform <> ".frontend"
-  callProcessAndLogOutput (Notice, Error) $ proc (result </> "bin" </> "deploy") mobileArgs
+  when (platform == "android") $ do
```
ryantrinkle: Should platform really be a string here?

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -150,13 +159,88 @@ deployPush deployPath getNixBuilders = do
...
+      keyStorePassword <- liftIO $ withEcho False getLine
+      putLog Notice "Re-enter the keystore password: "
+      keyStorePassword' <- liftIO $ withEcho False getLine
+      unless (keyStorePassword' == keyStorePassword) $ failWith "passwords do not match"
+      let keyToolConf = KeytoolConfig
+            { _keytoolConfig_keystore = keystorePath
+            , _keytoolConfig_alias = "obelisk"
+            , _keytoolConfig_storepass = keyStorePassword
+            , _keytoolConfig_keypass = keyStorePassword
+            , _keytoolConfig_dname = "CN=example.com, OU=engineering, O=obelisk, L=LinuxLand, S=NY, C=US" -- TODO Read these from config?
```
ryantrinkle: We need to use real data here.

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -150,13 +159,88 @@ deployPush deployPath getNixBuilders = do
...
+
+createKeystore :: MonadObelisk m => FilePath -> KeytoolConfig -> m ()
+createKeystore root config = do
+  let expr = "with (import " <> toImplDir root <> ").reflex-platform.nixpkgs; pkgs.mkShell { buildInputs = [ pkgs.jdk ]; }"
+  callProcessAndLogOutput (Notice,Notice) $ (proc "nix-shell" ["-E" , expr, "--run" , keytoolCmd]) { cwd = Just root }
+  where
+    keytoolCmd = unwords $ "keytool" :
+      [ "-genkeypair -noprompt"
+      , "-keystore", _keytoolConfig_keystore config
+      , "-keyalg RSA -keysize 2048 -validity 10000"
```
ryantrinkle: Where is this `10000` coming from?

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -150,13 +159,88 @@ deployPush deployPath getNixBuilders = do
...
+
+instance FromJSON KeytoolConfig
+instance ToJSON KeytoolConfig
+
+createKeystore :: MonadObelisk m => FilePath -> KeytoolConfig -> m ()
+createKeystore root config = do
+  let expr = "with (import " <> toImplDir root <> ").reflex-platform.nixpkgs; pkgs.mkShell { buildInputs = [ pkgs.jdk ]; }"
+  callProcessAndLogOutput (Notice,Notice) $ (proc "nix-shell" ["-E" , expr, "--run" , keytoolCmd]) { cwd = Just root }
+  where
+    keytoolCmd = unwords $ "keytool" :
```
ryantrinkle: `unwords` is not an acceptable way of escaping command line arguments.

lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -150,13 +159,88 @@ deployPush deployPath getNixBuilders = do
...
+      [ "-genkeypair -noprompt"
+      , "-keystore", _keytoolConfig_keystore config
+      , "-keyalg RSA -keysize 2048 -validity 10000"
+      , "-storepass", _keytoolConfig_storepass config
+      , "-alias", _keytoolConfig_alias config
+      , "-keypass", _keytoolConfig_keypass config
+      , "-dname ", quote $ _keytoolConfig_dname config
+      ]
+    quote x = "\"" <> x <> "\""
```
ryantrinkle: This is not correct escaping.  Consider what would happen if `x` contains special characters, e.g. `"`.

lib/command/src/Obelisk/Command/Nix.hs
```diff
@@ -180,3 +186,16 @@ nixCmd cmdCfg = withSpinner' ("Running " <> cmd <> desc) (Just $ const $ "Built
...
       , (\a -> " [" <> a <> "]") <$> (commonCfg ^. nixCmdConfig_target . target_attr)
       ]
+
+data NValue
+  = NValue_Text Text
+  | NValue_Path FilePath
+
+renderNValue :: NValue -> Text
+renderNValue = \case
+  NValue_Text x -> "\"" <> x <> "\""
```
ryantrinkle: This is incorrect.

lib/command/src/Obelisk/Command/Nix.hs
```diff
@@ -180,3 +186,16 @@ nixCmd cmdCfg = withSpinner' ("Running " <> cmd <> desc) (Just $ const $ "Built
...
+  NValue_Path p -> T.pack p
+
+renderAttrset :: Map Text NValue -> Text
+renderAttrset m = "{ " <> T.unwords (map (\(k,v) -> k <> " = " <> renderNValue v <> ";") $ M.toList m) <> " }"
+
```
ryantrinkle: If we are pretty-printing nix, please us hnix.

lib/command/src/Obelisk/Command/Deploy.hs
luigy: I believe this value came from the android example https://developer.android.com/studio/publish/app-signing#signing-manually

lib/command/src/Obelisk/Command/Deploy.hs
luigy: going to drop this default `Distinguished Name` configuration and instead let `keytool` prompt for this

lib/command/src/Obelisk/Command/Nix.hs
luigy: :+1:

lib/command/src/Obelisk/Command/Nix.hs
luigy: :+1:

lib/command/src/Obelisk/Command/Deploy.hs
luigy: :+1:

lib/command/src/Obelisk/Command/Deploy.hs
luigy: This now uses a data type

lib/command/src/Obelisk/Command/Deploy.hs
ryantrinkle: OK, yeah, `If you plan to support upgrades for an app, ensure that your app signing key has a validity period that exceeds the expected lifespan of that app. A validity period of 25 years or more is recommended. When your key's validity period expires, users will no longer be able to seamlessly upgrade to new versions of your app. `.  See also https://stackoverflow.com/questions/14272794/how-to-create-an-android-keystore-rsa-key-with-infinite-validity

I think we should go with 1 million days.

lib/command/src/Obelisk/Command/Deploy.hs
luigy: :+1:

Issue #284, opened 2018-10-29
Command line option for changing the deployment branch


mightybyte:
The way I was told to change the deployment branch was to edit `deploy/src/git.json` and change the branch, then run `ob deploy update`.  This is easy enough, but not at all discoverable.  It would be really helpful if there was an `ob deploy` subcommand that did this.

ryantrinkle:
There should also be a corresponding `ob thunk` command to switch branches.

3noch:
Perhaps both `ob deploy update` and `ob thunk update` could take an extra argument like `--ref <branch>`

ali-abrar:
ob thunk update takes `--branch` now.

mightybyte:
Is there some place you could mention this in `ob deploy` documentation?  There's no way I would have thought to look at the docs for `ob thunk` to solve this problem.

3noch:
I personally would vote for getting rid of `ob deploy update` and just making people realize that `src` is a thunk... Then you don't have to learn two APIs for the same thing.

ryantrinkle:
@3noch I'm not so sure about that.  They're doing the same thing under the hood, but I'm not sure they express the same intent.  Is `ob thunk update src` the only thing `ob deploy update` will *ever* do?  Perhaps, but I'm not sure.

3noch:
If the intent was for them to diverge at some point then it makes sense to keep them separate.

PR #308, opened 2018-11-09
ob thunk update [thunkDir] --branch [branch]

hSloan:
Addresses the following issue: https://github.com/obsidiansystems/obelisk/issues/284

### DIFF
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -627,15 +627,20 @@ setThunk :: MonadObelisk m => String -> FilePath -> m ()
 setThunk = setThunk' False
 
 setThunk' :: MonadObelisk m => Bool -> String -> FilePath -> m ()
-setThunk' noTrail branch thunkDir = do
-  _ <- unpackThunk' noTrail thunkDir
-  -- TODO: may have to get current dir before switching
-  initalDir <- liftIO getCurrentDirectory
-  liftIO $ setCurrentDirectory thunkDir
-  callProcessAndLogOutput (Debug, Error) $ proc "git" ["checkout", branch]
-  liftIO $ setCurrentDirectory initalDir
-  _ <- packThunk' noTrail thunkDir
-  updateThunkToLatest thunkDir
+setThunk' noTrail branch thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir >>= \case
+  Left err -> failWith $ T.pack $ "thunk set : " <> show err
+  Right (ThunkData_Packed _) -> do
+    -- TODO: create a ThunkPtr from the branch name
+    _ <- unpackThunk' noTrail thunkDir
+    -- TODO: if git checkout branch errors out, the directory should be restored back to it's previous state
+    callProcessAndLogOutput (Debug, Error) $ gitProc thunkDir ["checkout", branch]
+    _ <- packThunk' noTrail thunkDir
+    updateThunkToLatest thunkDir
+    -- TODO: use overwriteThunk thunkDir (someThunkPtr)
+  Right (ThunkData_Checkout _) -> do
+    callProcessAndLogOutput (Debug, Error) $ gitProc thunkDir ["checkout", branch]
+    _ <- packThunk' noTrail thunkDir
+    updateThunkToLatest thunkDir
 
 getThunkPtr :: MonadObelisk m => FilePath -> m ThunkPtr
 getThunkPtr = getThunkPtr' True
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -633,12 +633,8 @@ setThunk' noTrail branch thunkDir = checkThunkDirectory thunkDir >> readThunk th
   Right (ThunkData_Packed thunkptr) -> do
     let thunkSrc = _thunkPtr_source thunkptr
         repository = case thunkSrc of
-          ThunkSource_Git gitSource -> (T.unpack $ render $ _gitSource_url gitSource)
-          ThunkSource_GitHub gitHubSource -> T.unpack $ "git@github.com:"
-            <> (untagName $ _gitHubSource_owner gitHubSource)
-            <> "/"
-            <>  (untagName $ _gitHubSource_repo gitHubSource)
-            <> ".git"
+          ThunkSource_Git gitSource -> T.unpack $ render $ _gitSource_url gitSource
+          ThunkSource_GitHub gitHubSource -> T.unpack $ render $ _gitSource_url $ forgetGithub True gitHubSource
     -- check to see if branch exists before modifying directory
     (exitCode, _) <- gitLsRemoteExitCode repository branch
     case exitCode of
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -29,6 +29,8 @@ module Obelisk.Command.Thunk
   , getThunkPtr'
   , parseGitUri
   , uriThunkPtr
+  , setThunk
+  , setThunk'
   ) where
 
 import Control.Applicative
@@ -620,6 +622,20 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
       callProcessAndLogOutput (Debug, Error) $ proc "rm" ["-rf", thunkDir]
       liftIO $ createThunk thunkDir thunkPtr
       pure thunkPtr
+
+setThunk :: MonadObelisk m => String -> FilePath -> m ()
+setThunk = setThunk' False
+
+setThunk' :: MonadObelisk m => Bool -> String -> FilePath -> m ()
+setThunk' noTrail branch thunkDir = do
+  _ <- unpackThunk thunkDir
+  -- TODO: may have to get current dir before switching
+  initalDir <- liftIO getCurrentDirectory
+  liftIO $ setCurrentDirectory thunkDir
+  callProcessAndLogOutput (Debug, Error) $ proc "git" ["checkout", branch]
+  liftIO $ setCurrentDirectory initalDir
+  _ <- packThunk thunkDir
+  updateThunkToLatest thunkDir
 
 getThunkPtr :: MonadObelisk m => FilePath -> m ThunkPtr
 getThunkPtr = getThunkPtr' True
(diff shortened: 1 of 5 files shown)


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +623,20 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
       liftIO $ createThunk thunkDir thunkPtr
       pure thunkPtr
 
+setThunk :: MonadObelisk m => String -> FilePath -> m ()
+setThunk = setThunk' False
+
+setThunk' :: MonadObelisk m => Bool -> String -> FilePath -> m ()
+setThunk' noTrail branch thunkDir = do
+  _ <- unpackThunk' noTrail thunkDir
```
3noch: What if it's already unpacked?

3noch: Also if this command fails, I think it should leave the user's thunk in *exactly* the same state it was before.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +623,20 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+  -- TODO: may have to get current dir before switching
+  initalDir <- liftIO getCurrentDirectory
+  liftIO $ setCurrentDirectory thunkDir
+  callProcessAndLogOutput (Debug, Error) $ proc "git" ["checkout", branch]
```
3noch: Use `gitProc` instead as it ensures you run it in the right directory.

3noch:
I wonder if just adding an argument like `--ref` to these commands is the right approach.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +624,37 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+setThunk = setThunk' False
+
+setThunk' :: MonadObelisk m => Bool -> String -> FilePath -> m ()
+setThunk' noTrail branch thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir >>= \case
+  Left err -> failWith $ T.pack $ "thunk set : " <> show err
+  Right (ThunkData_Packed thunkptr) -> do
+    let thunkSrc = _thunkPtr_source thunkptr
+        repository = case thunkSrc of
+          ThunkSource_Git gitSource -> (T.unpack $ render $ _gitSource_url gitSource)
+          ThunkSource_GitHub gitHubSource -> T.unpack $ "git@github.com:"
```
luigy: There is `forgetGithub` which gets you a gitSource

ryantrinkle:
I think a better syntax for this would be `ob thunk update --branch some-branch`.  Switching branches without updating doesn't seem to make much sense, since the commit you're on may not exist in that branch.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +624,33 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+      ExitSuccess -> do
+        _ <- unpackThunk' noTrail thunkDir
+        callProcessAndLogOutput (Debug, Error) $ gitProc thunkDir ["checkout", branch]
+        _ <- packThunk' noTrail thunkDir
+        updateThunkToLatest thunkDir
+      ExitFailure errNum -> failWith $ T.pack $ if errNum == 2
+        then "Error: branch not found"
+        else ("Error Code: " <> show errNum <> "issue checking out the desired branch")
+  Right (ThunkData_Checkout _) -> do
+    callProcessAndLogOutput (Debug, Error) $ gitProc thunkDir ["checkout", branch]
```
luigy: I'm not sure about supporting unpacked thunks for the `thunk set` command. Thoughts @ryantrinkle ?

ryantrinkle: Agreed; I don't see a reason to support it, and it seems dangerous.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +624,33 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+  Right (ThunkData_Packed thunkptr) -> do
+    let thunkSrc = _thunkPtr_source thunkptr
+        repository = case thunkSrc of
+          ThunkSource_Git gitSource -> T.unpack $ render $ _gitSource_url gitSource
+          ThunkSource_GitHub gitHubSource -> T.unpack $ render $ _gitSource_url $ forgetGithub True gitHubSource
+    -- check to see if branch exists before modifying directory
+    (exitCode, _) <- gitLsRemoteExitCode repository branch
+    case exitCode of
+      ExitSuccess -> do
+        _ <- unpackThunk' noTrail thunkDir
```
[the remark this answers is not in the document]
ryantrinkle: Definitely should not unpack.

hSloan:
@ryantrinkle What if I want to set a thunk to a specific hash of a branch (not necessarily the latest)?

ryantrinkle:
@hSloan Do we have that functionality yet?  That seems somewhat separate to me.

hSloan:
@ryantrinkle Ah, ok. I will add that in a separate PR. I brought it up because I lost that functionality during the changes. Just wanted to make sure that wasn't going to be an expected feature. Thanks

ryantrinkle:
It's worth checking. We shouldn't lose functionality.

hSloan:
@luigy ready for another review, thanks!

lib/selftest/src/Obelisk/SelfTest.hs
```diff
@@ -22,7 +22,7 @@ import qualified Network.HTTP.Client as HTTP
 import qualified Network.HTTP.Types as HTTP
 import qualified Network.Socket as Socket
 import Shelly
-import System.Directory (withCurrentDirectory)
+import System.Directory (withCurrentDirectory, getDirectoryContents)
```
luigy: Forgot to add tests?

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +626,47 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+setThunk :: MonadObelisk m => String -> FilePath -> m ()
+setThunk branch thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir >>= \case
+  Left err -> failWith $ T.pack $ "thunk set : " <> show err
+  Right (ThunkData_Packed thunkptr) -> do
+    let thunkSrc = _thunkPtr_source thunkptr
+        repository = case thunkSrc of
+          ThunkSource_Git gitSource -> T.unpack $ render $ _gitSource_url gitSource
+          ThunkSource_GitHub gitHubSource -> T.unpack $ render $ _gitSource_url $ forgetGithub True gitHubSource
+    -- check to see if branch exists before modifying directory
+    (exitCode, _) <- gitLsRemoteExitCode repository branch
```
luigy: The unused value is different than `remoteResults`?

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +626,47 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+    case exitCode of
+      ExitSuccess -> do
+        remoteResults <- readGitProcessNoRepo ["ls-remote", repository, branch]
+        uri <- mkURI $ T.pack repository
+        let rev = head $ T.words remoteResults -- TODO: Don't use head
```
luigy: This should be a proper error

lib/command/src/Obelisk/Command/Utils.hs
```diff
@@ -130,6 +134,21 @@ gitLsRemote repository mRef = do
   putLog Debug $ "git ls-remote maps: " <> T.pack (show maps)
   pure maps
 
+gitLsRemoteExitCode
+  :: MonadObelisk m
+  => String
+  -> String
+  -> m (ExitCode, GitLsRemoteMaps)
+gitLsRemoteExitCode repository branchName = do
```
luigy: This looks pretty similar to `gitLsRemote` maybe we should just combine the two or write one using the other

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +626,47 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+        sha256 <- nixPrefetchGit uri rev False
+        let newThunkPtr = ThunkPtr
+              { _thunkPtr_source = case _thunkPtr_source thunkptr of
+                  ThunkSource_Git gitSource -> ThunkSource_Git GitSource
```
luigy: This could just be setting the branch instead creating a new `GitSource`

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +626,47 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
...
+                    { _gitSource_url = _gitSource_url gitSource
+                    , _gitSource_branch = Just $ N $ T.pack branch
+                    , _gitSource_fetchSubmodules = _gitSource_fetchSubmodules gitSource
+                    }
+                  ThunkSource_GitHub gitHubSource -> ThunkSource_GitHub GitHubSource
```
luigy: Same as above

lib/command/src/Obelisk/Command/Thunk.hs
hSloan: The unused value is the `GitRef` type which seems to give me a ref and not an actual sha1 rev.

lib/selftest/src/Obelisk/SelfTest.hs
hSloan: Yes, It should be there now.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -621,6 +627,42 @@ packThunk' noTrail thunkDir = checkThunkDirectory thunkDir >> readThunk thunkDir
       liftIO $ createThunk thunkDir thunkPtr
       pure thunkPtr
 
+-- set thunk to a desired branch
+setThunk :: MonadObelisk m => String -> FilePath -> m ()
```
luigy: Could this be a thin wrapper around `uriThunkPtr`?

hSloan: Ah! Thanks for pointing `uriThunkPtr` out.

PR #315, opened 2018-11-15
`addOptparseApplicativeCompletionScripts` is upstream

Ericson2314:
I think the nixpkgs bump is new enough.


ryantrinkle:
Nice! Is it any different from ours?

Ericson2314:
I upstreamed it before, and then turns out someone had already written one so they were combined. Pretty sure it's still basically the same.

ryantrinkle:
I'm not sure we should be switching to one that's deprecated.  https://github.com/NixOS/nixpkgs/blob/8d275f6d17c4c1db16c1cb18965308a790da8c63/pkgs/development/haskell-modules/lib.nix#L359

Ericson2314:
I confess I didn't want to switch to a newer nixpkgs less Hydra get stuck with a mass rebuild. It's just a rename so it's semantically fine.

PR #321, opened 2018-11-25
Add a FAQ to document common problems

mightybyte:
I encountered this entitlements error and it took me awhile to find the problem.  I figured this would be a good candidate for starting a FAQ.


luigy:
Thanks!

Issue #320, opened 2018-11-24
Use prerender in the `ob init` example app


mightybyte:
The example app generated by `ob init` doesn't use the `prerender` function.  This makes it not obvious how to go from `DomBuilder` which is needed by `_frontend_body` to `MonadWidget` which you end up wanting for most of your app.  Using something like this would make that more obvious:

```
  , _frontend_body = prerender (text "Loading...") app
```

hSloan:
Going to create a ticket that will make a better error message than the one experienced here.

I'm going to reference this issue https://github.com/obsidiansystems/obelisk/issues/241 because examples and documentation using `prerender` are desperately needed.

mightybyte:
@hSloan I don't think an error message would help much in this case.  The default skeleton app works fine.  It's just that it's not easy to figure out how to change the type from DomBuilder to MonadWidget.  I was thinking something simple like this would demonstrate the concept pretty clearly.

https://github.com/obsidiansystems/obelisk/pull/322

ryantrinkle:
@mightybyte FWIW, I'm hoping to deprecate MonadWidget once we sort out how to make the alternatives more ergonomic.

mightybyte:
@ryantrinkle I tend to think that's likely to be the wrong decision, but I'll reserve judgement until I can see the alternative.  It seems like this is the age-old constraints-vs-liberties tension.  It's really nice to never have to think about what constraint you're using when you're building an app and just use the thing that you know will always give you all the power you might need in the future.  I suspect we'll want this as long as the constraints can't be inferred.  And even if the constraints can be inferred, we're probably still want something like MonadWidget since writing type signatures is the widespread best practice in Haskell.

marcotoniut:
I'm seeing myself deprecating MonadWidget more and more in favor of my own (contained) App/Module/Widget Constraint type synonyms. Aside from making it easier for me to reason about the inner workings of Reflex and Reflex Dom, this practice has also been helpful when abstracting the smaller more reusable reflex components.

ali-abrar:
Perhaps a subroute that uses prerender would be a good compromise. That way there's a documented prerender example, but the default route still takes full advantage of server-side rendering and DOM rehydration.

ryantrinkle:
@ali-abrar Yeah; it could also be just some widget on the page that uses JS; whether or not it's a subroute, it will need some prerender.

cgibbard:
[...]

PR #322, opened 2018-11-27
Use prerender in the skeleton frontend

mightybyte:
A very simple way to address issue #320.


ryantrinkle:
We should definitely have a clear example of how to use `prerender`, but the unfortunate thing about this is that it also eliminates the benefit of server-side rendering.  I also think we should figure out how to make prerender unnecessary in almost all situations.

JBetz:
Closing since it sounds like we want a better way to demonstrate the use of `prerender`, and probably just as importantly, when *not* to use it.

PR #323, opened 2018-11-30
Disallow using '.' in 'ob thunk update'

hSloan:


tomsmalley:
Probably want to check the canonicalized path to catch other variations of `.`, like

hSloan:
> Probably want to check the canonicalized path to catch other variations of `.`, like

Thanks for the heads up. The deed is done.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -380,17 +380,22 @@ createThunkWithLatest target s = do
...
-    Right c -> case c of
-      ThunkData_Packed t -> return (target, t)
-      ThunkData_Checkout _ -> failWith "cannot update an unpacked thunk"
-  let src = _thunkPtr_source ptr
-  rev <- getLatestRev src
-  overwriteThunk overwrite $ ThunkPtr
-    { _thunkPtr_source = src
-    , _thunkPtr_rev = rev
-    }
+  canonicalCurrentPath <- liftIO $ canonicalizePath "."
```
luigy: > Probably want to check the canonicalized path to catch other variations of `.`, like
> [obelisk/lib/command/src/Obelisk/Command/Thunk.hs](https://github.com/obsidiansystems/obelisk/blob/ee5acbb7ad30105e8e98041616d2c995890c87d9/lib/command/src/Obelisk/Command/Thunk.hs#L571)
> 
> Line 571 in [ee5acbb](/obsidiansystems/obelisk/commit/ee5acbb7ad30105e8e98041616d2c995890c87d9)
> 
>  checkThunkDirectory :: MonadObelisk m => FilePath -> m ()

Can this reuse that here + update the error msg to reflect that?

3noch:
Why would we want to disallow this completely? If the thunk is packed then updating on `.` should be harmless. Updating unpacked thunks probably isn't safe in general.

luigy:
This is an improvement over the current broken behavior, but supporting `ob thunk update .` would be nice to have

3noch:
@luigy Agreed. Just seems like if we need to special case this anyway it might be no harder to just support it instead of prohibiting it.

PR #325, opened 2018-12-03
add '-no-user-package-db' flag to ghci and ghcid

hSloan:


ryantrinkle:
Nice! Is there anywhere else we should consider using this?  Maybe this should be a default for our GHC, just like -fexpose-all-unfoldings.

PR #329, opened 2018-12-19
Check if ss command is available on system

dfordivam:


luigy:
I don't think it's available for mac iirc?

vaibhavsagar:
I see, thanks!

PR #337, opened 2019-01-17
Backport master hotfixes to develop

Ericson2314:
I can't imagine why one would need this but not the other? Forgive me if there is a reason.


Ericson2314:
We seem to keep on merging PRs straight to develop :D

PR #338, opened 2019-01-17
Expose Obelisk's overlays 

Ericson2314:
This exposes the underlying overlays making up Obelisk. In general, overlays are the most flexible way to present pkgs in Nix; exposing this is the more equivalent of exposing an `Internals` module in Haskell.

In particular, libraries that build upon Obelisk libraries will need these for their CI, yet since they are not Obelisk projects cannot and should use obelisk's `project` function. This gets them what they need.


haskell-overlays/obelisk.nix
```diff
@@ -0,0 +1,23 @@
...
+  pkgs = self.callPackage ({ pkgs }: pkgs) {};
+  inherit (pkgs) cleanHaskellSource;
+in
+
+{
+  obelisk-executable-config = pkgs.obeliskExecutableConfig.haskellPackage self;
+  obelisk-executable-config-inject = pkgs.obeliskExecutableConfig.platforms.web.inject self;
+
+  obelisk-asset-manifest = self.callCabal2nix "obelisk-asset-manifest" ../lib/asset/manifest {};
+  obelisk-asset-serve-snap = self.callCabal2nix "obelisk-asset-serve-snap" ../lib/asset/serve-snap {};
```
eskimor: Why no `cleanHaskellSource` here?

eskimor:
I would suggest to include a haskell-overlays/default.nix, including all others, which should be advertised as the default file to include. So we can freely change file names/add overlays later on.

haskell-overlays/obelisk.nix
ryantrinkle: Good catch.  @Ericson2314 ?

nixpkgs-overlays/default.nix
```diff
@@ -0,0 +1,15 @@
...
+let
+  inherit (self) lib;
+  inherit (import ../dep/gitignore.nix { inherit lib; }) gitignoreSource;
+in {
+  obeliskCleanSource = src:
+    # WARNING: The order of application here seems to matter a great deal to
+    # how quickly `ghcid` is able to reload changes. As a rule of thumb,
+    # always apply `gitignoreSource` first.
+    # See https://github.com/obsidiansystems/obelisk/pull/666 and related.
+    lib.cleanSource (gitignoreSource src);
```
Ericson2314: I do have doubts about exposing this here, but otherwise the overlays are not self contained. Could call it `_obeliskCleanSource` or something to indicate it is subject to change.

PR #339, opened 2019-01-18
Downgrade spinners

ryantrinkle:
The moon spinner does not render properly in urxvt, and presumably other terminals.

Longer-term, we should really see if we can detect whether unicode is allowed at all, and fallback to something like `/-\|` if not.


hSloan:
Possible Duplicate: Please refer to this PR: https://github.com/obsidiansystems/obelisk/pull/343

It downgrades the spinner and gets rid of the use of other unicode chars that cause obelisk to crash when the bash `LANG` variable is unset.

ali-abrar:
Merging this to fix the issue with certain terminals. That's not to say that @hSloan's PR is invalid, just that it requires more review and this incremental step seems good in the meantime.

PR #341, opened 2019-01-21
Add reader/route instances

alexfmpe:


3noch:
Can we add a default implementation based on MonadTrans to make these less boilerplatey in the long run?

hSloan:
This PR is considered an enhancement, it is meant to reduce existing orphan instances within existing application that use these instances

ali-abrar:
Just waiting on a rebuild due to base branch change before merging

PR #342, opened 2019-01-21
Add renderBackendRoute

ali-abrar:
Depends on #340 

Useful if you want to, e.g., generate a backend route on the frontend. 

This feels a bit dissimilar to the other things in this module - maybe that's a bad sign.

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -833,8 +833,8 @@ makePrisms ''ObeliskRoute
 deriveGEq ''Void1
 deriveGCompare ''Void1
 
--- | Given a backend route and the standard backend route encoder,
--- render the route (path and query string).
+-- | Given a backend route and a checked route encoder, render the route (path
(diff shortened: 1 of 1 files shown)


lib/route/src/Obelisk/Route.hs
```diff
@@ -853,4 +853,33 @@ unsafeRenderBackendRoute backendRouteEncoder =
...
 
+-- | Renders a full backend url, using the provided route prefix (typically the scheme and authority).
+-- This is typically retrived from the "route" configuration in Obelisk applications. The prefix is
+-- assumed not to have a trailing slash.
+renderBackendUrl
+  :: Text -- ^ Base application route
+  -> Encoder Identity Identity (R (Sum br a)) PageName
+  -> R br
+  -> Text
+renderBackendUrl prefix enc = (prefix <>) . renderBackendRoute enc
```
3noch: Because this function just does `<>` for you and introduces an important *assumption* (no trailing slash) I'd vote to just remove it. Just using `<>` is no harder and makes the assumption far more obvious.

PR #344, opened 2019-01-24
Add new loader for GitHub, utilizing fetchTarball

matthewbauer:
This can use Nix 2.0’s builtins.fetchTarball if it is available. If
not, it will fall back on old behavior.

The benefit of builtins.fetchTarball is that we don't need to
import Nixpkgs just for fetchTarball. This means that we
should use less memory for each evaluation.

### DIFF
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -416,8 +416,8 @@ setThunk target gs branch = do
 gitHubStandaloneLoaders :: NonEmpty Text
 gitHubStandaloneLoaders =
   gitHubStandaloneLoaderV4 :|
-  [ gitHubStandaloneLoaderV2
-  , gitHubStandaloneLoaderV3
+  [ gitHubStandaloneLoaderV3
+  , gitHubStandaloneLoaderV2
   , gitHubStandaloneLoaderV1
   ]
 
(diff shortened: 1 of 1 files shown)


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -415,8 +415,9 @@ setThunk target gs branch = do
 -- This tool will only ever produce the newest one when it writes a thunk.
 gitHubStandaloneLoaders :: NonEmpty Text
 gitHubStandaloneLoaders =
-  gitHubStandaloneLoaderV2 :|
-  [ gitHubStandaloneLoaderV3
+  gitHubStandaloneLoaderV4 :|
+  [ gitHubStandaloneLoaderV2
+  , gitHubStandaloneLoaderV3
```
3noch: Might as well order them?

3noch: Oh I see why you did this. Once it's not the default order doesn't matter so might as well order them IMO to prevent anyone thinking there's a reason for this.

3noch:
Might be worth noting that V3 is *still* preferred overall once `bultins.fetchgit` isn't so slow. Has that been fixed?

matthewbauer:
Ok I didn't realize that builtins.fetchGit was being avoided.

ali-abrar:
@matthewbauer @3noch are we waiting on further changes here?

3noch:
@ali-abrar This new loader uses `builtins.fetchGit` which @ryantrinkle wants to avoid until it's fixed in nix.

matthewbauer:
Yeah I can rework it to not use fetchGit

3noch:
Merged into #594

Issue #302, opened 2018-11-05
Add the ability to run app locally with GHCJS-generated code


mightybyte:
There are some circumstances where you need to test the GHCJS-generated code.  The current `ob run` behavior doesn't seem o allow this.

3noch:
This is done with `nix-build -A exe`...albeit not documented?

ElvishJerricco:
@3noch That wouldn't include spinning up the backend and serving assets.

3noch:
It doesn't *include* it but you can do it: `mkdir tmp && ln -s $(nix-build -A exe --no-out-link)/* tmp && cd tmp && ./backend`.

3noch:
Oh but I guess @mightybyte wants a way to do a minimal build of just GHCJS output but still use `ghci` for the backend? That would indeed be nice. `nix-build -A exe` not only builds frontend and backend, but it also runs the GHCJS output through closure-compiler so your build times are pretty long.

mightybyte:
At the moment I don't particularly care how the backend gets run.  I just know that there is a need to run the actual generated JS in some way.  From there I'll leave it to you all to figure out the best scheme, but I would note that users will probably want an easy way to do both forms sooner or later.

ali-abrar:
Documented [here](https://github.com/obsidiansystems/obelisk#locally)

PR #345, opened 2019-01-25
Allow skipping closure-compiler

tomsmalley:


ryantrinkle:
Is this something we want to be configured at the project level?  I can see a use-case there, e.g. as an escape hatch when something isn't compatible with closure compiler, but it seems like we might also want a way to build without closure compiler in a one-off way, e.g. with a command line argument to nix-build or something.  Is that possible with this approach?

tomsmalley:
No, but I can add it to the skeleton's default.nix args I suppose

default.nix
```diff
@@ -164,12 +164,18 @@ in rec {
...
+  } ''
     mkdir $out
     cd $out
     ln -s "${haskellLib.justStaticExecutables frontend}/bin/frontend.jsexe/all.js" all.unminified.js
-    closure-compiler --externs "${reflex-platform.ghcjsExternsJs}" -O ${optimizationLevel} --jscomp_warning=checkVars --create_source_map="all.js.map" --source_map_format=V3 --js_output_file="all.js" all.unminified.js
-    echo "//# sourceMappingURL=all.js.map" >> all.js
+    ${if optimizationLevel == null then ''
+      ln -s all.unminified.js all.js
+    '' else ''
+      closure-compiler --externs "${reflex-platform.ghcjsExternsJs}" -O ${optimizationLevel} --jscomp_warning=checkVars --create_source_map="all.js.map" --source_map_format=V3 --js_output_file="all.js" all.unminified.js
```
3noch: Nit: Using `${pkgs.closurecompiler}/bin/closure-compiler` here would avoid having to fiddle with `buildInputs` above.

3noch:
As far as I can tell, this design makes it difficult to deploy uncompressed JS. In order to get uncompressed you need to `nix-build -arg projectOverrides '{ __closureCompilerOptimizationLevel = null }'`. However `ob deploy` won't pass this argument to `nix-build` and doesn't have a way to let you do so. The only way to get around this would be to change the default in `default.nix` directly which would mean your deployment would match your defaults anyway.

3noch:
It's probably worth noting, however, that "dev" usually means `ob run` or similar which doesn't even use GHCJS, so there's already a pretty big leap from that to production. It would be nice to support a more production-like development setup as #302 requests.

default.nix
tomsmalley: Nice catch :+1:

mightybyte:
I would also like to be able to disable closure compiling on a one-off basis to speed up the situations where I need to test with the generated javascript.

tomsmalley:
@ali-abrar resolved

PR #347, opened 2019-01-29
Check whether staticFilesImpure is a derivation and build it if so

ali-abrar:


ali-abrar:
@Ericson2314

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -54,11 +54,24 @@ run = do
...
+          , "-f"
+          , root
+          , "(let a = import " <> importableRoot <> " {}; in toString (a.reflex.nixpkgs.lib.isDerivation a.passthru.staticFilesImpure))"
+          , "--raw"
+          -- `--raw` is not available with old nix-instantiate. It drops quotation
+          -- marks and trailing newline, so is very convenient for shelling out.
+          ]
+      -- Check whether the impure static files are a derivation (and so must be built)
+      readProcessAndLogStderr Debug $ if isDerivation == "1"
+        then proc "nix-build" ["-E", "(import " <> importableRoot <> "{}).passthru.staticFilesImpure"]
```
Ericson2314: Because no `--raw` for `nix build` or `nix-build`, I think we need to remove trailing newline

ali-abrar:
Trailing newline removed.

PR #353, opened 2019-02-09
Enable WebSockets through nginx configuration

### LINKED ISSUE #352: nginx doesn't proxy WebSocket requests correctly
Attempts to connect to backend web socket endpoints  fail with "426 Upgrade Required", and I think it's because nginx isn't proxying the requests correctly. I tried adding `proxyWebsockets = true;` to the nginx config, but that didn't solve the problem, though it did seem to take care of most of the work.

Maybe the pattern matching on the location needs to be extended to deal with "ws://" requests?

Here's the [relevant snippet](https://github.com/obsidiansystems/obelisk/blob/master/default.nix#L202
) for reference:
```
     services.nginx = {
        enable = true;
        virtualHosts."${routeHost}" = {
          enableACME = enableHttps;
          forceSSL = enableHttps;
          locations.${baseUrl} = {
            proxyPass = "http://localhost:" + toString internalPort;
          }
        };
      };
```

And a relevant article: https://medium.com/@robin.raymond/websocket-proxys-with-nginx-and-the-beauty-of-nixos-e1b95dff1e04

JBetz:
Without this, connection attempts fail with "426 Upgrade Required". Resolves #352.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -206,6 +206,7 @@ in rec {
           forceSSL = enableHttps;
           locations.${baseUrl} = {
             proxyPass = "http://localhost:" + toString internalPort;
+            proxyWebsockets = true;
           };
         };
       };


ryantrinkle:
Awesome; thanks!

3noch:
Now I'm curious why this worked before...

PR #356, opened 2019-02-14
Fix <base> tag rendering.

eskimor:
The generated uris for `static @` should really be absolute and not relative. In a project of mine, we now have some frontend uris which change the path to no longer being "/", without this change static resource loading is broken.

### DIFF
--- a/lib/backend/src/Obelisk/Backend.hs
+++ b/lib/backend/src/Obelisk/Backend.hs
@@ -37,7 +37,6 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Text.Encoding
 import Obelisk.Asset.Serve.Snap (serveAsset)
-import Obelisk.ExecutableConfig.Inject (injectExecutableConfigs)
 import Obelisk.Frontend
 import Obelisk.Route
 import Obelisk.Snap.Extras (doNotCache, serveFileIfExistsAs)
@@ -167,9 +166,8 @@ mkRouteToUrl validFullEncoder =
 
 renderGhcjsFrontend :: MonadIO m => (route -> Text) -> route -> Frontend route -> m ByteString
 renderGhcjsFrontend urlEnc route f = do
-  let baseTag  = elAttr "base" ("href" =: "/") blank --TODO: Figure out the base URL from the routes
-      ghcjsPreload = elAttr "link" ("rel" =: "preload" <> "as" =: "script" <> "href" =: "ghcjs/all.js") blank
+  let ghcjsPreload = elAttr "link" ("rel" =: "preload" <> "as" =: "script" <> "href" =: "ghcjs/all.js") blank
       ghcjsScript = elAttr "script" ("language" =: "javascript" <> "src" =: "ghcjs/all.js" <> "defer" =: "defer") blank
   liftIO $ renderFrontendHtml urlEnc route
-    (baseTag >> _frontend_head f >> injectExecutableConfigs >> ghcjsPreload)
+    (_frontend_head f >> ghcjsPreload)
     (_frontend_body f >> ghcjsScript)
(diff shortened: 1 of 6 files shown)


3noch:
Ideally the `/static` portion would be determined by some "root static path" configuration/route instead of hardcoded.

ryantrinkle:
@eskimor The way we do this right now is to add a `<base>` tag to the head, which should always set the app's base path to the same place.  If this isn't working, we should see if we can fix it some other way.

The issue with making `static` produce an absolute path is that then that path would have to be baked in at compile-time, would have to change between platforms, etc.  Seems complicated, to me.

Can you give some more info on the problems you're seeing?

3noch:
Should obelisk have a way to set up `<base>` for you based on route config?

ryantrinkle:
@3noch It does: see the head of this: view-source:https://reflex-frp.org/docs

ryantrinkle:
Oh, maybe it's not based on the route config; but it definitely sets the base for you.

3noch:
Oh this must be a newer feature. Awesome.

ryantrinkle:
This was added along with routes, because otherwise routes are constantly broken in the exact way @eskimor described.

eskimor:
Thanks @ryantrinkle  - I was not aware of the \<base\> tag. What was causing problems for pact-web was twofold:

1. The base tag does not get added in `renderJsaddleFrontend`.
2. It gets added in `renderGhcjsFrontend` - but only after the head of the frontend gets rendered, which is too late if the head of the frontend already contained some references (which is the case for pact-web).

Both are now fixed by this PR.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -171,5 +171,5 @@ renderGhcjsFrontend urlEnc route f = do
       ghcjsPreload = elAttr "link" ("rel" =: "preload" <> "as" =: "script" <> "href" =: "ghcjs/all.js") blank
       ghcjsScript = elAttr "script" ("language" =: "javascript" <> "src" =: "ghcjs/all.js" <> "defer" =: "defer") blank
   liftIO $ renderFrontendHtml urlEnc route
-    (_frontend_head f >> injectExecutableConfigs >> baseTag >> ghcjsPreload)
+    (baseTag >> _frontend_head f >> injectExecutableConfigs >> ghcjsPreload)
```
ryantrinkle: Huh, I wouldn't have expected the order to matter here; can you elaborate, and maybe add a comment mentioning that it's important?

lib/run/src/Obelisk/Run.hs
```diff
@@ -152,7 +152,8 @@ renderJsaddleFrontend :: (route -> Text) -> route -> Frontend route -> IO ByteSt
 renderJsaddleFrontend urlEnc r f =
   let jsaddleScript = elAttr "script" ("src" =: "/jsaddle/jsaddle.js") blank
       jsaddlePreload = elAttr "link" ("rel" =: "preload" <> "as" =: "script" <> "href" =: "/jsaddle/jsaddle.js") blank
-  in renderFrontendHtml urlEnc r (_frontend_head f >> injectExecutableConfigs >> jsaddlePreload) (_frontend_body f >> jsaddleScript)
+      baseTag  = elAttr "base" ("href" =: "/") blank --TODO: Figure out the base URL from the routes
+  in renderFrontendHtml urlEnc r (baseTag >> _frontend_head f >> injectExecutableConfigs >> jsaddlePreload) (_frontend_body f >> jsaddleScript)
```
ryantrinkle: Is it possible to share the rendering logic with Obelisk.Backend?

eskimor: I'll look into it!

lib/backend/src/Obelisk/Backend.hs
eskimor: Will do and I will test again to make really sure - but I tested twice already, so I am pretty sure already.

eskimor:
@ryantrinkle ok moved the common parts to `renderFrontendHtml` where they should be - at least that what makes sense to me and added comment about the order.

lib/run/src/Obelisk/Run.hs
eskimor: Done

ryantrinkle:
Thanks!

PR #357, opened 2019-02-19
Db more routes combinators

danbornside:


lib/route/src/Obelisk/Route.hs
```diff
@@ -859,4 +864,20 @@ renderObeliskRoute e r =
       enc = (pageNameEncoder . hoistParse (pure . runIdentity) e)
   in (T.pack . uncurry (<>)) $ encode enc r
 
+readShowEncoder :: (MonadError Text parse, Read a, Show a, Applicative check) => Encoder check parse a PageName
+readShowEncoder = singlePathSegmentEncoder . unsafeTshowEncoder
+
+
+integralEncoder :: (MonadError Text parse, Applicative check, Integral a) => Encoder check parse a Integer
+integralEncoder = prismEncoder (Numeric.Lens.integral)
```
3noch: extra parens

benkolera:
Was removing the deriveShowTagIdentity intentional? Because maybeEncoder requires a show instance, this breaks anything with a `(Maybe (R Subroute))`. Reflex-examples fails to compile with:

```
common/src/Common/Route.hs:71:58: error:
    • Could not deduce (Data.Dependent.Sum.ShowTag Example Identity)
        arising from a use of ‘maybeEncoder’
      from the context: b ~ Maybe (R Example)
        bound by a pattern with constructor:
                   FrontendRoute_Examples :: FrontendRoute (Maybe (R Example)),
                 in a case alternative
        at common/src/Common/Route.hs:71:7-28
```

ali-abrar:
@benkolera Thanks for pointing this out. Didn't mean to bundle this with @danbornside's encoders.

ali-abrar:
See https://github.com/obsidiansystems/obelisk/pull/365

PR #359, opened 2019-02-20
cliapp, command: Use lenient utf8 decoding

ali-abrar:


ali-abrar:
I was getting some errors like this at the end of `ob deploy push`:

PR #360, opened 2019-02-22
Deploy config tweaks

ali-abrar:
Opening this on behalf of @luigy

### DIFF
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -143,6 +143,7 @@ deployPush deployPath getNixBuilders = do
         [ "-e"
         , "ssh " <> unwords sshOpts
         , "-qarvz"
+        , "--chown=backend:backend"
         , deployPath </> "config"
         , "root@" <> host <> ":/var/lib/backend"
         ]
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -140,7 +140,8 @@ deployPush deployPath getNixBuilders = do
   withSpinner "Uploading config" $ ifor_ buildOutputByHost $ \host _ -> do
     callProcessAndLogOutput (Notice, Warning) $
       proc "rsync"
-        [ "-e ssh " <> unwords sshOpts
+        [ "-e"
+        , "ssh " <> unwords sshOpts
         , "-qarvz"
         , deployPath </> "config"
         , "root@" <> host <> ":/var/lib/backend"
--- a/default.nix
+++ b/default.nix
@@ -238,7 +238,7 @@ in rec {
           inherit uid;
           group = group;
         };
-        users.extraGroups.${group}.gid = gid;
+        groups.${group}.gid = gid;
       };
     };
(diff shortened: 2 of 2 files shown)


lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -140,8 +140,10 @@ deployPush deployPath getNixBuilders = do
   withSpinner "Uploading config" $ ifor_ buildOutputByHost $ \host _ -> do
     callProcessAndLogOutput (Notice, Warning) $
       proc "rsync"
-        [ "-e ssh " <> unwords sshOpts
+        [ "-e"
+        , "ssh " <> unwords sshOpts
         , "-qarvz"
+        , "--chown=backend:backend"
```
3noch: This assumes the name of the user and group. Could this be exposed from the nix instead?

ryantrinkle: That would be ideal, but I'm not sure whether that's actually exposed anywhere.  However, one thing is that there should probably also be a chmod, too.

ryantrinkle: That would be ideal, but I'm not sure whether that's actually exposed anywhere.  However, one thing is that there should probably also be a `chmod`ded, too.

3noch: Using *only* default ob deploy config means it will always be `backend` but if the users builds their deployment with some custom nix (as we do in at least one place), this assumption may be wrong.

3noch: Then again, if it's custom then syncing configs may be broken in other ways (like if you have multiple backends for some reason...). @luigy what happens if this `rsync` fails, does the whole deploy fail? If there is a way to opt out of this step then that might be sufficient. I doubt we want to support all possible configurations of nix...

ryantrinkle: I think before we worry about supporting a custom nix approach, we should try to understand the motivations behind that custom approach.  We should not try to make this thing configurable enough to support all use cases; instead, we should try to build it out of smaller, more composable pieces.  That'll be a lot of work, but I think it'll be time better spent, in the end.

3noch: To do "custom" you currently have to build your own deployment drv out of the pieces that obelisk provides. So in that sense it's already composable pieces. The trouble is, of course, that `ob deploy` as a Haskell library has a hard time knowing about this if you actually use it.

eskimor: Reasons I would/could go custom with for example Gonimo (not yet obelisk based - working on it):

- Have multiple instances running on one server. (We do this right now, for alpha & beta - so I only need one server) - Which was kind of nice and makes good use of nix capabilities, but I could live without that.
- Have other services running along side of the application, e.g. a database, some website, ...
- Have it integrated in some bigger setup with multiple servers that is managed together - ala nixops.

3noch: @eskimor Good summary. Those are the same reasons I've needed "custom" in the past.

Issue #361, opened 2019-02-25
Question on obelisk routing


benkolera:
I've got a question about the obelisk routing. I need to encode the following routes:

* /
* /article/:documentSlug
* /profile/:username
* /profile/:username/favourites

And while I have no problems making an encoder where there is a dynamic text part at the leaf of the GADT tree (e.g. the article route), I'm a little stuck when the dynamic bit happens in a non leaf (e.g. the profile route). I'm not sure whether it is my encoding of the routes, whether I'm missing an important combinator or Obelisk just doesn't handle this yet.

Here are the data types that I defined to model this.

```haskell
newtype DocumentSlug = Username { unDocumentSlug :: Text } deriving (Eq, Ord, Show)
makeWrapped ''DocumentSlug
newtype Username = Username { unUsername :: Text } deriving (Eq, Ord, Show)
makeWrapped ''Username

data FrontendRoute :: * -> * where
  FrontendRoute_Home :: FrontendRoute ()
  FrontendRoute_Article :: FrontendRoute DocumentSlug
  FrontendRoute_Profile :: FrontendRoute (Username, Maybe (R ProfileRoute))
 
data ProfileRoute :: * -> * where
  ProfileRoute_Favourites :: ProfileRoute ()
```

Which makes sense to me, but I get stuck when I go to try to make the encoder for FrontendRoute_Profile.

```haskell
backendRouteEncoder
  :: Encoder (Either Text) Identity (R (Sum BackendRoute (ObeliskRoute FrontendRoute))) PageName
backendRouteEncoder = handleEncoder (const (InL BackendRoute_Missing :/ ())) $
  pathComponentEncoder $ \case
    InL backendRoute -> case backendRoute of
      BackendRoute_Missing -> PathSegment "missing" $ unitEncoder mempty
    InR obeliskRoute -> obeliskRouteSegment obeliskRoute $ \case
      FrontendRoute_Home -> PathEnd $ unitEncoder mempty
      FrontendRoute_Article -> PathSegment "article" $ singlePathSegmentEncoder . unwrappedEncoder
      FrontendRoute_Profile -> PathSegment "profile" $ unicorn

unicorn :: Encoder check parse (Username, Maybe (R ProfileRoute)) PageName
unicorn = pathDynamicComponent unwrappedEncoder profileRMayEnc
  where
    profileRMayEnc :: (MonadError Text check, check ~ parse) => Encoder check parse (Maybe (R ProfileRoute)) PageName
[...]

PR #362, opened 2019-02-27
Adding a combinator for handling with a non closed (non GADT) segment in a non leaf position of the route

benkolera:
This was to address a limitation that I hit when trying to encode this route structure mentioned in issue #361 . Namely:

/profile/:username
/profile/:username/favourites

This is tested [here](https://github.com/qfpl/reflex-realworld-example/blob/master/common/src/Common/Route.hs#L69) (this project is pointing at obelisk from my fork). If I change the ProfileRoute_Favourites segment to be PathEnd (unitEncoder mempty) it correctly fails the overlapping check because it clashes with the /profile/:username route, which is awesome and feels like it isn't a broken combinator from my naive view! 

I'm not sold on the name of the function. I was hoping that you'd have a good name for it or to just crush this PR with letting me know the intended way of encoding this kind of route! :slightly_smiling_face:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -41,7 +41,7 @@ module Obelisk.Route
   , checkEnum1EncoderFunc
   , unitEncoder
   , pathOnlyEncoder
-  , pathSegmentConsEncoder
+  , pathSegmentTupleEncoder
   , singletonListEncoder
   , unpackTextEncoder
   , prefixTextEncoder
@@ -413,7 +413,7 @@ pathComponentEncoderImpl :: forall check parse p. (Monad check, Monad parse)
 pathComponentEncoderImpl this rest =
   chainEncoder (lensEncoder (\(_, b) a -> (a, b)) Prelude.fst consEncoder) this rest
 
-pathSegmentConsEncoder
+pathSegmentTupleEncoder
   :: forall check parse a b.
      ( Monad check
      , Applicative parse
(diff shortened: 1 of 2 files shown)


benkolera:
When I was implementing the frontend using routes of this structure, it felt like you really wanted a complementary thing on the frontend that can pull the tuple apart and keep chaining RoutedT's below for sub routes (whatever b is in the tuple). 

See https://github.com/qfpl/reflex-realworld-example/commit/13709c5b47da78c47bf7ceb59b7cf7c602dbba2d for the use of the code from the second commit to this PR.

benkolera:
Closing this. There was a pathSegmentEncoder introduced by #357 that does this more generally. 

See https://github.com/qfpl/reflex-realworld-example/blob/1bcf46da3853273134b4fb4192295d34bf84fa95/common/src/Common/Route.hs#L69-L71 for using that new work.

cgibbard:
Is it strictly more general? I actually think that tuple encoder might be
handy anyway...

cgibbard:
Ah, right, Dan's function is perhaps a more general thing despite the less
polymorphic type, just because you can use it toward the end of encoding.

benkolera:
@cgibbard : General probably isn't the right word. It's more that my function mostly just does what Dan's does (it 'unconses' a Text segment off and gets the remaining PageName) and the only thing that my function does that Dan's doesn't actually overlaps with the Bifunctor (,) Encoder. With Dan's function, you can rewrite my function as:

```haskell
pathSegmentEncoder . bimap this rest
```

which makes it a syntactic shortcut at best, and probably a questionable one. :slightly_smiling_face: 

What is definitely handy would be a tutorial or docs that explained how to think of Encoders as categories and get them comfy with the power that the premade instances give them. It took me a little bit to figure out that bimap was what I needed and how to use it. I'll write these docs in my real-world example (or blog posts surrounding them), but I can submit them somewhere else if you think they'll be useful there.

PR #371, opened 2019-03-13
obelisk-route: add SetRoute instance for RequesterT

ali-abrar:


cgibbard:
How did we... not... already need this? :D

PR #373, opened 2019-03-16
 Provide access to cookies pre- and post-hydration

ali-abrar:


ali-abrar:
This makes it possible to "hydrate" widgets that depend on cookies, rather than having to hide those widgets under `prerender` calls.

ali-abrar:
Just needed to ifdef out MonadJSM on ghcjs. Builds fine for me now, but let's see what hydra (eventually) says.

PR #378, opened 2019-03-17
Complete export of 'Routed' and 'RoutedT'

alexfmpe:


ryantrinkle:
Oh, actually, what is the purpose of this?  I think these were not exported in order to preserve invariants.

alexfmpe:
I wanted `RoutedT(..)` for defining some instances - just checked and was able to define the instances with `mapRoutedT` instead.

ryantrinkle:
OK, sounds good.  We can also add `unsafeRoutedT` if necessary.  The key thing here is that `RoutedT` provides a strictness guarantee, which could be violated by incorrect uses of `RoutedT`.

benkolera:
Could we change the type of mapRoutedT from

(m a -> n a) -> RoutedT t r m a -> RoutedT t r n a

To

(m a -> n b) -> RoutedT t r m a -> RoutedT t r n b


While we want to keep the internals of routedT hidden, I found myself
wanting to peel off an eventwriter inside of RoutedT which means the type
of mapRoutedT doesn't quite work because it takes a natural transformation
that has to preserve the return type. Eg I want to be able to write:

mapRoutedT runEventWriterT :: RoutedT t r (EventWriterT t w m) a -> RoutedT
t r m (a,w)

I can pr this if you'd like that as the compromise.

ryantrinkle:
@benkolera Yes, that looks like a good generalization to me.

PR #380, opened 2019-03-18
Generalise mapRoutedT

benkolera:
So that we can write things like:

```haskell
mapRoutedT runEventWriterT :: RoutedT t r (EventWriterT t w m) a -> RoutedT t r m (a, Event t w)
```

Because otherwise we risk needing to expose an `unsafeRoutedT` or something that is potentially invariant breaking. This relaxation of mapRoutedT covers all cases that I have hit. See #378 for the discussion that prompted this.


ryantrinkle:
Thanks!

PR #383, opened 2019-03-19
Obelisk.Command.Deploy: use `--use-substitutes`

vaibhavsagar:
From `man nix-copy-closure`:

--use-substitutes / -s
  Attempt to download missing paths on the target
  machine using Nix’s substitute mechanism. Any paths that cannot be
  substituted on the target are still copied normally from the source.
  This is useful, for instance, if the connection between the source and
  target machine is slow, but the connection between the target machine
  and nixos.org (the default binary cache server) is fast.


vaibhavsagar:
There might be a perfectly good reason we aren't doing this already, but I asked @Ericson2314 and he didn't seem to know.

ryantrinkle:
Sounds great to me!

ryantrinkle:
Thanks!

vaibhavsagar:
My pleasure!

PR #384, opened 2019-03-22
Bump the recommended AMI version up to 18.09. 

benkolera:
This worked for me on a deploy just then (not that exact AMI, because I deployed to ap-southeast-2 :slightly_smiling_face: ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ In this section we will demonstrate how to deploy your Obelisk app to an Amazon 
 
 First create a new EC2 instance:
 
-1. Launch a NixOS 17.09 EC2 instance (we recommend [this AMI](https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#LaunchInstanceWizard:ami=ami-40bee63a))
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
Nix 2.0 had a serious bug where `nix copy` ran out of memory https://github.com/NixOS/nix/issues/1681. I remember that it was mitigated in time for NixOS 18.09 but I'm not sure if it's still a concern. IIRC this is why we continued to recommend 17.09 even after 18.03 came out.

ryantrinkle:
We should test this on an 18.09 nano instance; the larger instances never actually triggered the issue because they had enough RAM.

benkolera:
I'm unsure of whether it's in 18.09. But @ali-abrar thought that it was dated info via IRC today.

If 18.09 is no good, 19.03 is right around the corner!

benkolera:
Ahh yeah I tried it on a micro as per the docs. Can test it on a nano tomorrow if you'd like.

ryantrinkle:
That would be great, thanks!  It was also broken on micro before, so it
sounds like the chance is good :)

benkolera:
It worked on the nano just fine. Looks like the bug is gone! :slightly_smiling_face:

ali-abrar:
Thanks!

ryantrinkle:
@benkolera Fantastic; thanks for checking!

PR #391, opened 2019-03-29
fix hydration of HEAD (following up #356 )

xplat:


ali-abrar:
Closing in favor of #402

PR #393, opened 2019-03-29
asset gathering not strict enough

danbornside:


ali-abrar:
Can you add a comment about why this increased strictness is necessary?

lib/asset/manifest/src/Obelisk/Asset/Gather.hs
```diff
@@ -33,7 +35,7 @@ gatherHashedPaths root = go ""
         let relativePath = subdir </> sub
         isFile <- doesFileExist $ root </> relativePath
         if isFile
-          then do hashedRelativePath <- toHashedPath root relativePath
+          then do !hashedRelativePath <- force <$> toHashedPath root relativePath
```
ryantrinkle: `toHashedPath` itself should be changed, not its usage here.

PR #400, opened 2019-04-02
Bump reflex-platform

kmicklas:
The `hydration` branch has already been merged and removed, and this removes code relating to old GHCs which is already removed upstream.


kmicklas:
Hydra eval: https://hydra.reflex-frp.org/eval/191653

Issue #394, opened 2019-03-30
'ob init' demo project crashes on android


malteneuss:
I'm new to obelisk and i'm trying to get [it](https://github.com/obsidiansystems/obelisk/#android)
to run on my android phone.

`nix-build -A android.frontend -o result-android` successfully builds an apk, but `result-android/bin/deploy` doesn't show the app on my phone.

in android i get the following logs
```
I/ActivityManager( 1507): Start proc 9650:com.example.minimal/u0a169 for activity com.example.minimal/systems.obsidian.HaskellActivity
I/example.minima( 9650): Late-enabling -Xcheck:jni
I/example.minima( 9650): The ClassLoaderContext is a special shared library.
D/HaskellActivity( 9650): Logger running
D/HaskellActivity( 9650): jsaddle: unknown RTS option: -N2
...
D/HaskellActivity( 9650): jsaddle: 
D/HaskellActivity( 9650): Haskell main action exited with status 1
D/OpenGLRenderer( 9650): Skia GL Pipeline
D/SurfaceFlinger(  314): duplicate layer name: changing Surface(name=AppWindowToken{3fe7f91 token=Token{cc315b8 ActivityRecord{c57201b u0 com.example.minimal/systems.obsidian.HaskellActivity t10319}}})/@0xde737e7 - animation-leash to Surface(name=AppWindowToken{3fe7f91 token=Token{cc315b8 ActivityRecord{c57201b u0 com.example.minimal/systems.obsidian.HaskellActivity t10319}}})/@0xde737e7 - animation-leash#1
D/SurfaceFlinger(  314): duplicate layer name: changing Surface(name=AppWindowToken{770949c token=Token{1097c0f ActivityRecord{2ee5d6e u0 com.android.launcher3/.lineage.LineageLauncher t10163}}})/@0x5a1ccc2 - animation-leash to Surface(name=AppWindowToken{770949c token=Token{1097c0f ActivityRecord{2ee5d6e u0 com.android.launcher3/.lineage.LineageLauncher t10163}}})/@0x5a1ccc2 - animation-leash#1
W/ziparchive( 2434): Unable to open '/data/app/com.example.minimal-QV_uY2DZK9RD0YYrAQXvHw==/base.apk': No such file or directory
E/droid.launcher( 2434): Failed to open APK '/data/app/com.example.minimal-QV_uY2DZK9RD0YYrAQXvHw==/base.apk' I/O error
[...]

Issue #369, opened 2019-03-12
Android app crash: Surface : getSlotFromBufferLocked: unknown buffer: 0xb812f398


aveltras:
I'm having trouble trying obelisk on Android right now.
The app produced crashes with the following log output (there might be lines coming from other processes than the app, i don't know how to filter these).
The crash happens instantly when i try to open the app on the phone.

```
03-12 16:20:03.706  1059  1655 I ActivityManager: Start proc 20283:systems.obsidian.obelisk.examples.minimal/u0a270 for activity systems.obsidian.obelisk.examples.minimal/systems.obsidian.HaskellActivity
03-12 16:20:03.707  1895  1895 I ContextualWidget: m_Listeners.size = 0
03-12 16:20:03.708 20283 20283 I art     : Late-enabling -Xcheck:jni
03-12 16:20:03.708 20283 20283 I art     : VMHOOK: rlim_cur : 0 pid:20283
03-12 16:20:03.712  1059  1945 W HtcSystemUPManager: HtcSystemUPHandler The policy is disabled. AppId: activity_history, category: launch
03-12 16:20:03.743  1059  1106 D StatusBarManagerService: setSystemUiVisibility(0x8600)
03-12 16:20:03.743  1059  1106 D StatusBarManagerService: disable:userId=0 what=0x0 which=0x1 pkg=Window{10eb53d u0 Starting systems.obsidian.obelisk.examples.minimal}
03-12 16:20:03.743  1059  1106 D StatusBarManagerService: hiding MENU key
03-12 16:20:03.743  1616  1616 I PhoneStatusBar: setSystemUiVisibility vis=8600 mask=ffffffff oldVal=8700 newVal=8600 diff=100
03-12 16:20:03.743  1616  1616 I PhoneStatusBar: updatePolicy(false,false,false,true)
03-12 16:20:03.743  1616  1616 I PhoneStatusBar: hiding the MENU button
03-12 16:20:03.747 20283 20283 W ActivityThread: Application systems.obsidian.obelisk.examples.minimal can be debugged on port 8100...
03-12 16:20:03.755  1895  2415 E Surface : getSlotFromBufferLocked: unknown buffer: 0xb812f398
03-12 16:20:03.763 20283 20283 W linker  : /data/app/systems.obsidian.obelisk.examples.minimal-1/lib/arm/libHaskellActivity.so: unused DT entry: type 0x1d arg 0xb49
[...]

PR #401, opened 2019-04-03
Bump reflex-platform

kmicklas:
Fixes https://github.com/obsidiansystems/obelisk/issues/394 and https://github.com/obsidiansystems/obelisk/issues/369.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,10 @@
 
 This project's release branch is `master`. This log is written from the perspective of the release branch: when changes hit `master`, they are considered released, and the date should reflect that release.
 
+## Unreleased
+
+* Fixed crashes of Android apps on 32-bit ARM devices.
+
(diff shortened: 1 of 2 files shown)


ali-abrar:
@kmicklas would you mind including some more info in the commit message? We've got a lot of these "bump x" commits and I'd like us to start stating the reason.

ali-abrar:
Something like "Bump nixpkgs to fix android 32 bit builds" perhaps?

kmicklas:
@ali-abrar Done.

ali-abrar:
Thanks. One last thing: I think a changelog entry would be good to have here, since we're basically (re-)adding support for 32-bit android. Might be interesting to know. You can add it under "Unreleased".

rvl:
Excellent, thanks @kmicklas - I will try using the obelisk develop branch soon!

PR #402, opened 2019-04-03
fix hydration of HEAD (following up #356) (replacing #391)

xplat:
Take 2, hopefully this one looks better to people?


lib/executable-config/lookup/obelisk-executable-config.cabal
```diff
@@ -8,6 +8,7 @@ library
   if os(ios)
     hs-source-dirs: src-ios
     build-depends: jsaddle-wkwebview
+    build-depends: directory
```
kmicklas: Why does iOS need it but not Android?

ali-abrar:
@xplat have you tested this on each platform?

lib/frontend/src/Obelisk/Frontend.hs
```diff
@@ -77,6 +89,23 @@ data Frontend route = Frontend
...
 
+baseTag :: forall route js t m. ObeliskWidget js t route m => RoutedT t route m ()
+baseTag = elAttr "base" ("href" =: "/") blank --TODO: Figure out the base URL from the routes
+
+removeHTMLConfigs :: JSM ()
+removeHTMLConfigs = void $ runMaybeT $ do
+  doc <- MaybeT currentDocument
+  hd <- MaybeT $ getHead doc
+  es <- collToList =<< getElementsByTagName hd ("script" :: Text)
+  for_ es $ \e -> do
```
kmicklas: Can they all be beneath some element we can identify by a fixed ID instead of enumerating all `script` tags?

lib/executable-config/lookup/obelisk-executable-config.cabal
xplat: Android uses a special wrapper library to find resource files, but ios just treats them as files.

lib/frontend/src/Obelisk/Frontend.hs
xplat: They're already directly beneath HEAD, and the content model for HEAD doesn't exactly have the flexibility people are used to in the BODY

lib/executable-config/inject/src/Obelisk/ExecutableConfig/Inject.hs
```diff
@@ -22,7 +22,7 @@ inject key = do
 -- whose value is the provided configuration
 injectPure :: DomBuilder t m => Text -> Text -> m ()
 injectPure key value =
-  let attrs = ("type" =: "text/plain" <> "id" =: ("config-" <> key) <> "data-executable-config-inject" =: "1")
+  let attrs = ("type" =: "text/plain" <> "id" =: ("config-" <> key) <> "class" =: "obelisk-executable-config-inject")
```
xplat: I'm not 100% on this change -- on the one hand it does make the code simpler, otoh, shouldn't we leave the class namespace for the real stuff on the page to use?  Or at least we should document that we reserve classes starting with 'obelisk-' ...

lib/executable-config/lookup/obelisk-executable-config.cabal
```diff
@@ -31,3 +31,5 @@ library
     filepath,
     text,
     transformers
+
+  ghc-options: -Wall
```
xplat: If we ever want to put this on Hackage we'll have to remove this again, because Hackage. I am not personally against having it in Obelisk tho ...

kmicklas: It's in all the other Obelisk libraries.

lib/executable-config/inject/src/Obelisk/ExecutableConfig/Inject.hs
kmicklas: I mean, by that same logic we would need to reserve the `config-*` IDs for the client code as well.

3noch: `data-*` attrs were added largely so that you wouldn't feel the need to overload `class` for everything. Why not just use them for their intended purpose?

kmicklas: @3noch Hmm, I wasn't aware of those before. A cursory search brought this up: https://intu.io/blog/dont-use-data-attributes-to-find-html-elements-with-js/

xplat: That author's arguments both for and against using data attributes as locators seem pretty weak.

3noch: I agree that this argument is pretty arbitrary. The point is we can avoid stealing class namespace from designers and make it quite clear to the DOM that we are doing something highly custom with these nodes.

kmicklas: Is there an API for looking up elements by these attributes? I couldn't find any.

3noch: `document.querySelectorAll('[data-foo]');`

3noch: To be clear, I don't care that much...just more FYI type interaction.

kmicklas: Nah that's cool, I'll use that (and for the keys as well).

kmicklas:
This still needs to be tested again in GHCJS and I also noticed that `HasFrontendConfigs` doesn't work inside a `prerender`.

kmicklas:
I've confirmed this works on GHCJS and fixed the issues with `prerender`/`Performable`.

PR #403, opened 2019-04-03
Provide a way to accept the Android SDK licence agreement

kmicklas:


srid:
How exactly does this work? The README is not very clear on this. I have the following:

```nix
with import ./.obelisk/impl {
  inherit system iosSdkVersion;
  config.android_sdk.accept_license = true;
};
```

which results in:

```
error: anonymous function at /nix/store/5pz37k73p3h87vkm96mws95frnkff84n-source/default.nix:1:1 called with unexpected argument 'config', at /home/srid/code/heed/default.nix:4:6
```

ryantrinkle:
You may need a newer version of obelisk; I don't think this is on master
yet.

PR #406, opened 2019-04-10
Added -dedupe option to ghcjs builds by default.

eskimor:
- Smaller JavaScript
- Seemingly even faster JavaScript (not measured, just feels like it)
- Way faster deployment, due to closure compiler having far less work to
do.

Size benefits of skeleton app:

Without -dedupe:

  unminified + uncompressed: 9.1 MB
  unminified + compressed: 867 kB
  minifed + uncompressed: 2.8 MB
  minified + compressed: 520.3 kB

With -dedupe:

  unminified + uncompressed: 5.3 MB
  unminified + compressed: 634.7 kB
  minifed + uncompressed: 1.5 MB
  minified + compressed: 361.6 kB

Those reductions improve loading times for bigger apps quite
dramatically, also the time the closure compiler takes is suddenly
manageable again. (Like one minute, instead of half an hour or more.)


eskimor:
I am guessing there is some good reason we are not doing this, but in case there is not, this would speed up our deployments and our page loads quite a bit.

3noch:
:open_mouth: I thought we were already doing this!

eskimor:
Me too actually, but code size grew quite a bit and I checked whether there is anything I can do about it - and just checked whether adding -dedupe makes a difference  and it does.

dalaing:
This looks pretty similar to #57, although it might have bit rotted by now.
I'm just glad to see it happening :)

The DGHCJS_BROWSER bit might also be useful to add at the same time.

eskimor:
Haha - yeah! Now that you say it, I have seen it before. We should definitely have one of those PRs merged. I am currently checking -DGHCJS_BROWSER again (I used it wrongly and therefore it made no difference) and will add it then too.

Whether or not the user should only provide a library is a separate issue. I see the benefits, but also the drawbacks: E.g. what if -dedupe breaks things for a user? He should be at least be able to disable it for his project. So if we go for library only, we should still provide the means for customization. (Like the disabling/configuring of the closure compiler which we have already.)

I am also wondering whether we should include -O by default. It kind of makes sense and in fact improved the loading time of a project I am working on a bit, but at the same time the build takes way more memory now which is easily too much for not so beefy machines. It is really ridiculous how much memory ghc needs.

dalaing:
I'm not really bothered either way by the separate library thing - I think
that was a side point that came up in the old PR that was somewhat
orthogonal to turning on the relevant ghcjs flags (unless I missed
something big).

ali-abrar:
> Whether or not the user should only provide a library is a separate issue. I see the benefits, but also the drawbacks: E.g. what if -dedupe breaks things for a user? He should be at least be able to disable it for his project. So if we go for library only, we should still provide the means for customization. (Like the disabling/configuring of the closure compiler which we have already.)

Didn't we, in fact, find that dedupe broke string concatenation in a couple of (albeit rare) instances?

3noch:
Seemed like it was up in the air which of dedup or the jsstring "hacks" were at fault.

eskimor:
I can say for sure that removing `-dedupe` fixed the issue. So this should definitely not get merged until that issue gets resolved. Resolving that issue would certainly pay off, as `-dedupe` is really quite a productivity and performance booster.

PR #407, opened 2019-04-12
Doc: add FAQ entry on `ob run`

smondet:


FAQ.md
```diff
@@ -59,3 +59,63 @@ error:
...
+    * obtains a free port number.
+* Then runs `ghcid`
+    * with a `--test` command that reruns `Obelisk.Run.run` at each restart.
+
+It is defined in `./obelisk/lib/run/src/Obelisk/Run.hs`:
+
+* It creates a thread (`forkIO`) which starts the main backend.
+  This runs `runSnapWithCommandLineArgs` and
+    * passes routes to the backend
+      (result of the `_backend_run` field of the `Backend x y` “user” record)
```
kmicklas: Are the type parameters just called `x` and `y` in the definition of the record? At least here they should be renamed to be semantically clear. Also I'm not sure "user" is necessary.

FAQ.md
```diff
@@ -59,3 +59,63 @@ error:
...
+    * or to `serveDefaultObeliskApp` (`./lib/backend/src/Obelisk/Backend.hs`) to
```
kmicklas: I'm not sure what this "or" means. What's the condition?

FAQ.md
```diff
@@ -59,3 +59,63 @@ error:
...
+      serve static assets.
+* Starts `runWidget`
+  which itself runs 
+  [runSettingsSocket](https://hoogle.haskell.org/?hoogle=runSettingsSocket)
+  (the *“TCP listen loop”*):
```
kmicklas: I don't think all the stuff below "TCP listen loop" is necessary. At this point you're better off reading the source. (And then we also don't have to worry about staleness.) Also I think you should remove the parenthesis and use regular ASCII quotes.

FAQ.md
smondet: the previous “passes routes” should be factored out of the bullet points, routes are passed to the 1st or the 2nd.

FAQ.md
smondet: I like what is below because it forces the mentioning of the jsaddle and the pre-rendering stuff (things I would have liked to read about in place like that before diving into the code).

FAQ.md
```diff
@@ -59,3 +59,63 @@ error:
...
+
+#### Short version:
+
+`ob run` starts a [`ndmitchell/ghcid`](https://github.com/ndmitchell/ghcid)
+process, within a carefully crafted `nix-shell`, which tries to build your
+project and,
+
+* either displays compilation errors,
+* or starts the Obelisk server, which serves:
+     * your backend's route handlers,
```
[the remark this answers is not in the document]
smondet: Usually in “Manuals of Style” bullet-point lists should be proper english

https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Lists#List_layout

smondet:
> Also can you add comments to the relevant source files to point out that this need to be updated if the implementation is changed?

@kmicklas  Do we use some form of standard format for those backlinks, or just `-- See also FAQ.md`?

FAQ.md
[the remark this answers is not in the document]
kmicklas: It's quite unusual for anything technical in nature.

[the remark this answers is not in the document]
ali-abrar: I'm with @smondet on this one.

ali-abrar:
> > Also can you add comments to the relevant source files to point out that this need to be updated if the implementation is changed?
> 
> @kmicklas Do we use some form of standard format for those backlinks, or just `-- See also FAQ.md`?

I think we have to make up the format

PR #414, opened 2019-04-21
Generalize backend formulation and expose a snaplet

mightybyte:
Obelisk usually gets in the way of various things I want to do with the backend.  This explores how we might change it to be more general and work with existing snaplet infrastructure more easily.


mankyKitty:
@mightybyte Quick liveness test, are these changes still something you'd like to see in obelisk proper?

PR #417, opened 2019-04-25
Add an "ob shell" command

cgibbard:
I happened to build this off of Dan Bornside's asset gathering strictness change, because that's what we had at hand, hopefully that's fine.


kmicklas:
@cgibbard https://github.com/obsidiansystems/obelisk/pull/393 looks to be already merged so why is that commit still showing up here? Either way can you rebase to not include it?

ryantrinkle:
@kmicklas I don't know why that was included here, either - maybe a github glitch? - but in any case I'm just going go merge, since it's passed CI.

PR #418, opened 2019-04-29
Allow skeleton's obelisk to be overridden and fix release.nix

ali-abrar:


ali-abrar:
Need to update hydra to build with `--arg self-args '{ config.android_sdk.accept_license = true; }'`

ryantrinkle:
We can update hydra by merging something like this to develop: https://github.com/reflex-frp/reflex-platform/blob/a15d3a2411e7ca7d4ee4853b57c72fe83faee272/jobsets.nix#L30-L34

PR #423, opened 2019-05-02
RFC: Add routeLinkAttr

srid:
Motivation: `routeLink` as it is makes it impossible to say add a class attribute (eg: for semantic-ui's menu tab) to the created `a` element . 

The solution this PR adapts is to extend the function to take in a custom `attr` parameter to set whatever element attributes we want; just as `elAttr` exists alongside `el`.


srid:
Open question: what if I want to pass in a dynamic of attributes (a la `elDynAttr`)?

srid:
Closing, because #547 looks better and up to date.

PR #426, opened 2019-05-05
'Iso' based utils

alexfmpe:


lib/route/src/Obelisk/Route.hs
```diff
@@ -312,12 +315,53 @@ isoEncoder f = unsafeMkEncoder $ EncoderImpl
   , _encoderImpl_decode = pure . view (from f)
   }
 
+-- | Lift an 'Iso' over the decoded format to an 'Iso' over encoders
```
ryantrinkle: Can't this be written using existing primitives?  E.g. `isoEncoder` and `(.)`

Even if we want to add this for convenience, it should be implemented that way so that it's obviously correct rather than requiring the reader to understand the details.

lib/route/src/Obelisk/Route.hs
```diff
@@ -312,12 +315,53 @@ isoEncoder f = unsafeMkEncoder $ EncoderImpl
...
 unwrappedEncoder = isoEncoder _Wrapped'
 
+-- | Combine two encoders into a more complex one given isomorphisms for decoded/encoded formats
+pairEncoders
+  :: (Applicative check, Monad parse)
+  => Iso' (d1, d2) d
+  -> Iso' (e1, e2) e
+  -> (Encoder check parse d1 e1, Encoder check parse d2 e2)
+  -> Encoder check parse d e
+pairEncoders isoD isoE = view (isoViaDecoded isoD) . view (isoViaEncoded isoE) . uncurry bimap
```
ryantrinkle: This would make more sense to me as something like
```haskell
pairEncoders isoD isoE (f, g) = isoEncoder isoE . bimap f g . isoEncoder (from isoD)
```
(not sure I got that exactly right)

lib/route/src/Obelisk/Route.hs
alexfmpe: Oh right. For some reason I thought I couldn't leverage `isoEncoder` here.
The one downside is that now we have stricter constraints we don't seem to actually need.

lib/route/src/Obelisk/Route.hs
alexfmpe: Indeed - fixed

alexfmpe: I'm no longer convinced `pairEncoders` is worth having though

3noch:
@alexfmpe Should we delete `pairEncoders` then?

alexfmpe:
I forgot the original use case - I'll need to revisit this and evaluate again

lib/route/src/Obelisk/Route.hs
cgibbard: The type of this thing is actually bizarre. It's like it comes from an alternate universe where we somehow use lenses or other optics to manipulate the internals of Encoders directly. You can apply isoEncoder to it and get an Encoder which transforms Encoders to other Encoders. Too higher order! :D

3noch:
> The type of this thing is actually bizarre. It's like it comes from an alternate universe where we somehow use lenses or other optics to manipulate the internals of Encoders directly. You can apply isoEncoder to it and get an Encoder which transforms Encoders to other Encoders. Too higher order! :D

Based on @cgibbard's comment I'm going to close this. We can achieve the same thing without being higher order.

PR #443, opened 2019-05-28
Add some new encoders to aid in correctly writing encoders that encode query parameters from record types.

cgibbard:


lib/route/src/Obelisk/Route.hs
```diff
@@ -547,13 +553,10 @@ pathOnlyEncoderIgnoringQuery = unsafeMkEncoder $ EncoderImpl
...
 
 pathOnlyEncoder :: (Applicative check, MonadError Text parse) => Encoder check parse [Text] PageName
-pathOnlyEncoder = unsafeMkEncoder $ EncoderImpl
-  { _encoderImpl_decode = \(path, query) ->
-      if query == mempty
-      then pure path
-      else throwError "pathOnlyEncoderImpl: query was provided"
-  , _encoderImpl_encode = \path -> (path, mempty)
-  }
+pathOnlyEncoder = second (unitEncoder mempty) . coidr
```
benkolera: Doesn't this break the old behaviour of throwing an error if there was a query? Is this intentional?

cgibbard: It just throws a slightly different error, since the unitEncoder will fail.

PR #451, opened 2019-06-11
Add an `_R` prism which is useful with `modifyRoute`

Ericson2314:
Write it and rewrite the existing `rPrism` in terms of some more general optics.
This will hopefully aid future attempts to put the most general mechanisms in
`dependent-sum` itself.


ali-abrar:
@Ericson2314 Did you mean to leave all that commented code in there?

Ericson2314:
Yes the code works and I mean to leave it to help with the follow-up work / explain the comments.

PR #453, opened 2019-06-16
Don't redirect the output of backend to files, let it go to syslog.

eskimor:
This way you can check the logs via journalctl -e -u backend and the disk should not get filled constantly.


3noch:
I prefer this as well

3noch:
Getting user reports due to this. Can we get this merged? `${backendArgs}` allows you to add it back in your use case if you need it.

mightybyte:
I would also like this.  I opened this issue on the topic.  https://github.com/obsidiansystems/obelisk/issues/484

PR #454, opened 2019-06-16
Add --scheme option to ob thunk pack/unpack commands

eskimor:
This way one can switch urls when unpacking/packing. E.g. use `ob thunk unpack --scheme ssh dep`, so you can conveniently push from the unpacked repo, then use `ob thunk pack --scheme https dep` to get back an `https` url for the packed thunk.


3noch:
I wonder if using `git ls-romet --get-url` would solve this. Need to think about that.

3noch:
The idea being: If you have an `insteadOf` config somewhere then `git ls-remote --get-url` will do the substitution whereas `git remote get-url` does not. Which URL should we be putting in the thunk? We shouldn't assume that everyone has the same `insteadOf` configuration.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -693,12 +727,15 @@ packThunk :: MonadObelisk m => ThunkPackConfig -> FilePath -> m ThunkPtr
...
-packThunk' noTrail (ThunkPackConfig force thunkConfig) thunkDir = checkThunkDirectory "Can't pack/unpack from within the thunk directory" thunkDir >> readThunk thunkDir >>= \case
+packThunk' noTrail (ThunkPackConfig force thunkConfig tScheme) thunkDir = checkThunkDirectory "Can't pack/unpack from within the thunk directory" thunkDir >> readThunk thunkDir >>= \case
   Left err -> failWith $ T.pack $ "thunk pack: " <> show err
   Right (ThunkData_Packed _) -> failWith "pack: thunk is already packed"
   Right (ThunkData_Checkout _) -> do
     withSpinner' ("Packing thunk " <> T.pack thunkDir)
                  (finalMsg noTrail $ const $ "Packed thunk " <> T.pack thunkDir) $ do
+      let overrideScheme = applyThunkScheme tScheme
+      let overrideSchemeThunkPtr = thunkPtr_source . _ThunkSource_Git %~ overrideScheme
+      thunkPtr <- overrideSchemeThunkPtr <$> getThunkPtr (not force) thunkDir
```
3noch: This "feigned" mutation via shadowing isn't common in Haskell (it is in OCaml!) so let's just try to put this all in the same line.

3noch: Would it make sense to move this logic into `modifyThunkPtrByConfig`?

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -136,6 +129,45 @@ data GitSource = GitSource
...
+   = ThunkData_Packed ThunkPtr
+   -- ^ Packed thunk
+   | ThunkData_Checkout (Maybe ThunkPtr)
+   -- ^ Checked out thunk that was unpacked from this pointer
+  deriving (Show, Eq, Ord)
+
+-- | Whether or not to override the scheme when packing/unpacking a thunk.
+data ThunkScheme
+  = ThunkScheme_Keep
+  -- ^ Keep the protocol https/ssh as it was.
```
3noch: It's also possible for it to be `file://` (and `git://` which we don't support yet) but it doesn't make sense for you to switch to or from `file://` because that doesn't have the same URI format.

lib/command/src/Obelisk/Command/Thunk.hs
414owen: Looking back on this. I think the scheme-overridden ptr wasn't used :open_mouth:

eskimor: I am confused: What shadowing?

ali-abrar:
Closing this. If it's still desired, please open an issue/pr on nix-thunk.

PR #461, opened 2019-06-24
WIP: Add a projectOverrides argument to the skeleton's default.nix(?)

cgibbard:
(also getTextConfig, and this PR includes the ob doc command one)

### DIFF
--- a/skeleton/default.nix
+++ b/skeleton/default.nix
@@ -1,11 +1,12 @@
 { obelisk ? import ./.obelisk/impl {
-  system = builtins.currentSystem;
-  iosSdkVersion = "10.2";
-  # You must accept the Android Software Development Kit License Agreement at
-  # https://developer.android.com/studio/terms in order to build Android apps.
-  # Uncomment and set this to `true` to indicate your acceptance:
-  # config.android_sdk.accept_license = false;
+    system = builtins.currentSystem;
+    iosSdkVersion = "10.2";
+    # You must accept the Android Software Development Kit License Agreement at
+    # https://developer.android.com/studio/terms in order to build Android apps.
+    # Uncomment and set this to `true` to indicate your acceptance:
+    # config.android_sdk.accept_license = false;
   }
+, projectOverrides ? {}
 }:
 with obelisk;
 project ./. ({ ... }: {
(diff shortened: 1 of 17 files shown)


lib/executable-config/lookup/src/Obelisk/Configs.hs
```diff
@@ -50,6 +51,9 @@ class Monad m => HasConfigs m where
 instance Monad m => HasConfigs (ConfigsT m) where
   getConfigs = ConfigsT ask
 
+getTextConfig :: HasConfigs m => Text -> m (Maybe Text)
+getTextConfig k = fmap T.decodeUtf8 <$> getConfig k
```
tomsmalley: Should this be `decodeUtf8With lenientDecode`?

cgibbard: You probably want the backend to fail in the case that its config can't be read, though perhaps we should reflect the encoding failure somehow, I don't know.

3noch: In that case we should probably use `fromMaybe (error $ "Failed to decode " <> T.unpack k) . decodeUtf8'`

Ericson2314: @3noch agreed, except I think it better to fail in the monad.

skeleton/default.nix
```diff
@@ -1,16 +1,17 @@
...
-  # Uncomment and set this to `true` to indicate your acceptance:
-  # config.android_sdk.accept_license = false;
+    system = builtins.currentSystem;
+    iosSdkVersion = "10.2";
+    # You must accept the Android Software Development Kit License Agreement at
+    # https://developer.android.com/studio/terms in order to build Android apps.
+    # Uncomment and set this to `true` to indicate your acceptance:
+    # config.android_sdk.accept_license = false;
   }
+, projectOverrides ? {}
```
3noch: Why encode this as part of the skeleton? Do you see this being used by the average obelisk user? It's easy enough to do yourself you really want it...

3noch:
We'll revisit this if it becomes a problem...

PR #463, opened 2019-06-24
Introduce CSS file to skeleton project

rashadg1030:


skeleton/frontend/src/Frontend.hs
```diff
@@ -1,24 +1,30 @@
...
 import Common.Route
-import Obelisk.Generated.Static
+
 
 frontend :: Frontend (R FrontendRoute)
 frontend = Frontend
-  { _frontend_head = el "title" $ text "Obelisk Minimal Example"
+  { _frontend_head = do
+      el "title" $ text "Obelisk Minimal Example"
+      elAttr "link" ("href" =: (static @ "main.css") <> "type" =: "text/css" <> "rel" =: "stylesheet") blank
```
3noch: `"href" =: (static @ "main.css")` can be `"href" =: static @"main.css"`

skeleton/frontend/src/Frontend.hs
```diff
@@ -1,24 +1,30 @@
...
 import Common.Route
-import Obelisk.Generated.Static
+
 
 frontend :: Frontend (R FrontendRoute)
 frontend = Frontend
-  { _frontend_head = el "title" $ text "Obelisk Minimal Example"
+  { _frontend_head = do
+      el "title" $ text "Obelisk Minimal Example"
+      elAttr "link" ("href" =: static @ "main.css" <> "type" =: "text/css" <> "rel" =: "stylesheet") blank
```
3noch: The space between `@` and `"main.css"` is inconsistent with pattern below. I think we prefer no space so it's a bit more obvious that `@` is not an operator.

rashadg1030: Right, my bad!

3noch:
Thanks! Sorry it took so long to get this in!

PR #464, opened 2019-06-24
add withHoogle to obelisk skeleton

hSloan:


skeleton/default.nix
```diff
@@ -6,9 +6,11 @@
     # Uncomment and set this to `true` to indicate your acceptance:
     # config.android_sdk.accept_license = false;
   }
+  , withHoogle ? false
```
alexfmpe:
```suggestion
, withHoogle ? false
```

3noch:
We're going to do this a different way.

PR #466, opened 2019-06-25
exe-config: Fix underflow in ghcjs config lookup (Fixes #465)

### LINKED ISSUE #465: Exe app/ghcjs causes mem leak/crash with prerender
This works fine in `ob run`, but with `nix-build -A exe` app the browser only shows "test" and then it crashes due to memory leak. Could be some issue with hydration.
```
  , _frontend_body = do
      text "test"
      void $ prerender (pure ()) $ do
        text "Welcome to Obelisk!"
        el "p" $ text $ T.pack commonStuff
        elAttr "img" ("src" =: static @"obelisk.jpg") blank
```

tomsmalley:
Fixes #465

### DIFF
--- a/lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
+++ b/lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
@@ -15,15 +15,15 @@ import GHCJS.DOM.Document (getHead)
 import GHCJS.DOM.DOMStringMap (get)
 import GHCJS.DOM.Element (getInnerHTML)
 import GHCJS.DOM.HTMLElement (HTMLElement(HTMLElement), getDataset)
-import GHCJS.DOM.NodeList (item, getLength)
+import GHCJS.DOM.NodeList (IsNodeList, item, getLength)
 import GHCJS.DOM.ParentNode (querySelectorAll)
-import GHCJS.DOM.Types (Node(Node), castTo)
+import GHCJS.DOM.Types (MonadJSM, Node(Node), castTo)
 
 getConfigs :: IO (Map Text ByteString)
 getConfigs = do
   Just doc <- currentDocument
   Just hd <- getHead doc
-  nodes <- collToList =<< querySelectorAll hd ("[data-obelisk-executable-config-inject-key]" :: Text)
+  nodes <- nodeListNodes =<< querySelectorAll hd ("[data-obelisk-executable-config-inject-key]" :: Text)
   fmap Map.fromList $ for nodes $ \node -> do
     e <- castTo HTMLElement node >>= \case
       Nothing -> error "Found node with data attribute obelisk-executable-config-inject-key that is not an HTMLElement."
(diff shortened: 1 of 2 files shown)


lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
```diff
@@ -35,7 +35,7 @@ getConfigs = do
   where
     collToList es = do
       len <- getLength es
-      list <- traverse (item es) [0..len-1]
+      list <- traverse (item es) $ if len == 0 then [] else [0..len-1]
```
3noch: And this, folks, is why subtraction is not a total function.

3noch: Super nitpick idea: `[0..max 0 (len-1)]`

tomsmalley: That actually behaves a little differently when `len = 0`

3noch: Aha right I forgot it was inclusive.

3noch: Haha.... wow. It's hard to deal with under/over flow! If subtraction were `Word -> Word -> Maybe Word` then `max 0 (len-1)` wouldn't type check either. Super tricky.

3noch: I usually just avoid unsigned types like the plague for this reason. I'm just curious why we aren't.

alexfmpe:
Yikes. Is there no way to traverse the collection directly without messing with indexes?
Failing that, maybe we should extract/upstream this fix into an util?

3noch:
Should we make this impl match what seems to be the same bug fixed elsewhere?

https://github.com/obsidiansystems/obelisk/blob/86a9584c6d7648bd5324ef57d62421fed1bf1978/lib/frontend/src/Obelisk/Frontend.hs#L113

3noch:
Your fix avoids `fromIntegral` so maybe you should actually port your fix over to that line as well. Also a comment would be great.

3noch:
Is there a version of `fromIntegral` that throws/`Maybe`s result so that we could "safely" cast to convert to signed type? I would much prefer simply getting rid of the unsignedness altogether.

tomsmalley:
Interesting, didn't even notice that one. Looks like `collToList` itself should be factored out

3noch:
@tomsmalley This bug is bad enough that finding the perfect solution to abstracting out this function is definitely not worth it if it keeps us from getting rid of it. Let's just make the two implementations the same for now with a big TODO on both of them. I'll approve this and hopefully we can get it merged. I'll try to make the change if you don't beat me to it.

Issue #372, opened 2019-03-15
Backend does not reload on deployment if configuration changes


eskimor:
When only changing a config in common (I tested with config/common/route) a deploy won't reload the backend and it continues to use the old config. To fix this a login to the server and doing a manual `systemctl restart backend` is needed.

lgastako:
We just encountered this when doing a deploy to update our email configuration.  The workaround is easy enough for now but it would be great if all deploys that changed anything on the server would restart (or if this is tricky, just all deploys would cause a restart).

JBetz:
To summarize discussions between myself, @ryantrinkle, @3noch, and @eskimor (see #469), the fundamental issue seems to be how to keep the NixOS system and config directory in synch.

Ideally, the backend would itself be able to monitor the config directory to notify the application of updates, without a need for a restart whatsover. This would be precisely correct and as a bonus minimize downtime for config changes, so is definitely something we'd like to build support for in the future.

As for our present situation, the nix store is unpermissioned, so we can't just store configs there, as it would it'd be insecure in any shared cache. Encrypting the configs and then putting them in the nix store *could* work, but then there'd be no way for a user to see the current configuration on their server, and we don't want that.

Other possibilities include passing some representation the config directory to the backend derivation. This would ensure synchronization at deploy time, and having hash-based symlinks to config directories as a recreation of the nix store just for the backend would be even more robust, but neither would be able to handle subsequent mutations. We also considered manually trigger a server restart based on whether rsync detects changes, but that breaks if the deployment fails between those steps, so only pushes the problem further. 

All this to say, our current solutions all came up short in some respect, but we're definitely open to other ideas. And for now, it seems better to re-evaluate this when the landscape changes than try to implement a partial solution.

3noch:
IMO using a hash-based file structure to mimic nix is the least-bad solution. You could give it very restrictive permissions and drop the write bit to make it clear that writing should not be done. `ob deploy` would *always* rewrite the configs so that after any given `ob deploy` the server's state is exactly what you'd expect. If someone mutates config on the server, it's probably because they're trying to debug anyway (since that config should immediately be copied back to the deployment setup). And mutating the config within a given hashed directory would actually work fine until it gets overwritten by `ob deploy`.

ryantrinkle:
Maybe the config dir should be a git repo?

3noch:
[...]

PR #469, opened 2019-07-03
WIP: Automatic reloading of backend on config changes.

eskimor:
Not yet tested at all. But as this is already lying around for a while and I almost lost it, thought I'll put it up already.

Problem this PR is trying to solve:

If you do `ob deploy push` with only config file changes (no new version of the app gets deployed), the backend will not pick up the new configs. (You would need to do a manual `systemctl restart backend`.) With this PR, the backend should be restarted automatically on config file changes.


ryantrinkle:
I like the comprehensiveness here, but can't this result in the backend being restarted twice during a single deploy?

default.nix
```diff
@@ -191,46 +191,64 @@ in rec {
...
-          description = "${user} service";
-          home = "/var/lib/${user}";
-          createHome = true;
-          isSystemUser = true;
-          group = group;
+        # Restart backend on configuration changes:
+        systemd.services.restart-${name} = {
+          serviceConfig = {
+            type = "oneshot";
+            ExecStart="${nixpkgs.systemd.out}/bin/systemctl restart ${name}.service";
```
3noch: Should probably be `${config.systemd.package}/bin/systemctl`

3noch:
Have you explored a more "stateless" approach where you just piggyback off of `restartIfChanged` using some sort of hash of the config contents?

eskimor:
> I like the comprehensiveness of this approach, but can't this result in the backend being restarted twice during a single deploy?

I think, you are right. I actually tried to avoid exactly that, but somehow I did not think this through apparently. Not only does it restart the backend twice, it would do so in quite a bad way:

- Once when the configs are updated
- Right afterwards at profile switch

Problem: At the first reload, the old backend would be restarted, but with already new config files, which might not be parse-able by the old backend.

So auto-reloading is probably a bad idea in general. It would be better to just reload on deployment in case there was no profile switch.

3noch:
@roberth Couldn't `ob deploy` calculate a hash of the config and pass it as an argument to the build? Then you could do something as silly as adding `# config hash = ${configHash}` to the systemd service and it will suddenly get a different path and restart.

eskimor:
@3noch That's a really smart idea, solving both problems at once. I like that! :-)

3noch:
c.f. #372

ryantrinkle:
@3noch I think this is approximately a good idea, but what happens if the hash gets out of sync with the config?  We can't just put the config into the nix store, but we also should design this so that they can't get out of sync.

3noch:
By "get out of sync" do you mean "someone ssh'ed into the server and tweaked configs manually"?

ryantrinkle:
@3noch No, although that is a use case we need to support.

Here's what we should be shooting for:

- if `ob deploy push` succeeds, at the end of that process, the backend is running with the correct config
- if `ob deploy push` fails, it doesn't leave things in a half-deployed state

3noch:
In that case I'm not sure what else would cause config to get out of sync? If, barring manual mutation on the server, the only way for configs to be changed is with `ob deploy` then the hash will be calculated every time and the configs copied every time.

3noch:
Maybe I should clarify that slapping a comment in the derivation wasn't actually the *end* of the idea. What would need to be done, in reality, is have separate directories for each hash-fingerprinted version of the config. Then the backend would symlink to the appropriate one. That is just replicating nix behavior but not in /nix/store

ryantrinkle:
After some more thought, I think if an `ob deploy push` fails partway through, we're OK anyway: the way to recover will be to do another `ob deploy push`, at which point the config will be fixed.

madeline-os:
Closing in favor of #940 (which was merged. yay)

PR #475, opened 2019-07-17
Allow users to configure extra ghci arguments via default.nix

### LINKED ISSUE #474: Support ghci options like -fobject-code and -fno-code
`-fobject-code` is *necessary* if you want to use unboxed sums in your code.

3noch:
Fixes #474

### DIFF
--- a/lib/cliapp/src/Obelisk/CliApp/Process.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Process.hs
@@ -20,6 +20,8 @@ module Obelisk.CliApp.Process
   , callProcess
   , callCommand
   , reconstructCommand
+  , shellQuoteAndEscapeDouble
+  , shellQuoteAndEscapeSingle
   ) where
 
 import Control.Monad ((<=<), join, void)
@@ -182,5 +184,10 @@ reconstructCommand :: Process.CmdSpec -> Text
 reconstructCommand (Process.ShellCommand str) = T.pack str
 reconstructCommand (Process.RawCommand c as) = processToShellString c as
   where
-    processToShellString cmd args = T.unwords $ map quoteAndEscape (cmd : args)
-    quoteAndEscape x = "'" <> T.replace "'" "'\''" (T.pack x) <> "'"
+    processToShellString cmd args = T.unwords $ map (shellQuoteAndEscapeSingle . T.pack) (cmd : args)
+
+shellQuoteAndEscapeSingle :: Text -> Text
+shellQuoteAndEscapeSingle x = "'" <> T.replace "'" "'\''" x <> "'"
+
+shellQuoteAndEscapeDouble :: Text -> Text
+shellQuoteAndEscapeDouble x = "\"" <> T.replace "\"" "\\\"" x <> "\""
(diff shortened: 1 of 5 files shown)


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -202,21 +203,24 @@ warnDifferentLanguages _ = return ()
...
   => FilePath -- ^ Path to .ghci
+  -> [String] -- ^ Extra ghci arguments
   -> Maybe String -- ^ Optional command to run at every reload
   -> m ()
-runGhcid dotGhci mcmd = callCommand $ unwords $ "ghcid" : opts
+runGhcid dotGhci ghciArgs mcmd = callCommand $ unwords $ "ghcid" : opts
   where
     opts =
       [ "-W"
       --TODO: The decision of whether to use -fwarn-redundant-constraints should probably be made by the user
```
3noch: Note that we could get rid of this `TODO` now if we moved `-fwarn-redundant-constraints` into the default list of ghci arguments and removed it from the built-in list.

3noch: We also have the option of detecting "unspecified" and reverting to this as a default even for users who upgrade obelisk without modifying their `default.nix`.

3noch:
An alternative design would be to inject some configuration in the `.ghci` file. This would be strictly more powerful, but also with risk of really screwing with things.

lib/cliapp/src/Obelisk/CliApp/Process.hs
```diff
@@ -181,5 +183,10 @@ reconstructCommand :: Process.CmdSpec -> Text
 reconstructCommand (Process.ShellCommand str) = T.pack str
 reconstructCommand (Process.RawCommand c as) = processToShellString c as
   where
-    processToShellString cmd args = T.unwords $ map quoteAndEscape (cmd : args)
-    quoteAndEscape x = "'" <> T.replace "'" "'\''" (T.pack x) <> "'"
+    processToShellString cmd args = T.unwords $ map (shellQuoteAndEscapeSingle . T.pack) (cmd : args)
+
+shellQuoteAndEscapeSingle :: Text -> Text
```
vaibhavsagar: Is there a more robust shell-escaping strategy we could use instead? I found http://hackage.haskell.org/package/shell-escape and http://hackage.haskell.org/package/posix-escape, but neither of those packages seem to have been updated recently.

3noch: Great question. Mind if I open a ticket for this and bump it to a different PR?

3noch: However, if we went with the "add things to .ghci" design then it would make *this particular feature* skirt the problem.

vaibhavsagar: Sure, I'm not expecting you to rework existing code.

3noch: #476

3noch:
Merge conflicts resolved. We need this to be merged for at least of our projects. Can we get this reviewed + merged?

3noch:
Obviated by #529

PR #481, opened 2019-07-23
Add encodeToText function

meditans:
Similar to 'encode', this is a convenience function to get directly
the textual encoding of a route (which can be used, for example, in a href),
given that one has an 'Encoder' to a 'PageName'.


lib/route/src/Obelisk/Route.hs
```diff
@@ -239,6 +240,17 @@ tryDecode (Encoder (Identity impl)) x = _encoderImpl_decode impl x
...
 encode (Encoder (Identity impl)) x = _encoderImpl_encode impl x
 
+-- | Similar to 'encode' above, this is a convenience function to get directly
+-- the textual encoding of a route (which can be used, for example, in a href),
+-- given that one has an 'Encoder' to a 'PageName'.
+encodeToText :: forall a. Encoder Identity Identity a PageName -> a -> Text
+encodeToText e a =
+  let
+    e' :: Encoder Identity (Either Text) a PathQuery
+    e' = pageNameEncoder <<< hoistParse (pure . runIdentity) e
```
cgibbard: Is the hoistParse actually necessary here? I'd expect that you can just use encode on pageNameEncoder <<< e without doing that bit.

meditans: `e` has a type signature: ```Encoder Identity Identity a PageName```, while 
```
pageNameEncoder :: (Applicative check, MonadError Text parse) => Encoder check parse PageName PathQuery
```
So that `hoistParse (pure . runIdentity)` is required to change the `parse` monad to `Identity` to something that has `MonadError Text`. There are of course alternative approaches, which one would you prefer?

cgibbard: ahhh, okay, nevermind then, looks good :)

ali-abrar:
@meditans this looks like it's mixed in with a bunch of other stuff, including some changes to how we compile to js. Could you explain the other changes in this PR or, preferably, separate them out into other PRs?

Ericson2314:
Separated now.

ali-abrar:
How does this relate to `renderBackendRoute`, `renderFrontendRoute`, and `renderObeliskRoute`?

ali-abrar:
@meditans  @cgibbard ^ ?

meditans:
Yeah, this feature is probably subsumed by `renderBackendRoute`, `renderFrontendRoute`, and `renderObeliskRoute` from `Obelisk.Route`. I wasn't aware of those at the time.

@cgibbard for confirmation

PR #482, opened 2019-07-26
exe-config: fix ghcjs config gathering

tomsmalley:
Previously the keys in the map would evaluate to `<undefined>`, leading to really weird behaviour: `getConfig` returning the same config no matter the path, and, it seems, hangs under some circumstances

### DIFF
--- a/lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
+++ b/lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
@@ -30,7 +30,8 @@ getConfigs = do
       Just htmlE -> return htmlE
     dataset <- getDataset e
     (,)
-      <$> get dataset ("obelisk-executable-config-inject-key" :: Text)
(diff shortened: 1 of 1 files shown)


3noch:
If I could approve this twice I would.

xplat:
if i could approve it at all i would ...

3noch:
If I could approve you to give approval I would...

3noch:
Turns out I can. :joy:

lib/executable-config/lookup/src-ghcjs/Obelisk/ExecutableConfig/Lookup.hs
```diff
@@ -30,7 +30,8 @@ getConfigs = do
       Just htmlE -> return htmlE
     dataset <- getDataset e
     (,)
-      <$> get dataset ("obelisk-executable-config-inject-key" :: Text)
+      -- the key is camelCased: https://html.spec.whatwg.org/multipage/dom.html#dom-dataset
```
ali-abrar: delightful

PR #486, opened 2019-07-28
Add 'entity_'

alexfmpe:


lib/beam/src/Obelisk/Database/Beam/Entity.hs
```diff
@@ -59,6 +60,16 @@ deriving instance Ord (Columnar f k) => Ord (Id k f)
 deriving instance Show (Columnar f k) => Show (Id k f)
 deriving instance Read (Columnar f k) => Read (Id k f)
 
+entity_ ::
```
kmicklas: I feel like this name is not adequately scary for sneakily inserting a default value. Maybe `defaultIdEntity`?

PR #487, opened 2019-07-28
Add isomorphism between 'EntityKey' and 'Columnar'

alexfmpe:
Names need some bikeshedding

### DIFF
--- a/lib/beam/src/Obelisk/Database/Beam/Entity.hs
+++ b/lib/beam/src/Obelisk/Database/Beam/Entity.hs
@@ -88,8 +88,8 @@ unsafeUnaryEKEncoder
   :: KeyT v ~ Id k
   => (Show k, Read k)
   => (Applicative check, MonadError Text parse)
-  => Encoder check parse (EntityKey v) PageName
-unsafeUnaryEKEncoder = singlePathSegmentEncoder . unsafeTshowEncoder . isoEncoder isoEK
+  => Encoder check parse (EntityKey v) Text
+unsafeUnaryEKEncoder = unsafeTshowEncoder . isoEncoder isoEK
 
 checkedEntity
   :: Text -- ^ The table name in the schema.
--- a/lib/beam/src/Obelisk/Database/Beam/Entity.hs
+++ b/lib/beam/src/Obelisk/Database/Beam/Entity.hs
@@ -59,6 +59,25 @@ deriving instance Ord (Columnar f k) => Ord (Id k f)
 deriving instance Show (Columnar f k) => Show (Id k f)
 deriving instance Read (Columnar f k) => Read (Id k f)
 
+fromEK
+  :: KeyT value ~ Id k
+  => EntityKeyT f value -> Columnar f k
+fromEK (EntityKey (Id x)) = x
+
+toEK
+  :: KeyT value ~ Id k
+  => Columnar f k -> EntityKeyT f value
+toEK = EntityKey . Id
+
+isoEK
+  :: KeyT value ~ Id k
+  => Iso
+       (EntityKeyT f1 value)
+       (EntityKeyT f2 value)
+       (Columnar f1 k)
+       (Columnar f2 k)
+isoEK = iso fromEK toEK
+
 checkedEntity
   :: Text -- ^ The table name in the schema.
   -> KeyT e (CheckedFieldModification (EntityT e))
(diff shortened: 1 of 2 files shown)


alexfmpe:
Actually, how about renaming `EntityT` to `RowT` ?

3noch:
I'm curious what your reasoning is for that suggestion.

alexfmpe:
I find `EntityT` a very vague name. `RowT` is more obvious, given that what it does is exactly turning some value which exists independently of the database (e.g. frontend only) into something you can read and insert as a row of a database table.

In a personal project I use 
```haskell
data EntityT (value :: (* -> *) -> *) (f :: * -> *) = Entity
  { _entity_key :: KeyT value f
  , _entity_metadata :: MetadataT f
  , _entity_value :: value f
  } deriving Generic
```
where `MetadataT` contains data about the database table's *row* (e.g. created/updated timestamps)

In fact, I'd prefer 'entity' to be used for the wrapped type: `_row_entity`. 
The term 'value' in this context is also confusing because it doesn't match Beam's use of the term which usually means we're working with `Columnar Identity` : https://hackage.haskell.org/package/beam-core-0.8.0.0/docs/Database-Beam-Query.html#v:insertValues

3noch:
I see. Both seem arbitrary to me. The original `EntityT` seems like it deserves a more specific name, as is. Perhaps `KeyedT`. And at this point we're no longer talking about tables directly, so the `T` makes no sense, thus `Keyed` seems ideal. It does exactly one thing: extend some inner type with a key.

kmicklas:
I don't like `Row` because I often associate multiple levels of sub-rows with the same key type. So instead of `MetadataT` I have various things like `WithMetadataT` and `type KeyT (WithMetadataT value) = KeyT value`. This allows me to forget either the key or the metadata or both very easily with lenses.

I like `KeyedT` though, that might be more clear.

@alexfmpe I'll take a look at all this tonight. I've actually been slowly working on an improved version of this library which I'd like to release to Hackage soon now that it's a bit more stable. I'm happy to review changes to this branch but I don't think this really belongs in Obelisk long term.

3noch:
I guess the `T` indicates `Columnar` fields, so I suppose it makes sense after all.

lib/beam/src/Obelisk/Database/Beam/Entity.hs
```diff
@@ -59,6 +65,32 @@ deriving instance Ord (Columnar f k) => Ord (Id k f)
 deriving instance Show (Columnar f k) => Show (Id k f)
 deriving instance Read (Columnar f k) => Read (Id k f)
 
+fromEK
+  :: KeyT value ~ Id k
+  => EntityKeyT f value -> Columnar f k
+fromEK (EntityKey (Id x)) = x
```
kmicklas: I sort of honestly prefer to just define the `Iso` as `entityIdKey` or something like that and only use the lens combinators.

lib/beam/src/Obelisk/Database/Beam/Entity.hs
```diff
@@ -59,6 +65,32 @@ deriving instance Ord (Columnar f k) => Ord (Id k f)
...
+  :: KeyT value ~ Id k
+  => Iso
+       (EntityKeyT f1 value)
+       (EntityKeyT f2 value)
+       (Columnar f1 k)
+       (Columnar f2 k)
+isoEK = iso fromEK toEK
+
+unsafeUnaryEKEncoder
+  :: KeyT v ~ Id k
```
kmicklas: I'd call this `unsafeTshowIdKeyEncoder`.

lib/beam/src/Obelisk/Database/Beam/Entity.hs
alexfmpe: Makes sense. Though, `entityIdKey` or `entityKeyId` ?

lib/beam/src/Obelisk/Database/Beam/Entity.hs
alexfmpe: Ah right, `Key` already implies single column

kmicklas: I used `Id` to mean a single column, `Key` can be multiple columns.

Ericson2314:
https://github.com/kmicklas/beam-keyed is the WIP library btw.

JBetz:
Closed in favor of https://github.com/kmicklas/beam-keyed. #485 and #486 have also been closed for the same reason.

Issue #476, opened 2019-07-17
Make shell escaping more robust


3noch:
As @vaibhavsagar pointed out in #475 we should do better escaping shell arguments.

3noch:
Libraries @vaibhavsagar found that might help:
  * http://hackage.haskell.org/package/shell-escape
  * http://hackage.haskell.org/package/posix-escape

414owen:
So for the actual commands we're running, we're not going through a shell, we're launching processes directly (the difference is between calling System.Process.proc vs System.Process.shell), so this escaping is just necessary to show a process and its args as if it had been launched through a shell.

madeline-os:
This seems to have been implemented. Let me know if I am misunderstanding.

PR #489, opened 2019-08-04
 ob run: Query local packages from Nix project

rvl:
This enhances the `getLocalPkgs` function so that packages defined in the Nix project get added to the GHCi config.

This PR is slightly related to #475. The `readProcessJSONAndLogStderr` function could also be used there.


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -88,21 +86,44 @@ run = do
       , "Frontend.frontend"
       ]
 
+-- | Nix syntax requires relative paths to be prefixed by @./@ or
+-- @../@. This will make a 'FilePath' can be embedded in a Nix
```
vaibhavsagar:
```suggestion
-- @../@. This will make a 'FilePath' that can be embedded in a Nix
```

rvl:
Thanks, bit busy for the next few days, but I will fix the typo noted. Also I realised that the extra local packages must be added to `shells.ghc` and `shells.ghcjs`, so will push a change to `default.nix`. And I will also upload my project which uses this feature.

lib/command/src/Obelisk/Command/Run.hs
3noch: Probably shouldn't block merge at this point

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -88,21 +86,44 @@ run = do
...
-getLocalPkgs :: Applicative f => f [FilePath]
-getLocalPkgs = pure ["backend", "common", "frontend"]
+-- | Relative paths to local packages of an obelisk project.
+--
+-- These are a combination of the obelisk predefined local packages,
+-- and any packages that the user has set with the @packages@ argument
+-- to the Nix @project@ function.
+getLocalPkgs :: (MonadObelisk m, MonadIO m) => FilePath -> m [FilePath]
+getLocalPkgs root = do
+  root' <- liftIO $ makeAbsolute root
```
3noch: Why not use `toNixPath` here since you went to the trouble of pulling out its logic into a named function?

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -88,21 +86,44 @@ run = do
...
+  projectPackages <- readProcessJSONAndLogStderr Debug $
+    proc "nix"
+      [ "eval"
+      , "(let proj = import " <> root' <> " {}; in (map toString (proj.obelisk.reflex-platform.nixpkgs.lib.attrValues proj.passthru.packages)))"
```
3noch: What do you think of slapping an `or {}` after `proj.passthru.packages` just in case that attr path isn't valid? I'm not sure I want obelisk to completely stop working if this changes. However, I also don't like silent "failure" so I'm on the fence. Possibly even adding an explicit `or error "proj.passthru.packages is not defined in your default.nix but is needed by obelisk"`.

lib/command/src/Obelisk/Command/Run.hs
rvl: Fixed, thanks.

lib/command/src/Obelisk/Command/Run.hs
rvl: We need an absolute path here because nix always returns absolute paths for the project packages. I added a comment.

lib/command/src/Obelisk/Command/Run.hs
rvl: I added the defensive `proj.passthru.packages or {}`. But actually it's always defined because of a default parameter value.

rvl:
Thanks for the reviews. I have fixed everything and rebased.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -88,21 +86,45 @@ run = do
...
+-- and any packages that the user has set with the @packages@ argument
+-- to the Nix @project@ function.
+getLocalPkgs :: (MonadObelisk m, MonadIO m) => FilePath -> m [FilePath]
+getLocalPkgs root = do
+  root' <- liftIO $ makeAbsolute root
+  -- package paths from nix project will be absolute paths
+  projectPackages <- readProcessJSONAndLogStderr Debug $
+    proc "nix"
+      [ "eval"
+      , "(let proj = import " <> root' <> " {}; in (map toString (proj.obelisk.reflex-platform.nixpkgs.lib.attrValues (proj.passthru.packages or {}))))"
```
3noch: I guess here (and elsewhere) we are very sloppy with escaping... we should at least stub it out. There is an issue that's *almost* right for this (#476) but that's for shell escaping, not for nix expression escaping...

3noch: I *assume* this gives non-nix paths for some projects? If not then I'm not sure how it actually helps...

rvl: I have changed it to import `root` (the relative path) instead of `root'` (the absolute path). So if there are spaces or nix syntax in the project directory it shouldn't matter.

> I assume this gives non-nix paths for some projects? If not then I'm not sure how it actually helps...

Could you please clarify what this means? I don't understand.

3noch: I mean that `projectPackages` is a list of paths, but none of them are in the nix store?

3noch:
I'm suspicious of these changes for a reason I've forgotten. Need to go back and remember....

3noch:
Oh now I remember. Pulling in *all* of the packages in `packages` is likely going to have bad effects for many obelisk projects. GHCI can only load one set of extensions at once and if any of the modules disagree then they will fail to load IIRC. If we're going to do this we probably need to use a separate attribute, distinct from `packages` that lets people specify exactly what they want in-scope during `ob repl` etc.

alexfmpe:
I've talked about this sort of thing before with @ryantrinkle, who suggested having only unpacked thunks from `packages` be thrown into GHCI.

alexfmpe:
I think there's also some progress on adding actual multi package support to GHCI.
@Ericson2314 ?

3noch:
Even restricting to unpacked thunks seems too onerous IMO. The problem will surface for a package regardless of how it's retrieved so you will always want to be able to specify which packages get this extra treatment.

alexfmpe:
The way I see it, the point of having a thunk be unpacked is editing it, in which case you'd want it thrown in more often than not.
Though I don't see a problem with some `ghciPackages` option that'd default to (unpacked or not) `packages`

3noch:
Yes the point is editing, but you *definitely* need a way to make it not get loaded into GHCi or unpacking might break ob run...

Ericson2314:
I might like see a 3-way "definitely build", "definitely ghci", and "do automatically based in unpackedness"

ryantrinkle:
Regarding "one set of extensions" and such: what if we inject a preprocessor (`-pgmL` or similar) that injects `LANGUAGE` pragmas and/or `OPTIONS_GHC` pragmas at the top of each file.  Is there anything we couldn't do that way?

3noch:
That's a very interesting idea.

3noch:
We're going to try this out via preprocessor and discriminating on packedness of dependencies.

rvl:
The reason that I made this PR is that, in addition to the standard 3 local packages (`frontend`, `backend`, `common`), I had an additional `mypackage`, that I needed to be loaded into `ob repl`. My simplifying assumption was that `mypackage` would share the same flags and default extensions as the other local packages. Additional non-local package dependencies would still go into the project's haskellPackages `overrides` as before.

3noch:
@rvl A lot of people have been rooting for this PR. We really appreciate your work! Sorry it's taken so long to get attention. You're right that on average the cabal settings for multiple packages will probably work well together. The difficulty has been: "How do we get this to a point where we can confidently release it to *everyone* by default?" I think the preprocessor idea finally unlocks that piece of the puzzle because we can actually load each module the way it was meant to be loaded and not have to cross our fingers. Do you foresee issues with this plan?

rvl:
The preprocessor idea sounds like more work to implement but the end result will be better for users. It will also help in the case where users have (for example) modified default extensions in `frontend.cabal` but not in the other two cabal files.

Ericson2314:
Eventually GHC will have proper support for this so it's only temporary complexity. We've had trouble with different settings even just within frontend, common, backend, so I think something like this is diffinitely worth it in the short term.

rvl:
@3noch It's looking good, thanks!

3noch:
@rvl This is now in the `qa` branch to undergo more rigorous testing. Please report any issues you find!

PR #500, opened 2019-08-22
Bump reflex-platform (now uses ghc 8.6.5)

ali-abrar:


3noch:
You deleted `github.json`. Is that how you bump it??

ali-abrar:
> You deleted `github.json`. Is that how you bump it??

That would be something.

PR #503, opened 2019-08-23
Add MonadFail constraints in obelisk-cliapp for GHC 8.6

kmicklas:


Ericson2314:
Definitely seems like progress; can revist whether we want `MonadFail` or something else later.

PR #507, opened 2019-09-04
Expose exe.backend, exe.frontend, exe.assets

### LINKED ISSUE #506: The `exe` attribute seems to have subtle differences from `ghc.backend` and `ghc.frontend`
I can do `nix-build -A exe` and the result will come completely from cache.  But then I follow that up with `nix-build -A ghc.backend` or `nix-build -A ghc.frontend` and the result is not cached.  Shouldn't the former have already built the latter?

3noch:
Fixes #506 

(Not really fixes...but creates a way to get what @mightybyte wanted)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -238,15 +238,21 @@ in rec {
     };
   };
 
-  serverExe = backend: frontend: assets: optimizationLevel: version:
-    pkgs.runCommand "serverExe" {} ''
+  serverExe = backend': frontend': assets': optimizationLevel: version:
+    let
+      backend = haskellLib.justStaticExecutables backend';
+      frontend = compressedJs frontend' optimizationLevel;
+      assets = mkAssets assets';
+    in pkgs.runCommand "serverExe" {} ''
       mkdir $out
       set -eux
       ln -s "${haskellLib.justStaticExecutables backend}"/bin/* $out/
-      ln -s "${mkAssets assets}" $out/static.assets
-      ln -s ${mkAssets (compressedJs frontend optimizationLevel)} $out/frontend.jsexe.assets
+      ln -s "${assets}" $out/static.assets
+      ln -s ${mkAssets frontend} $out/frontend.jsexe.assets
       echo ${version} > $out/version
-    '';
+    '' // {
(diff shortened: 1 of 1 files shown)


3noch:
I'm still testing this...

3noch:
Ok I've successfully gotten these subattrs from cache.

ali-abrar:
Did @mightybyte actually want exe.frontend and exe.backend to be exposed? I thought the issue was that building `exe` didn't build ghc.backend and ghc.frontend.

3noch:
@ali-abrar I *interpreted* his ticket as "The thing I thought was inside `exe` is actually not...so how do I get that?"

3noch:
However, I have *personally* wanted this in the past so I can build only the GHCJS half, for example.

mightybyte:
Hmmmm, I'm not exactly sure how all these things relate to each other.  Maybe that's part of the problem.  Some documentation on the structure of the attributes returned by an obelisk project would probably go a long way to helping with this whole constellation of concerns.  I think what I was really wanting when I opened issue #506 was to have a single `nix-build` command that would reliably build all the things so that my CI would have everything cached that my team might want to build.  At that time, I thought that was `nix-build -A exe`.  Is that the case?

If having this all in a single `nix-build` command doesn't make sense, then I would like to know what build commands I need to run to have everything I need cached.

3noch:
@mightybyte What do you think of #510? That's my attempt at the root issue.

3noch:
@mightybyte As regards `ghc.backend` and `ghc.frontend`. Nothing actually needs those in normal usage. They are effectively what you get when you run `ob run` but of course `ob run` doesn't actually *build* them in the `nix-build` way.

ali-abrar:
> However, I have _personally_ wanted this in the past so I can build only the GHCJS half, for example.

How do you then use the ghcjs half?

3noch:
If frontend-only you'd just open it in a browser. Or maybe you want to deliver the JS differently/separately from how ob deploy does it.

madeline-os:
Closed in favor of the merged #968

PR #510, opened 2019-09-05
RFC: Documented nix targets

3noch:
In skeleton, running `nix-build -A help` produces: 

```
error: Help:

The following targets are documented:

exe - A full, deploy-ready build of the application for the native paltform
exe.assets - Deploy-ready asset directory
exe.backend - Deploy-ready backend executable
exe.frontend - Deploy-ready (optimized) frontend JavaScript build
linuxExe - A full, deploy-ready build of the application for a Linux platform
linuxExe.assets - Deploy-ready asset directory
linuxExe.backend - Deploy-ready backend executable
linuxExe.frontend - Deploy-ready (optimized) frontend JavaScript build

obelisk: The project-specific version of obelisk
server - Function used by 'ob deploy' to build a full server for this project. Arguments: `hostName`, `adminEmail`, `routeHost`, `enableHttps`, `version`
```

If we like this approach I will add more documentation and also *document* this feature in the README etc.


ali-abrar:
Does this need to be on top of #507? It would be a lot easier to accept as a separate change.

How will we know when new help docs need to be written? Can we add a test?

3noch:
No it was mostly an accident. But this is an RFC at this point so I just wanted to get the idea out there for you and others to consider.

3noch:
@ali-abrar I guess you would know that you needed to add docs when you added a target that you expected people to use semiregularly...

About adding a test.... any ideas?

3noch:
@ali-abrar I guess adding tests to this seems strange to me. I could see something like adding "documentation coverage", like "90% of our nix-build targets must be documented" or similar. I could modify `help.nix` to provide this kind of information. But it seems overkill to me at this point. We don't have such tests for our haddock documentation, so why require it for our nix build targets?

3noch:
> How will we know when new help docs need to be written?

I suppose my answer to this is: The same way we know we need a better README. Intuition and complaint.

ali-abrar:
I think the difference is that the readme doesn't correspond one to one
with something that can be checked automatically. We can inspect the
attributes that our nix file exposes and compare the list to the attributes
we've got help for in an automated way though, right?

3noch:
Yes!

3noch:
I've rebased onto `develop` and made some changes. The output now looks like this:
```
error: Help:

The following targets are available:

exe - Deploy-ready application build for the native system (a dummy is supplied for the version)
ghc: (attrset) GHC builds of any package in the application's dependency tree
ghcjs: (attrset) GHCJS builds of any package in the application's dependency tree
linuxExe - Deploy-ready application build for a Linux system (a dummy is supplied for the version)
linuxExeConfigurable - Function taking a version to create a deploy-ready application build for a Linux system
server - Function used by 'ob deploy' to create a NixOS server with this application installed
shells: (attrset) Shells for development (meant to be used with nix-shell)
shells.ghc - Shell with access to all application dependencies built by GHC
shells.ghcjs - Shell with access to all application dependencies built by GHCJS
```

romefeller:
@madeline-os or @cgibbard you may now merge and close this one.

Ericson2314:
I also do not know whether we want this. `nix-build -A help` is nice, but there are costs to having a one-off documentation system the rest of the Nix community uses. There is also the matter of deciding to what extend this stuff is part of Obelisk's intended public interface or just a leaking implementation detail.

madeline-os:
We are going to keep this as a draft while we examine the tooling other projects are using and figure out what works best for obelisk.

madeline-os:
We are going to keep this as a draft while we examine the tooling other nix projects are using and figure out what works best for obelisk.

PR #512, opened 2019-09-06
WIP: minimalish docker image

danbornside:
minimalish docker image building.  

```bash
export NAME=projectName
export VERSION=$(git rev-parse HEAD)
export CONFIGDIR=/path/to/config
docker load < $(nix-build -A dockerImage --no-out-link --argstr name $NAME --argstr version 
$VERSION )
docker run -p 127.0.0.1:8000:8000/tcp -v "$CONFIGDIR:/var/run/backend/config' "${NAME}:${VERSION}"

```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -237,7 +237,7 @@ in rec {
     };
   };
 
-  dockerImage = args@{exe, name}: let
+  dockerImage = args@{exe, name, version}: let
     dockerImageSetupScript = nixpkgs.dockerTools.shellScript "dockersetup.sh" ''
       set -ex
 
@@ -256,6 +256,7 @@ in rec {
 
   in nixpkgs.dockerTools.buildImage {
     name = name;
+    tag = version;
     contents = [ nixpkgs.iana-etc nixpkgs.cacert ];
     runAsRoot = dockerImageSetupScript;
     keepContentsDirlinks = true;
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
I would suggest avoiding `runAsRoot` if at all possible, because using it means constructing the Docker image with QEMU, which is slower than using `extraCommands` or something along those lines.

default.nix
```diff
@@ -237,6 +237,43 @@ in rec {
...
+      echo 'nobody:::'                             >> /etc/gshadow
+
+      mkdir -p    /var/run/backend
+      ln -sft /var/run/backend '${exe}'/*
+      ${nixpkgs.findutils}/bin/find /var/run/backend
+
+      chown 99:99 /var/run/backend
+    '';
+
+  in nixpkgs.dockerTools.buildImage {
```
3noch: Would be nice to expose these arguments as an attrset so that people could use them as a basis for a customized image.

3noch:
Can we make this more composable? I.e. by exposing the building blocks?

ali-abrar:
What needs to happen for us to remove the WIP?

danbornside:
> What needs to happen for us to remove the WIP?
Wanted to verify that it worked for a few people.  It does.  I think this is Good Enough™.

3noch:
Like I said, I'd like to see the guts of this at least pulled out into a function so I can get the "how to create the image" without the image itself. Then I can customize the image if I need to (which is likely).

ingenieroariel:
`version` is required - would it make sense to make it optional and default to `latest` ?

ingenieroariel:
I tested the latest version of the PR locally using NixOS 19.03 and works really well - awesome job!

PR #520, opened 2019-09-07
WIP: obelisk-command: Allow loading local packages into repl + override backend/frontend in 'run'

tomsmalley:
This commit allows arguments to be passed to `ob run/watch/repl` to
load and import extra local packages and modules, plus allows overriding
the default 'Frontend.frontend' and 'Backend.backend' used in 'ob run'.

It assumes the packages follow the same convention as backend/frontend: that is, they are in a directory of the same name.

Now you can do things like:
```
ob watch --import desktop:Desktop
ob run --import desktop:Desktop --frontend Desktop.desktopFrontend
```

### DIFF
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -13,8 +13,9 @@ import qualified Data.Binary as Binary
 import Data.Bool (bool)
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Char as C
 import Data.List
-import Data.Maybe (catMaybes)
+import Data.Maybe (catMaybes, listToMaybe)
 import qualified Data.Text as T
 import Data.Text.Encoding
 import Data.Text.Encoding.Error (lenientDecode)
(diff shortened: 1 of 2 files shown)


3noch:
IIRC we are trying to avoid kinds of configuration that could lead to "works on my machine." I'd there a way this could be accomplished from the nix configuration instead?

3noch:
See #475 for example of what I mean.

tomsmalley:
Maybe? I’m not sure I follow. This isn’t intended for loading thunked stuff, rather libraries that sit alongside frontend/common/backend

ali-abrar:
What does ob deploy do with these libraries that sit alongside frontend/backend/common?

Ericson2314:
Why do extra libraries need a special deployment story? Aren't they just like any other dep unless one does this special ghci-ing?

alexfmpe:
#489 seems like it could handle the "extra libraries" part - then we'd just need a way to override the frontend to be run

3noch:
This will almost entirely be handled by #489. Being able to switch which `frontend` gets used seems like a different design question that I'm not sure how to address yet. If you have a compelling use case, could you open an issue about it?

PR #521, opened 2019-09-09
WIP: read configs from environment variables.

danbornside:
This PR provides an alternative to files for passing cofiguration into obelisk, using environment variables. 

The bulk of the code here deals with the more limited alphabet available to environment variables compared with file paths.  underscores are used as a sort of escape character.  Forward slashes, hyphens, underscores and periods are replaced with `_`, `__h`, `__u`, and `__d`, respectively.  Any ascii code can be repaced with `__x` followed by two hex digits.

### DIFF
--- a/lib/executable-config/lookup/obelisk-executable-config-lookup.cabal
+++ b/lib/executable-config/lookup/obelisk-executable-config-lookup.cabal
@@ -24,7 +24,10 @@ library
       else
         hs-source-dirs: src-other, src
         other-modules: Obelisk.Configs.Internal.Directory
+                     , Obelisk.Configs.Internal.Environment
         build-depends: directory
+                     , attoparsec
+                     , witherable
 
   build-depends:
     base, 
--- a/lib/executable-config/lookup/src-other/Obelisk/Configs/Internal/Environment.hs
+++ b/lib/executable-config/lookup/src-other/Obelisk/Configs/Internal/Environment.hs
@@ -1,0 +1,116 @@
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Obelisk.Configs.Internal.Environment where
+
+import Control.Applicative
+import Control.Monad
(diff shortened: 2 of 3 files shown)


3noch:
An alternative design would be to create something akin to `NIX_PATH`, like `OBELISK_PATH` that just lets you set variables in a more key-value map like fashion: `OBELISK_PATH=config/common/rotue=http://localhost:8000`. I can't say I like this alternative better. I do like it better in the cases where we'd have to do silly "URI-encoding" in our heads to get the right variable name. But in the average case, simply translating `/` to `_` seems nicer. But perhaps a prefix should be added to all of them, like `OBELISK_CONFIG_COMMON_ROUTE`.

3noch:
I see you already added `OBELISK_CONFIG_` to the beginning... nice.

ryantrinkle:
If we don't expect a single obelisk backend to read some configs from files and some configs from environment variables, then a better approach would be to tell it where to get things from, rather than trying both and then combining.  It's good that this patch disallows overlapping configs.  However, it'd be even better to use the fact that we know where configs *should* be coming from to just get them from there.

3noch:
Should we tell it where to get the configs in a file or an environment variable? 😜

ryantrinkle:
@3noch It should be pluggable at build time, in the deployment configuration repo.

3noch:
@ryantrinkle Sorry I'm confused. The deployment config repo sounds like a runtime config, not a build-time one.

ryantrinkle:
@3noch The deployment config repo is what you get when you do `ob deploy init`.  It can feed parameters into the build process.

The key thing here is: when should the choice of where to get the configs be decided?  Suppose someone has a single obelisk app, and they'd like to make one deployment to a bare NixOS machine and another by deploying a docker container.  These are two separate deployments of the same app, so the app's source repository should not be any different.  However, the deployment config repo should know where the deployment is going to be placed, so it is able to know how the configs will be retrieved in that environment.

3noch:
I think one reason this hasn't been merged is because there isn't much UX data on what users actually need here so it's somewhat premature to pick a design.

PR #526, opened 2019-09-14
Fix selftest

alexfmpe:


lib/selftest/src/Obelisk/SelfTest.hs
```diff
@@ -166,9 +165,6 @@ main = do
           it ("can enter "    <> shell) $ inTmpObInit $ \_ -> inShell "exit"
           it ("can build in " <> shell) $ inTmpObInit $ \_ -> inShell $ "cabal new-build --" <> fromString compiler <> " all"
 
-        it "can build reflex project" $ inTmpObInit $ \_ -> do
```
alexfmpe: Fails with either `error: Android builds are not supported on this platform.` or `error: iOS builds are not supported on this platform.`.

default.nix
```diff
@@ -128,7 +128,7 @@ in rec {
   obelisk = ghcObelisk;
   obeliskEnvs = ghcObeliskEnvs;
   command = ghcObelisk.obelisk-command;
-  shell = pinBuildInputs "obelisk-shell" ([command] ++ commandRuntimeDeps pkgs) [];
```
alexfmpe: Not sure what this `[]` was supposed to do, but it causes `error: attempt to call something which is not a function but a set`

3noch: Probably some older nixpkgs thing that went away when bumped.

lib/selftest/src/Obelisk/SelfTest.hs
3noch: Yeah this is not my favorite design that `nix-build` doesn't work without `-A`. Perhaps we can add a `-A all-supported-platforms` or something that just filters out anything that doesn't get through `tryEval`.

3noch: Or something much less hacky...

3noch: I think that would be a better solution that removing this test case. What do you think?

ali-abrar:
Thanks

PR #532, opened 2019-10-04
Small cleanups of Nix

Ericson2314:


release.nix
```diff
@@ -58,30 +61,36 @@ let
...
+    skeleton = import ./skeleton { inherit obelisk; };
+    serverSkeletonExe = skeleton.exe;
+    # TODO fix nixpkgs so it doesn't try to run the result of haskell shells as setup hooks.
+    serverSkeletonShell = local-self.nixpkgs.runCommand "shell-safe-for-dep" {} ''
+      touch "$out"
+      echo "return" >> "$out"
+      cat "${skeleton.shells.ghc}" >> "$out"
+    '';
+    androidSkeleton = (import ./skeleton { inherit obelisk; }).android.frontend;
+    iosSkeleton = (import ./skeleton { inherit obelisk; }).ios.frontend;
```
ali-abrar: `system` needs to be overridden here

release.nix
```diff
@@ -45,7 +48,7 @@ let
   in pkgAttrs: builtins.concatLists (map extractDeps (builtins.attrValues pkgAttrs));
 
   perPlatform = lib.genAttrs cacheBuildSystems (system: let
-    obelisk = import ./. { inherit system; };
+    obelisk = import ./. (self-args // { inherit system; });
```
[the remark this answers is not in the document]
Ericson2314: I don't think so because we want to CI each platform. If you just want to CI one you should probably change `cacheBuildSystems` instead rather than forcing the attributes to lie.

release.nix
```diff
@@ -58,30 +61,36 @@ let
...
-    iosObelisk = import ./. (self-args // {
-      system = "x86_64-darwin";
-      iosSdkVersion = "10.2";
-    });
-    iosSkeleton = (import ./skeleton { obelisk = iosObelisk; }).ios.frontend;
+    command = obelisk.command;
+    skeleton = import ./skeleton { inherit obelisk; };
+    serverSkeletonExe = skeleton.exe;
+    # TODO fix nixpkgs so it doesn't try to run the result of haskell shells as setup hooks.
+    serverSkeletonShell = local-self.nixpkgs.runCommand "shell-safe-for-dep" {} ''
```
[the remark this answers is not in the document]
Ericson2314: stdenv.sh in nixpkgs will do a bash `source` of any depedendeny which is a file rather than directory (ugh!). `nix-shell` builds a single file, which isn't bash, and that blows up the build.

release.nix
Ericson2314: It actually doesn't because we only include `iosSkeleton` for the platforms that cross compiler to iOS. `system` for us is the build platform, not host platform.

ali-abrar: Will this result in ios being built on a linux machine with a darwin remote builder?

Ericson2314: On Linux machine
```
nix-repl> x86_64-linux.profiled.iosSkeleton
error: attribute 'iosSkeleton' missing, at (string):1:1

nix-repl> x86_64-linux.profiled.iosSkeleton
error: attribute 'iosSkeleton' missing, at (string):1:1

nix-repl> x86_64-linux.unprofiled.iosSkeleton
error: attribute 'iosSkeleton' missing, at (string):1:1

nix-repl> x86_64-darwib.unprofiled.iosSkeleton
# starts getting a whole bunch of cabal2nix from darwin cache
```

So the key points are

1. `release.nix` doesn't meaningfully depend on `builtins.currentSystem`; the platforms are what they say
2. iOS builds are not done on Linux builders
3. Anyone can ask a Darwin remote builder to do an iOS build.

PR #534, opened 2019-10-08
CI profiled builds too

Ericson2314:


3noch:
The CI failures might be exposing real issues since this is building so much more than before...

Ericson2314:
Indeed. I was hoping to rule some things out with https://github.com/reflex-frp/reflex-platform/pull/603, but there might be other obelisk-only failures that have sneaked through.

PR #537, opened 2019-10-11
Update the README by fixing spacing and adding full stops

mdimjasevic:
This patch updates the README by fixing spacing and adding full stops in the Android section.


ali-abrar:
Thanks!

PR #539, opened 2019-10-16
Expose haskell overlays parameter

danbornside:


3noch:
This is now possible via the `reflex-platform-func` parameter to obelisk. You just have to be a bit more creative with CPS-style.

PR #540, opened 2019-10-18
Add variant of routeLink which scrolls to the top of the page

tomsmalley:
Seems like we have a couple of options:
- replace `routeLink` with the scroll to top version, and introduce a `routeLinkNoScroll` or something to replace it
- replace `routeLink` and remove old behaviour
- leave `routeLink` and add `routeLinkScrollToTop`

I think this should really be the default behaviour of `routeLink`, but I went with the last option since it doesn't break code using it (note the additional `Prerender` constraint).

Still TODO: add to changelog after deciding the correct implementation


3noch:
I agree this should be default behavior

3noch:
Instead of breaking code using it, it probably *fixes* code using it

xplat:
This should definitely be the default.  `pushState` without changing scroll position to top would almost always be used by things that *don't* try to look like links.

lpsmith:
Would it be a good idea to parameterize the action to take,  so that one could make something happen reliably before or after a scroll-to-top js call?   I suppose you don't actually need that as long as you also provide a way to avoid the scroll-to-top js call,  and then also provide a `defaultRouteLinkAction` or some such that performs the scroll-to-top js call.   That would avoid proliferating variants of the `routeLink` function,  somewhat,  which I'm also doing in #582.

PR #542, opened 2019-10-26
Dedupe routelink impls

alexfmpe:


JBetz:
Is this dependent on #540?

alexfmpe:
Yes, it targets #540 specifically

Ericson2314:
The idea is merge this first, then simplify #540?

alexfmpe:
Whichever works for me - I just didn't want to push this directly into #540 because it goes through some hoops to prevent the factored functions to have top-level union of constraints. But I couldn't open it against develop until #540 was merged

alexfmpe:
I reached for e.g. `Dict (PostBuild t m)` because `PostBuild t m => ()` seemed awkward, but maybe doing it that way is actually better for the consumer, as long as inference works nicely. That could make the 'inline' constraints seamless in many cases?

alexfmpe:
> maybe doing it that way is actually better for the consumer, as long as inference works nicely. That could make the 'inline' constraints seamless in many cases?

Welp, turns out I don't actually know how to achieve this. Right off the bat I get
```
src/Obelisk/Route/Frontend.hs:(529,6)-(537,8): error:
    • Illegal qualified type: Prerender js t m => ()
      GHC doesn't yet support impredicative polymorphism
```
and even if I add the questionable `{-# LANGUAGE ImpredicativeTypes #-}`, inference fails with errors I don't understand.
Given that `Dict` 'just works', I'm in favor of keeping it around.
I was thinking `routeLinkImpl` is also unlikely to be used directly often, which made this is a non-issue, but with the proliferation of variants that was added by the combination of #540 #547, I'm not so sure anymore

3noch:
I'd like to see us try to write this in terms of combinators instead of passing around `Dict`s. Maybe that's not possible? But I'd much prefer that over the complex type signature of `routeLinkImpl`

PR #543, opened 2019-10-27
Add some new functionality to obelisk-route to support simple paramet…

ryantrinkle:
…ers in the URL path

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -21,6 +21,8 @@
 module Obelisk.Route
   ( R
   , pattern (:/)
+  , (:.)
+  , pattern (:.)
   , hoistR
   , PageName
   , PathQuery
@@ -43,6 +45,9 @@ module Obelisk.Route
   , checkEnum1EncoderFunc
   , unitEncoder
   , pathOnlyEncoder
+  , addPathSegmentEncoder
+  , pathParamEncoder
+  , pathLiteralEncoder
   , singletonListEncoder
   , unpackTextEncoder
   , prefixTextEncoder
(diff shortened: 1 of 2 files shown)


lib/route/src/Obelisk/Route.hs
```diff
@@ -200,6 +205,48 @@ mapSome f (Some a) = Some $ f a
...
+--------------------------------------------------------------------------------
+-- Dealing with pairs (i.e. non-dependently-typed subroutes/paths)
+--------------------------------------------------------------------------------
+
+infixr 5 :.
+type (:.) = (,)
+
+{-# COMPLETE (:.) #-}
+pattern (:.) :: a -> b -> a :. b
+pattern a :. b = (a, b)
```
3noch: Note that this does conflict with the same name in `postgresql-simple`.

lib/route/src/Obelisk/Route.hs
```diff
@@ -200,6 +205,48 @@ mapSome f (Some a) = Some $ f a
 hoistR :: (forall x. f x -> g x) -> R f -> R g
 hoistR f (x :=> Identity y) = f x :/ y
 
+--------------------------------------------------------------------------------
+-- Dealing with pairs (i.e. non-dependently-typed subroutes/paths)
+--------------------------------------------------------------------------------
+
+infixr 5 :.
+type (:.) = (,)
```
3noch: What value does this bring over just using tuples?

3noch: It looks like it's only used in one place?

ryantrinkle: When people use it right-associatively, it looks nicer than nested tuples.  Without something like this, people will use tuples with more entries than pairs, which can't be broken down very easily by other primitives.

3noch: Ah makes sense thanks.

PR #551, opened 2019-11-12
Fix 'ghcid-this-lib' script

alexfmpe:


vaibhavsagar:
The created `.ghc.environment.*` files don't cause any issues?

alexfmpe:
I don't remember

ali-abrar:
Sounds like this might need some QA before merging.

vaibhavsagar:
I think this is a good change regardless, but it would be nice if we could pass a flag or something to disable those files being generated.

alexfmpe:
This is what develop was giving me
```
$ pwd
/home/alexfmpe/repos/obelisk/lib/command


$ git clean -fdx
Removing dist-newstyle/
Removing dist/


$ ../../scripts/ghcid-this-lib .
building '/nix/store/wsrxmiism0s42gy0hl015k2wp26q9g3b-cabal2nix-obelisk-command.drv'...
installing
Warning: The configure command is a part of the legacy v1 style of cabal
usage.

Please switch to using either the new project style and the new-configure
command or the legacy v1-configure alias as new-style projects will become the
default in the next version of cabal-install. Please file a bug if you cannot
replicate a working v1- use case with the new-style commands.

For more information, see: https://wiki.haskell.org/Cabal/NewBuild

Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
update' to download it.
Resolving dependencies...
Configuring obelisk-command-0.1...
Loading cabal repl ...
Warning: The repl command is a part of the legacy v1 style of cabal usage.

Please switch to using either the new project style and the new-repl command
or the legacy v1-repl alias as new-style projects will become the default in
the next version of cabal-install. Please file a bug if you cannot replicate a
working v1- use case with the new-style commands.

For more information, see: https://wiki.haskell.org/Cabal/NewBuild

cabal: The 'repl' command does not support multiple targets at once.
```

PR #554, opened 2019-11-18
Warn about issue 460

blargg:
Testing in firefox is a bit concerning, since even basic examples will fail. It would be good to warn people about this in the directions so they can expect this issue, and let them know that it will not appear in a full build/deploy.


ali-abrar:
Thanks!

PR #556, opened 2019-11-19
Use --pure shells for ob run

3noch:
`ob repl` already uses `--pure`. I ran into a situation where `ob run` would fail to build the app but `ob repl` worked fine. This makes them consistent and also more resilient to local configuration problems.


alexfmpe:
Do we ever not want `--pure` ? `ob shell` and `ob doc` seem to use `False` for `isPure`

3noch:
I was asking myself that same question... That's a bigger change than I'd want to do in *this* PR. But it's pertinent.

kmicklas:
I think we definitely don't want pure for `ob shell`.

3noch:
@kmicklas Good point.

ali-abrar:
@3noch can you add a note to the changelog? once we have that, this looks good to go.

PR #558, opened 2019-11-20
Add runBackendWithAssetsFromDirectory for custom asset directory

matthewbauer:
By default, we get assets from the current directory. This works well
for some use cases, but when packaging apps, it is nice to have these
directories separate.

### DIFF
--- a/lib/backend/obelisk-backend.cabal
+++ b/lib/backend/obelisk-backend.cabal
@@ -13,6 +13,7 @@ library
                  data-default,
                  dependent-sum,
                  dependent-sum-template,
+                 filepath,
                  lens,
                  mtl,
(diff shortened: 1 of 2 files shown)


3noch:
I'll try to incorporate this in the latest swath of `runBackendWith` changes...

matthewbauer:
> I'll try to incorporate this in the latest swath of `runBackendWith` changes...

Sounds good. For now, this change is not really needed for anything. I think in principle it's good to separate static assets from where you are launching your app from.

3noch:
Actually I think this is covered by other changes now. You can pass your own `StaticAssets` in directly which lets you specify paths.

PR #560, opened 2019-11-21
Support https with ob run

### LINKED ISSUE #522: Support https in `ob run`

JBetz:
Resolves #522. 

This has `ob run` use the protocol in `config/common/route` to determine whether to run the application over http or https. If https, it will check for the required private key and certificate file in a user-level obelisk config directory, and generate a new pair if not already present.

The certificate is self-signed, and therefore the browser will issue a warning about it being an invalid certificate unless it gets added to root, so there are some notes added to the README to address that, as well as the lifespan of the certificate and how to generate a new one.


ali-abrar:
I don't like the idea of putting things in the user's home directory. Is there some other approach we can take here?

ali-abrar:
A user may, for instance, have multiple obelisk apps (on different versions of obelisk) now sharing a single user-level configuration directory. What if they don't agree?

JBetz:
@ali-abrar Another option would be to generate an in-memory certificate each time `ob run` is executed rather than storing one in the user's home directory. That also means the user wouldn't have to worry about the certificate expiring. The only downside I can think of with that approach is that there would be no way to add that certificate to root.

3noch:
That definitely makes the behavior more consistent between machines at least

ryantrinkle:
Yes, let's not add anything to the home directory for now.  If we did want the user to be able to add a particular certificate to their root set, we should do it by creating a CA cert in ~/.config/obelisk with a [Name Constraint](https://tools.ietf.org/html/rfc5280#section-4.2.1.10) that restricts it to localhost or something else sensible, so that someone who exfiltrates the key can't MITM absolutely everything for that user.  Then, each `ob run` can generate SSL certs from the CA cert as needed, probably in memory.

One major difficulty with having anything in ~/.config/obelisk is that the user cannot be expected to run a monotonically increasing version of obelisk over time.  If the user checks out a very old obelisk project, `ob run` will hand off to a very old obelisk in order to run it.  That means that any cross-project obelisk state needs to be fully backward- _and forward-_ compatible.

JBetz:
Makes sense. I've updated it to use an in-memory certificate, as it seems like that's our best option for now.

PR #565, opened 2019-11-27
Pin an exact revision of 19.09

Ericson2314:


default.nix
```diff
@@ -249,8 +243,10 @@ in rec {
           (pkgs.path + /nixos/modules/security/acme.nix)
         ];
         nixpkgs.overlays = [
-          (self: super: {
-            simp_le = nixos1909.simp_le;
+          (self: super: let
+            nixos1909 = import (hackGet ./dep/nixpkgs-19.09) {};
```
matthewbauer: @ericson2314 Shouldn't this just be `import ./dep/nixpkgs-19.09 {}`? hackGet means the default.nix is never read.

3noch:
Hah. I don't know how I missed this. I think I'm used to GitLab which more aggressively gives you permalinks and GitHub only does it if you specifically ask for it. It's actually really annoying...

PR #566, opened 2019-11-29
Get ghcid with fix from ghcid#288

### LINKED ISSUE #563: Error messages sometimes get eaten
I stumbled across a sharp edge recently,  that I'd set a soft `ulimit` on my machine that restricts the amount of memory that a single process can map,  and then the error message that reminded me about the `ulimit` never gets displayed.   A demo:

```
$ mkdir demo-error
$ ob init
$ ulimit -S -v 12000000
$ ob run
```

At this point there's a couple of error messages one could possibly get,  more commonly:

```
.ob-wrapped: callCommand: ghcid -W --command='ghci -Wall -ignore-dot-ghci -fwarn-redundant-constraints -no-user-package-db -ghci-script /run/user/1000/ob-ghci-69b85b0cc79a5ee5/.ghci'  --reload=config --outputfile=ghcid-output.txt --test='Obelisk.Run.run 45165 (runServeAsset "/home/lpsmith/ob/four-function-calculator/static") Backend.backend Frontend.frontend' (exit 1): failed
```

Less commonly,  a call to `dlopen` will fail:  (which particular shared object it fails on is nondeterministic)

```
can't load .so/.DLL for: /nix/store/q6sdxxb5xj2j5av7wsf7zhkjri23gr0l-memory-0.14.18/lib/ghc-8.6.5/x86_64-linux-ghc-8.6.5/libHSmemory-0.14.18-9cZN4KO0sHc84TFjg4stB9-ghc8.6.5.so (/nix/store/q6sdxxb5xj2j5av7wsf7zhkjri23gr0l-memory-0.14.18/lib/ghc-8.6.5/x86_64-linux-ghc-8.6.5/libHSmemory-0.14.18-9cZN4KO0sHc84TFjg4stB9-ghc8.6.5.so: failed to map segment from shared object)
```

The error message that finally oriented me in the right direction, I found via `strace -f -s 1024`:   (though arguably,  I should have been reminded of the `ulimit` after discovering the previous message,  because it does mention `mmap` failing, but the dynamic linking sent me in the wrong direction at first.)

```
2569  <... read resumed> "ghc: failed to create OS thread: Cannot allocate memory\n", 8192) = 56
```
For whatever reason(s),  this error message was never displayed,  instead you seem to always get one of the previous two types of messages.

matthewbauer:
This adds a custom ghcid pointing to
https://github.com/ndmitchell/ghcid/pull/288 for better error
messages. Also uses staticWhich to find ghcid for use in ob run. As a
result, we now include ghcid in Obelisk runtime.

Fixes #563

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -40,7 +40,7 @@ let
         });
         hnix-store-core = self.callHackage "hnix-store-core" "0.1.0.0" {};
 
-        ghcid = haskellLib.justStaticExecutables (self.callCabal2nix "ghcid" (hackGet ./dep/ghcid) {});
+        ghcid = self.callCabal2nix "ghcid" (hackGet ./dep/ghcid) {};
       })
 
       pkgs.obeliskExecutableConfig.haskellOverlay
@@ -55,7 +55,12 @@ let
         obelisk-asset-serve-snap = self.callCabal2nix "obelisk-asset-serve-snap" (cleanSource ./lib/asset/serve-snap) {};
         obelisk-backend = self.callCabal2nix "obelisk-backend" (cleanSource ./lib/backend) {};
         obelisk-cliapp = self.callCabal2nix "obelisk-cliapp" (cleanSource ./lib/cliapp) {};
-        obelisk-command = haskellLib.overrideCabal (self.callCabal2nix "obelisk-command" (cleanSource ./lib/command) {}) { librarySystemDepends = [ pkgs.nix self.ghcid ]; };
+        obelisk-command = haskellLib.overrideCabal (self.callCabal2nix "obelisk-command" (cleanSource ./lib/command) {}) {
+          librarySystemDepends = [
+            pkgs.nix
+            (haskellLib.justStaticExecutables self.ghcid)
+          ];
+        };
         obelisk-frontend = self.callCabal2nix "obelisk-frontend" (cleanSource ./lib/frontend) {};
         obelisk-run = self.callCabal2nix "obelisk-run" (cleanSource ./lib/run) {};
         obelisk-route = self.callCabal2nix "obelisk-route" (cleanSource ./lib/route) {};
--- a/default.nix
+++ b/default.nix
@@ -45,6 +45,8 @@ let
           doCheck = false;
         });
         hnix-store-core = self.callHackage "hnix-store-core" "0.1.0.0" {};
+
+        ghcid = haskellLib.justStaticExecutables (self.callCabal2nix "ghcid" (hackGet ./dep/ghcid) {});
       })
 
       pkgs.obeliskExecutableConfig.haskellOverlay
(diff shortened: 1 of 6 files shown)


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -227,7 +227,7 @@ runGhcid
   => FilePath -- ^ Path to .ghci
   -> Maybe String -- ^ Optional command to run at every reload
   -> m ()
-runGhcid dotGhci mcmd = callCommand $ unwords $ "ghcid" : opts
+runGhcid dotGhci mcmd = callCommand $ unwords $ $(staticWhich "ghcid") : opts
```
JBetz: Do you know how this working at all without ghcid being provided to the shell? I don't remember ever needing to install ghcid in order to use `ob run`, both before and after we switched to a pure shell.

matthewbauer: I think reflex-platform provides ghcid for us, so you always got it unless you really messed with the nix project expressions.

3noch:
Merged to `qa`

default.nix
```diff
@@ -39,6 +39,8 @@ let
           doCheck = false;
         });
         hnix-store-core = self.callHackage "hnix-store-core" "0.1.0.0" {};
+
+        ghcid = haskellLib.justStaticExecutables (self.callCabal2nix "ghcid" (hackGet ./dep/ghcid) {});
```
Ericson2314: Better to do this at use site.
```suggestion
        ghcid = self.callCabal2nix "ghcid" (hackGet ./dep/ghcid) {};
```

default.nix
```diff
@@ -53,7 +55,7 @@ let
         obelisk-asset-serve-snap = self.callCabal2nix "obelisk-asset-serve-snap" (cleanSource ./lib/asset/serve-snap) {};
         obelisk-backend = self.callCabal2nix "obelisk-backend" (cleanSource ./lib/backend) {};
         obelisk-cliapp = self.callCabal2nix "obelisk-cliapp" (cleanSource ./lib/cliapp) {};
-        obelisk-command = haskellLib.overrideCabal (self.callCabal2nix "obelisk-command" (cleanSource ./lib/command) {}) { librarySystemDepends = [ pkgs.nix ]; };
+        obelisk-command = haskellLib.overrideCabal (self.callCabal2nix "obelisk-command" (cleanSource ./lib/command) {}) { librarySystemDepends = [ pkgs.nix self.ghcid ]; };
```
Ericson2314:
```suggestion
        obelisk-command = haskellLib.overrideCabal (self.callCabal2nix "obelisk-command" (cleanSource ./lib/command) {}) {
          librarySystemDepends = [
            pkgs.nix
            (haskellLib.justStaticExecutables self.ghcid)
          ];
        };
```

3noch:
@Ericson2314 Good catch.

3noch:
Updated `qa`

PR #567, opened 2019-11-29
WIP: Associate :/ With R

Ericson2314:
Makes importing easier

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -24,8 +24,8 @@ Types and functions for defining routes and 'Encoder's.
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE ViewPatterns #-}
 module Obelisk.Route
-  ( -- * Primary Types
-    R
+  ( R ((:/))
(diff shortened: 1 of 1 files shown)


alexfmpe:
Please yes

3noch:
Does this break modules that import `(:/)` as `pattern (:/)`? If so, we should probably export it both ways if possible.

Ericson2314:
@3noch I believe it doesn't break that, but we should confirm.

3noch:
@Ericson2314 This was based on the wrong branch. It now has conflicts. :/

Ericson2314:
I think it also needs a GHC change!

3noch:
wat?

Ericson2314:
I think normal data constructors can be associated with a type synonym, but not patterns!

madeline-os:
This is waiting on https://gitlab.haskell.org/ghc/ghc/-/issues/19507

ryantrinkle:
If https://gitlab.haskell.org/ghc/ghc/-/issues/19507 is ever closed, please feel free to reopen this.

PR #568, opened 2019-11-30
Add HLint support to ob watch/run

alexfmpe:
On top of #566


3noch:
What if we didn't take an argument but instead inferred whether or not you wanted hlint based on the presence of an `.hlint.ymal` file?

3noch:
The reasoning being: everyone using the repo should get the same exact output of `ob run` if they're on the same commit.

alexfmpe:
> What if we didn't take an argument but instead inferred whether or not you wanted hlint based on the presence of an .hlint.ymal file?

Tricky. hlint will crawl up your directory hierarchy searching for `.hlint.yaml`. That goes beyond the scope of the repo.
I opted for on-by-default for discoverability (easier to do `ob --help` or lookup the Readme when trying to turn it off rather than remembering it might potentially exist).

3noch:
Right. It would only enable it if you had an .hlint.yaml file adjacent to .obelisk/impl.

3noch:
Maybe even better would be a way to configure what your .hlint.yaml file should be via nix. Then you could even thunk it to some "company-wide config" or something....

Issue #70, opened 2018-06-07
Make frontend and backend be libraries only


ryantrinkle:
See the discussion here: https://github.com/obsidiansystems/obelisk/pull/57

`main.hs` for both frontend and backend, as well as `devel.hs`, should be eliminated from the skeleton.  Obelisk should supply this functionality itself.

In the future, we should document how the user can supply their own executables, but that is not necessary right now.

eskimor:
Isn't this a duplicate of #87 ?

ryantrinkle:
Yes; closed that one, since this one has more information.

PR #570, opened 2019-12-02
Simplify frontend executable

### LINKED ISSUE #424: Make frontend executable main.hs simpler/more future-proof
Ideally the frontend and backend executable definitions in the skeleton are as simple and future-proof as possible, because they don't get type-checked/tested by `ob run`.

Currently the frontend skeleton main is:

```haskell
main :: IO ()
main = do
  let Right validFullEncoder = checkEncoder backendRouteEncoder
  run $ runFrontend validFullEncoder frontend
```

I don't see any reason in principle why this can't be made into something like:

```haskell
main :: IO ()
main = runFrontend frontend
```

The backend is already basically this simple.

JBetz:
Resolves #424 by moving as much as possible inside `Obelisk.Frontend`.

### DIFF
--- a/lib/frontend/src/Obelisk/Frontend.hs
+++ b/lib/frontend/src/Obelisk/Frontend.hs
@@ -149,30 +149,33 @@ runFrontend
   .  Encoder (Either Text) Identity (R (FullRoute backendRoute route)) PageName
   -> Frontend (R route)
   -> IO ()
-runFrontend fullRouteEncoder frontend = run $ do
-  let Right validFullEncoder = checkEncoder fullRouteEncoder
-  let mode = FrontendMode
-        { _frontendMode_hydrate =
+runFrontend fullRouteEncoder frontend =
+  case checkEncoder fullRouteEncoder of
+    Right validFullEncoder -> run $ do
+      let mode = FrontendMode
+            { _frontendMode_hydrate =
 #ifdef ghcjs_HOST_OS
-          True
+              True
 #else
-          False
+              False
 #endif
-        , _frontendMode_adjustRoute =
+            , _frontendMode_adjustRoute =
 #ifdef ghcjs_HOST_OS
-          False
+              False
 #else
-          True
+              True
 #endif
-        }
-  configs <- liftIO Lookup.getConfigs
-  when (_frontendMode_hydrate mode) removeHTMLConfigs
-  -- There's no fundamental reason that adjustRoute needs to control setting the
-  -- initial route and *also* the useHash parameter; that's why these are
-  -- separate here.  However, currently, they are always the same.
-  when (_frontendMode_adjustRoute mode) $ do
-    setInitialRoute $ _frontendMode_adjustRoute mode
-  runFrontendWithConfigsAndCurrentRoute mode configs validFullEncoder frontend
+            }
+      configs <- liftIO Lookup.getConfigs
+      when (_frontendMode_hydrate mode) removeHTMLConfigs
+      -- There's no fundamental reason that adjustRoute needs to control setting the
+      -- initial route and *also* the useHash parameter; that's why these are
+      -- separate here.  However, currently, they are always the same.
+      when (_frontendMode_adjustRoute mode) $ do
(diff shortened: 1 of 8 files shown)


lib/frontend/src/Obelisk/Frontend.hs
```diff
@@ -145,10 +146,11 @@ data FrontendMode = FrontendMode
...
 runFrontend
   :: forall backendRoute route
-  .  Encoder Identity Identity (R (FullRoute backendRoute route)) PageName
+  .  Encoder (Either Text) Identity (R (FullRoute backendRoute route)) PageName
   -> Frontend (R route)
-  -> JSM ()
-runFrontend validFullEncoder frontend = do
+  -> IO ()
+runFrontend fullRouteEncoder frontend = run $ do
+  let Right validFullEncoder = checkEncoder fullRouteEncoder
```
3noch: Turn into `error "helpful message"` please :smile:

skeleton/frontend/src-bin/main.hs
```diff
@@ -1,10 +1,6 @@
...
 import Common.Route
 import Obelisk.Frontend
-import Obelisk.Route.Frontend
-import Reflex.Dom
 
 main :: IO ()
-main = do
-  let Right validFullEncoder = checkEncoder fullRouteEncoder
-  run $ runFrontend validFullEncoder frontend
+main = runFrontend fullRouteEncoder frontend
```
alexfmpe: Can't this be something like `main = Frontend.main` instead ?

JBetz: Point, that would be maximally future-proof.

JBetz: Okay, did this for the frontend, as well as the backend to make it symmetric.

3noch:
Need to think more about this.

skeleton/backend/src/Backend.hs
```diff
@@ -4,8 +4,12 @@
 module Backend where
 
 import Common.Route
+import Frontend
 import Obelisk.Backend
 
+backendMain :: IO ()
+backendMain = runBackend backend frontend
```
3noch: I don't feel like moving this here adds anything. Both before and after, it was part of the skeleton.

lib/frontend/src/Obelisk/Frontend.hs
```diff
@@ -145,32 +146,36 @@ data FrontendMode = FrontendMode
...
+      configs <- liftIO Lookup.getConfigs
+      when (_frontendMode_hydrate mode) removeHTMLConfigs
+      -- There's no fundamental reason that adjustRoute needs to control setting the
+      -- initial route and *also* the useHash parameter; that's why these are
+      -- separate here.  However, currently, they are always the same.
+      when (_frontendMode_adjustRoute mode) $ do
+        setInitialRoute $ _frontendMode_adjustRoute mode
+      runFrontendWithConfigsAndCurrentRoute mode configs validFullEncoder frontend
+    Left err ->
+      error $ "Route encoder check failed: " <> show err
```
3noch: I still prefer having this check in the skeleton for two reasons

  1. It makes the error customizable and handling also somewhat customizable
  2. In order to actually *use* a route encoder, you usually have to run `checkEncoder` first anyway, so now we force users to have *two* places in the code where the `checkEncoder` happens.

What I think would be better is to put a `checkedFullRouteEncoder` in the skeleton that has already been checked and use that in `runFrontend`. Users can customize that if they want to but it also makes the actual `main` simpler.

skeleton/backend/src/Backend.hs
alexfmpe: Well, it makes ob repl/watch/run typecheck everything except the trivial `main = backendMain`. Doesn't matter for `ob init` but It's pretty annoying to have a full build fail on this one line rather than at ghci(d) if/when this part of the skeleton changes, which is the main point of #424

3noch:
This will be obviated by fixing https://github.com/obsidiansystems/obelisk/issues/70

PR #572, opened 2019-12-04
Decrease android certificate validity from 1b days to 1m

matthewbauer:
With 1,000,000,000 days, I get this error on ‘ob deploy test android’:

> com.android.ide.common.signing.KeytoolException: Failed to read key
obelisk from store
"/nix/store/4fa5z7arnfkaf6jipj18k062ydkwhzab-android_keystore.jks":
java.io.IOException: Parse Generalized time, invalid format

It looks like the number is too big to be parsed by Android / Java parsing. 10^9, 10^8, 10^7 all fail but 10^6 succeeds. This seems like a weird cut-off, roughly below 2^20. But, 10^6 is still valid for 2737 years.


3noch:
> is still valid for 2737 years.

That should be enough...

ryantrinkle:
Geez guys, I never thought I'd see you endorse planned obsolescence!

PR #577, opened 2019-12-06
Add warning to obeliskEnvs.command which caused major confusion

3noch:


default.nix
```diff
@@ -106,7 +106,9 @@ in rec {
   pathGit = ./.;  # Used in CI by the migration graph hash algorithm to correctly ignore files.
   path = reflex-platform.filterGit ./.;
   obelisk = ghcObelisk;
-  obeliskEnvs = ghcObeliskEnvs;
+  obeliskEnvs = ghcObeliskEnvs // {
```
matthewbauer: Maybe we should just filter ghcObelisk for just the `obelisk-*` attrs? I'm not sure whether we have similar collisions like this.

3noch: Good idea

ingenieroariel:
@Ericson2314 what are your thoughts on this?

default.nix
3noch: I did this instead.

Issue #232, opened 2018-10-01
`ob` command handoff doesn't happen on Fedora without user intervention.


danbornside:
Fedora has, by default, a per-user umask of `0002` instead of the more restrictive `0022`.  so when someone on such a configuration does a git-checkout of an obelisk project, the obelisk thunk ends up with permissions of `775`, but the obelisk command refuses to hand-off if any of those directories (`.`, `.obelisk` or `.obelisk/impl`) are group writable.

It would be much nicer if `ob` honored the umask, checking that traversed directories are at least as restrictive as `umask | 0002`

danbornside:
To add insult to injury, even if you had chmoded the relevant dirs, if you happen to `ob thunk unpack ... ob thunk pack`, the file permissions revert to whatever is determined by the umask, and handoff stops working again.

3noch:
So the solution would be care as much about group-writable as the umask does. We'd still put our foot down about world writable though, as you suggest.

3noch:
This was fixed in #578

PR #578, opened 2019-12-10
Make obelisk respect the system umask

meditans:
Fixes https://github.com/obsidiansystems/obelisk/issues/232

### DIFF
--- a/lib/command/src/Obelisk/Command/Project.hs
+++ b/lib/command/src/Obelisk/Command/Project.hs
@@ -15,6 +15,7 @@ module Obelisk.Command.Project
   , toImplDir
   ) where
 
+import Control.Concurrent.MVar (MVar, newMVar, withMVarMasked)
 import Control.Monad
 import Control.Monad.IO.Class (liftIO)
 import Control.Monad.State
@@ -24,7 +25,8 @@ import qualified Data.Text as T
 import System.Directory
 import System.FilePath
 import System.IO.Temp
-import System.Posix (FileStatus, UserID, deviceID, fileID, fileMode, fileOwner, getFileStatus, getRealUserID)
+import System.IO.Unsafe (unsafePerformIO)
+import System.Posix (FileStatus, FileMode, CMode (..), UserID, deviceID, fileID, fileMode, fileOwner, getFileStatus, getRealUserID)
 import System.Posix.Files
 import System.Process (CreateProcess, cwd, proc, waitForProcess, delegate_ctlc)
 
@@ -34,6 +36,7 @@ import GitHub.Data.Name (Name)
 import Obelisk.App (MonadObelisk)
 import Obelisk.CliApp
 import Obelisk.Command.Thunk
+
 --TODO: Make this module resilient to random exceptions
 
 --TODO: Don't hardcode this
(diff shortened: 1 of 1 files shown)


lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -137,12 +138,29 @@ findProjectObeliskCommand target = do
         ]
       return Nothing
 
+-- We need the umask not be too strict with the group permissions on the folders
+-- Obelisk is using, since some systems have a default umask of 0002 instead of
+-- 0022, and we want to respect that. Since we have to set the umask to read it,
```
3noch: Please put this warning as part of the docs for `getDefaultUmask`.

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -137,12 +138,29 @@ findProjectObeliskCommand target = do
...
+-- because of http://man7.org/linux/man-pages/man2/umask.2.html#NOTES we have to
+-- be careful to call this operation in the initial single-threaded part of the
+-- program.
+
+-- | Get the umask for the Obelisk process.
+getDefaultUmask :: IO FileMode
+getDefaultUmask = do
+  initialMask <- setFileCreationMask safeUmask
+  void (setFileCreationMask initialMask)
+  pure initialMask
```
3noch: I would feel better about this if we had a global `MVar` to make sure this is atomic.

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -193,6 +214,17 @@ walkToImplDir projectRoot myUid = do
...
+isOwnedBy :: FileStatus -> UserID -> Bool
+isOwnedBy s uid = fileOwner s == uid
+
+-- | Check to see if a directory respect the umask, but check explicitly that
+-- it's not world writable in any case.
+respectsUmask :: FileStatus -> FileMode -> Bool
+respectsUmask s umask = isNotWorldWritable && umaskRespected
+  where
+    isNotWorldWritable = fileMode s .&. 0o02 == 0
+    umaskRespected = fileMode s .&. umask == 0
```
3noch: It seems it would be theoretically possible for a file to *not* be world writable but for the umask to make files world writable. In this case we should be fine with the file since it's not *actually* world-writable.

lib/command/src/Obelisk/Command/Project.hs
3noch: This might be one of the very few places where an *actual* global would be appropriate--because the race condition inherent to the system call itself.

```haskell
globalUmaskLock :: MVar ()
globalUmaskLock = unsafePerformIO (newMVar ())
{- NOINLINE globalUmaskLock -}
```

meditans: Perfect, implemented

lib/command/src/Obelisk/Command/Project.hs
meditans: Done

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -199,20 +205,21 @@ walkToProjectRoot this thisStat mDesiredUmask myUid = liftIO (doesDirectoryExist
...
---TODO: Is there a better way to ask if anyone else can write things?
---E.g. what about ACLs?
--- | Check to see if directory is only writable by a user whose User ID matches the second argument provided
-isWritableOnlyBy :: FileStatus -> UserID -> Bool
-isWritableOnlyBy s uid = fileOwner s == uid && fileMode s .&. 0o22 == 0
+-- | Check to see if directory is writable by a user whose User ID matches the
+-- second argument provided, and if the fact that other people can write to that
+-- directory is in accordance with the umask of the system, passed as the third
+-- argument.
+isWellOwnedAndWellMasked :: FileStatus -> UserID -> FileMode -> Bool
```
3noch: "WellMasked" doesn't seem quite right. The mask affects what permissions are assigned by default.  But once the file is created it's no longer using a mask...it just has permissions. Maybe `hasSafePermissions` or something...

3noch:
Merged to `qa`

PR #582, opened 2019-12-18
Expose events from the routeLink functions

lpsmith:
The reflex-frp.org marketing site needs access to some of these events;  this is one of the changes in `reflex-frp.org`'s current branch of obelisk that seems like a small improvement to me.

### DIFF
--- a/lib/route/src/Obelisk/Route/Frontend.hs
+++ b/lib/route/src/Obelisk/Route/Frontend.hs
@@ -556,8 +556,7 @@ dynRouteLink
   -> m a
 dynRouteLink = routeLinkDynAttr mempty
 
-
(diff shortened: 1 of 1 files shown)


3noch:
Thanks. I was wondering when we'd need this :stuck_out_tongue_closed_eyes:

lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -554,6 +556,20 @@ dynRouteLink
   -> m a
 dynRouteLink = routeLinkDynAttr mempty
 
+
+-- | Like 'routeLinkDynAttr' but without custom attributes and also returns the events associatd with the link.
```
alexfmpe: associat***e***d

PR #584, opened 2019-12-18
Start using more robust shell escaping

### LINKED ISSUE #476: Make shell escaping more robust

414owen:
Here, we start passing around typechecked shell-escaped strings for some commands, as well as having better prettified command logs. Unicode is supported and escaped correctly.

If you don't need to do any shell scripting, we should prefer using the safer `*projectProc` functions, over `*projectShell` ones.

closes #476


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -232,15 +234,14 @@ runGhcid dotGhci mcmd = callCommand $ unwords $ $(staticWhich "ghcid") : opts
     opts =
       [ "-W"
       --TODO: The decision of whether to use -fwarn-redundant-constraints should probably be made by the user
-      , "--command='ghci -Wall -ignore-dot-ghci -fwarn-redundant-constraints " <> makeBaseGhciOptions dotGhci <> "' "
+      , "--command='ghci -Wall -ignore-dot-ghci -fwarn-redundant-constraints " <> unwords (makeBaseGhciOptions dotGhci) <> "' "
```
3noch: Should we escape the entire argument to `--comand` here? We're basically passing in a giant shell command IIUC.

lib/command/src/Obelisk/Command.hs
```diff
@@ -382,7 +384,7 @@ ob = \case
   ObCommand_Repl -> runRepl
   ObCommand_Watch -> inNixShell' $ static runWatch
   ObCommand_Shell so -> withProjectRoot "." $ \root ->
-    projectShell root False (_shellOpts_shell so) (_shellOpts_command so)
+    projectProc root False (_shellOpts_shell so) (pure . bash . BSU.fromString <$> _shellOpts_command so)
```
3noch: Let's use `Just` instead of `pure` when we're not gaining anything from the additional polymorphism.

3noch: The same could probably be said for `<$>` being just `map`. Using `<$>` for `map` is a *bit* more common so I'll let you decide that one. If it were me I'd use `map`...

3noch: Lol of course now I realize the `pure` is for list and `<$>` is for `Maybe`... this is why extra polymorphism is a problem IMO: It makes reading code harder.

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -233,6 +237,10 @@ filePermissionIsSafe s umask = not fileWorldWritable && fileGroupWritable <= uma
     fileGroupWritable = fileMode s .&. 0o020 == 0o020
     umaskGroupWritable = umask .&. 0o020 == 0
 
+inProjectProc :: MonadObelisk m => String -> [Bash] -> m ()
+inProjectProc shellName command = withProjectRoot "." $ \root ->
+  projectProc root True shellName (Just command)
```
3noch: Why do we have this function? Why do we have `inProjectShell`? All it adds is `withProjectRoot "."`? Seems like we should either never use this or always use it... Please investigate a bit and let me know what you think. I'd rather not perpetuate the pattern if it's not a good one.

414owen: Right, so the reason we can't always pass around  `[Bash]`, is because sometimes we want to use actual unescaped bash constructs, eg. [this haddock script](https://github.com/obsidiansystems/obelisk/blob/acc5e6db09462d76c7e7eaff3aa3634a7ce0c371/lib/command/src/Obelisk/Command.hs#L397), which wouldn't make sense if it were escaped.

That's why I added a separate typesafe version for simple programs.

lib/command/src/Obelisk/Command/Run.hs
414owen: Fair point

lib/command/src/Obelisk/Command/Project.hs
ingenieroariel: Perhaps we can add the intended use as a comment on that function? Something related to `separate typesafe version for simple programs`?

3noch:
Now that `selftest` actually tests this, I updated to use the latest `qa` and now `selftest` fails. So we need to fix that first.

414owen:
Hmmm, selftest is passing on my machine, with the latest qa.
I'm not running android or ios tests though (I need to move my nix store to a new partition before I can), was it failing on either of those?

```
Finished in 56.8265 seconds
26 examples, 0 failures
~/src/obelisk (use-posix-escape)
^_^ owen@nixos $
```

414owen:
Okay, we're fixing this upstream. Was a parallelization issue in SelfTest.

414owen:
fixed in https://github.com/obsidiansystems/obelisk/pull/608

lib/command/src/Obelisk/Command/Project.hs
3noch: I deleted `inProjectShell` in latest `qa` because it wasn't used in a uniform way. That answers this question in my mind. So we should delete this counterpart here.

lib/command/src/Obelisk/Command.hs
```diff
@@ -398,11 +400,11 @@ ob = \case
   ObCommand_Repl interpretPathsList -> withInterpretPaths interpretPathsList runRepl
   ObCommand_Watch interpretPathsList -> withInterpretPaths interpretPathsList runWatch
   ObCommand_Shell (ShellOpts shellAttr interpretPathsList cmd) -> withInterpretPaths interpretPathsList $ \root interpretPaths ->
-    nixShellForInterpretPaths False shellAttr root interpretPaths cmd
+    nixShellForInterpretPaths False shellAttr root interpretPaths (pure . bash . BSU.fromString <$> cmd)
   ObCommand_Doc shellAttr pkgs -> withProjectRoot "." $ \root ->
-    nixShellForInterpretPaths True shellAttr root emptyPathTree $ Just $ haddockCommand pkgs
+    nixShellForInterpretPaths True shellAttr root emptyPathTree $ Just [haddockCommand pkgs]
```
tomsmalley, line 405: I've yet to test `ob doc`. It doesn't work for the skeleton.

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -270,18 +272,18 @@ filePermissionIsSafe s umask = not fileWorldWritable && fileGroupWritable <= uma
...
     & nixShellConfig_pure .~ isPure
     & nixShellConfig_common . nixCmdConfig_target .~ (def & target_path .~ Nothing)
-    & nixShellConfig_run .~ (command <&> \c -> mconcat
-      [ "export NIX_PATH=nixpkgs=", T.unpack nixpkgsPath, "; "
-      , maybe "" (\v -> "export NIX_REMOTE=" <> v <> "; ") nixRemote
-      , c
+    & nixShellConfig_run .~ (command <&> \cs -> unwords $ concat
+      [ ["export", BSU.toString . bytes . bash $ "NIX_PATH=nixpkgs=" <> encodeUtf8 nixpkgsPath, ";"]
+      , maybe [] (\v -> ["export", BSU.toString . bytes . bash $ "NIX_REMOTE=" <> encodeUtf8 (T.pack v), ";"]) nixRemote
+      , fmap (BSU.toString . bytes) cs
```
3noch, line 286: Does it make sense to just move the `Bash` type into `nixShellConfig_run`?

3noch: Oh, that may be what you did. Is there a way to glob some of these combinations together and make this stuff less noisy?

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -480,17 +482,16 @@ runGhcid
...
-        <> ["--command='" <> unwords ("ghci" : ghciArgs) <> "'"] -- TODO: Shell escape
-    testCmd = maybeToList (flip fmap mcmd $ \cmd -> "--test='" <> cmd <> "'") -- TODO: Shell escape
-
+    opts = concat
+      [ ["-W"]
+      , ["--outputfile=ghcid-output.txt"]
+      , map (\x -> "--reload=" <> x) reloadFiles
+      , map (\x -> "--restart=" <> x) restartFiles
+      , maybe [] (\cmd -> ["--test=" <> cmd]) mcmd
+      , ["--command=" <> unwords ("ghci" : ghciArgs)]
```
3noch, line 493: unwords seems like something we should be escaping *again*?

3noch: Although I see there are no quotes, so maybe this is somehow *flattened*? I thought I tried that before and it didn't work.

lib/command/src/Obelisk/Command/Project.hs
3noch: Hm I don't see that change. We should force `Bash` at the deepest levels.

3noch: That said, we may also need to ensure that things are actually *using* bash. I don't of the top of my head know if `nix-shell --run` always uses bash or tries to be smart and use your shell.

lib/command/src/Obelisk/Command/Run.hs
tomsmalley: This turns into `$'--command=ghci args'`

3noch: Maybe we should be breaking `ghciArgs`?

PR #593, opened 2019-12-24
Add additional result-* links from nix-build to .gitignore

kmicklas:
If you `nix-build` an attribute set you get `result-2`, `result-3`, etc.


3noch:
Thanks!

PR #594, opened 2019-12-25
New thunk types for private and local repos

3noch:


lib/command/src/Obelisk/Command.hs
```diff
@@ -220,20 +221,35 @@ thunkDirectoryParser = fmap (dropTrailingPathSeparator . normalise) . strArgumen
   , help "Path to directory containing thunk data"
   ]
 
+thunkConfig :: Parser ThunkConfig
+thunkConfig = ThunkConfig
+  <$>
+    (   flag' (Just True) (long "private" <> help "Mark thunks as pointing to a private repository")
+    <|> flag' (Just False) (long "public" <> help "Mark thunks as pointing to a public repository")
+    <|> pure Nothing
+    )
```
alexfmpe: Shouldn't we fail parsing if both private and public are provided? This seems like it will pick the first, which could allow someone to accidentally add `public` rather than replace `private` after open-sourcing a repo.
Replacing the order here could then potentially break things

Ericson2314:
I wonder, if the change is fairly big already, should we also solve the problem of thunks always importing the repo root and instead make then work like `hackOn`?

lib/command/src/Obelisk/Command.hs
cardenaso11: this works as you intended, if private and public are both provided, it marks the last one as invalid, because it is not consumed. although, the error message is just that the last one is an invalid option, which might be worth changing

3noch:
I'd be fine with trying to tackle that *in this release set* but not necessarily in this PR. Let's discuss the design in an issue.

3noch:
@Ericson2314 See #601

414owen:
PR that fixes tests here: https://github.com/obsidiansystems/obelisk/pull/603

PR #595, opened 2019-12-25
QA

3noch:


3noch:
Need to add changelog entry about new reflex-platform-func attr in obelisk's default.nix

PR #597, opened 2019-12-27
Add missing error log

alexfmpe:


alexfmpe:
Seems too easy to forget to actually output the error(s). Feels like the return should use `Either [Text]` instead of `Maybe`

PR #599, opened 2019-12-27
Allow packages named differently from the containing directory

alexfmpe:


lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -108,15 +108,16 @@ runWatch = do
 
 -- | Relative paths to local packages of an obelisk project
 -- TODO a way to query this
-getLocalPkgs :: Applicative f => f [FilePath]
-getLocalPkgs = pure ["backend", "common", "frontend"]
+getLocalPkgs :: Applicative f => f [(FilePath, Maybe String)]
+getLocalPkgs = pure [("backend", Nothing), ("common", Nothing), ("frontend", Nothing)]
```
alexfmpe: This is technically the only committed use-case, yet the package-equals-folder assumption was
1) highly unexpected
2) unreported due to #597 
3) inconsistent with obelisk itself, since I ran into this whole thing when adding `.obelisk/impl/lib/route` to `getLocalPkgs` for some experiments, and it wouldn't find `obelisk-route.cabal`

kmicklas: I've definitely run into other confusing issues with the Obelisk libraries having names different than the directories - independent of this I'd prefer they weren't like that.

3noch:
Wow. This is a terrible assumption on obelisk's part. Thanks for finding it. I'm curious why we need to pass around both pieces of info. Can't we just look for a `.cabal` file to find the actual package name?

alexfmpe:
I don't really know. Is it possible to have 2 `.cabal` in the same directory?

3noch:
Yes it is, although it's *extremely* rare. I don't think I've ever seen it, ever. What if `FilePath` were a path to a *cabal* file instead of a directory?

alexfmpe:
You mean `.cabal` file or `package.yaml` ?
I think that would work right now, because `backend/frontend/common` is the only use case, but it means attempts to un-hardcode this (e.g. #489, #520) would need to specify the package file instead of the package folder in .nix, correct?

3noch:
I guess we could do our best to guess and only fail if it's ambiguous which would basically never happen...

xplat:
i second the idea of guessing, the usual tools do it ... in the long term i'd prefer if the entire hardcoding thing went away (maybe referring to cabal.project instead), as well as the mechanism changing.

3noch:
@alexfmpe How did you run into this exactly? Did you have `backend/some-other-name.cabal` instead of `backend/backend.cabal`, for example?

3noch:
Now to add tests

alexfmpe:
> @alexfmpe How did you run into this exactly? Did you have `backend/some-other-name.cabal` instead of `backend/backend.cabal`, for example?

https://github.com/obsidiansystems/obelisk/pull/599#discussion_r361559329 (3)

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -105,38 +107,50 @@ runWatch = do
...
+  => FilePath -- ^ Directory to search for cabal package
+  -> m (Either GuessPackageFileError (Either FilePath FilePath))
+  -- ^ 'Right' 'Left' for hpack package, 'Right' 'Right' for cabal package; 'FilePath' is relative to given directory.
+guessCabalPackageFile dir = do
+  candidates <- liftIO $
+        filterM doesFileExist
+    =<< filter (liftA2 (||) (== Hpack.packageConfig) (".cabal" `L.isSuffixOf`))
+    <$> listDirectory dir
+  pure $ case L.partition (== Hpack.packageConfig) candidates of
+    ([hpack], _) -> Right $ Left hpack
```
alexfmpe: This also catches `([hpack], [cabal])`. I'm not too familiar with `hpack`, I guess we can assume that to be the generated cabal from hpack *if* the name matches?
I don't think this should catch `([hpack], [cabal1,cabal2])` at any rate.

lib/command/src/Obelisk/Command/Run.hs
srid: @alexfmpe Just curious: what's the use case for doing (3)? To have `ob run` reload when obelisk-route changes?

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -105,38 +107,50 @@ runWatch = do
...
-        liftIO (readPackageConfig decodeOptions) >>= \case
-          Left err -> do
-            putLog Error $ T.pack $ "Failed to parse " <> hpackFp <> ": " <> err
-            return Nothing
-          Right (DecodeResult hpackPackage _ _ _) -> do
-            return $ Just $ renderPackage [] hpackPackage
-      else do
-        putLog Error $ T.pack "Found neither cabal nor hpack file"
-        return Nothing
+parseCabalPackage dir = do
```
alexfmpe: Maybe it would be better to have the bulk of this function run inside `ExceptT (NonEmpty Text)`, ending with

```haskell 
\case
  Left errs -> for errs (putLog Error) *> pure Nothing
  Right info -> pure info
```
I feel that any failure here should provide a user-facing reason, rather than be allowed to simply return `Nothing` without logs of any kind (e.g. #597).

lib/command/src/Obelisk/Command/Run.hs
alexfmpe: @srid Yeah I was trying to get that in the same ghci(d) as the skeleton to avoid slow compilation cycles (several obelisk packages depend on `obelisk-route`. Didn't quite get to something that works correctly enough always because `obelisk-run` is one of those dependents and `ob run` uses it, so I ran into several weird things on my repl.

lib/command/src/Obelisk/Command/Run.hs
3noch: I'm going with "If you have a package.yaml file, I don't care what your cabal situation is because it should be generated anyway"

lib/command/src/Obelisk/Command/Run.hs
3noch: At this point the function does provide a user facing reason in any place that it returns `Nothing`.

3noch: I see your point though.

3noch: Done.

default.nix
```diff
@@ -47,6 +47,8 @@ let
         pkgs = self.callPackage ({ pkgs }: pkgs) {};
         onLinux = pkg: f: if pkgs.stdenv.isLinux then f pkg else pkg;
       in {
+        shelly = self.callHackage "shelly" "1.9.0" {};
```
alexfmpe: Might want to upstream this: https://github.com/reflex-frp/reflex-platform/issues/542

3noch:
Dropped the changelog entry since this doesn't *actually* let someone use a different cabal package name. It only means that `ob run` will support it. :man_facepalming:

PR #603, opened 2020-01-02
Use file-cloned local obelisk impl in SelfTest

414owen:
This makes all tests green, but might be a better solution than the temp directory...
@3noch


3noch:
Great! Thank you!

PR #606, opened 2020-01-03
Make obelisk depend on reflex-dom-core instead of reflex-dom

matthewbauer:
All of the Obelisk libraries only need reflex-dom-core to work.
reflex-dom reexports reflex-dom-core, so this worked fine before, but
it had the side effect of pulling in jsaddle-webkit2gtk. For apps that
are just using jsaddle-warp, this pulls in a lot of GTK dependencies
that are never used. Switching everything to reflex-dom-core allows
individual apps avoid these extra dependencies.


3noch:
@matthewbauer Do you have any idea what the CI failures are about? If not I'll take a deeper look.

3noch:
Resolved merge conflicts.

3noch:
@matthewbauer Still having issues with CI. The errors are strange though, complaining about zlib?

Ericson2314:
Android zlib errors I believe are fixed in newer reflex platform.

matthewbauer:
I don't think it's been resolved - some weird error that only occurs when we don't pull in the extra dependencies from reflex-dom.

3noch:
#779 Should have solved the hard part of this.

PR #607, opened 2020-01-06
autodetect private repos

cardenaso11:
This should make it so the private repos added in https://github.com/obsidiansystems/obelisk/pull/594 can be automatically detected by obelisk, by connecting and attempting to ls-remote. Currently it attempts to access ssh repos as https.


3noch:
Need to see how this plays with `--private/--public` CLI flags still.

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -851,9 +853,9 @@ getLatestRev os = do
 uriThunkPtr :: MonadObelisk m => GitUri -> Maybe Text -> Maybe Text -> m ThunkPtr
 uriThunkPtr uri mbranch mcommit = do
   commit <- case mcommit of
-    Nothing -> gitGetCommitBranch uri mbranch >>= return . snd
+    Nothing -> gitGetCommitBranch uri mbranch >>= return . snd -- return . (\(_, y, _)->y) --perhaps do io here instead of making uriToTHunkSource effectful ?
```
3noch: Let's get rid of this comment.

3noch:
We need to make sure that passing `--public/--private` not only overrides any setting found by auto-detection, but actually skips auto-detection completely. I can imagine edge cases where auto-detection gets in the way / gets stuck on some connection that isn't working and having a way to override+skip will be a valuable fallback. Not to mention it's the only sensible UX for those flags.

PR #608, opened 2020-01-06
Fix selftest parallelization issues

414owen:
Use a copy of the obelisk impl directory, so we can run stateful tests in parallel deterministically again.


3noch:
Wonderful! Thank you!

PR #610, opened 2020-01-07
Link `cp` using `staticWhich` so selftests runs on OSX

ingenieroariel:


3noch:
Wonderful this is exactly the fix I wanted.

PR #612, opened 2020-01-08
Allow overriding reflex-platform source

kmicklas:


3noch:
Thanks!

Ericson2314:
I would rename `reflex-platform` to `reflex-platform-func` and `reflex-platform-host` back to `reflex-platform`. That way it's consistent with Nixpkgs.

PR #616, opened 2020-01-13
Add prerender and websockets to skeleton project

414owen:
closes https://github.com/obsidiansystems/obelisk/issues/320


skeleton/frontend/src/Frontend.hs
```diff
@@ -1,22 +1,55 @@
...
+import Data.FileEmbed
+import Language.Javascript.JSaddle (MonadJSM) 
 
 import Common.Api
 import Common.Route
+import Obelisk.Generated.Static
+
+appRoute :: T.Text
+appRoute =
+  T.decodeUtf8 $(embedFile "config/common/route")
```
ryantrinkle: This should use the existing config infrastructure instead

3noch:
Let's *just* do a minimal prerender example and leave websockets out of it.

Issue #241, opened 2018-10-13
ObeliskWidget is missing MonadJSM constraint


JBetz:
I need to use `GHCJS.DOM.Types.MonadJSM`'s `currentWindow` method in order to react to window events, but this constraint currently isn't provided by the `ObeliskWidget` that the frontend widgets are forced to run under. And adding it resulted in backend build errors, presumably because it relies on a `GHCJS` module.

Is there another way to get window events? And if not, can the `MonadJSM` constraint somehow be enabled for the frontend without interfering with the backend?

ryantrinkle:
Obelisk now pre-renders the page on the server side, so to create a widget that relies on JS, you can use [`prerender`](https://github.com/reflex-frp/reflex-dom/blob/8bde2d72193369c92a995820ca1f56690d3c8ec3/reflex-dom-core/src/Reflex/Dom/Prerender.hs#L47).  You supply it two widgets, and the first will be used for server-side rendering, while the second will be used for client-side rendering (and will have access to JS).

ryantrinkle:
Leaving this open so I remember to add documentation for this.

PR #617, opened 2020-01-13
Add prerender to skeleton

414owen:
Closes https://github.com/obsidiansystems/obelisk/issues/320
Works on https://github.com/obsidiansystems/obelisk/issues/241

### DIFF
--- a/skeleton/frontend/frontend.cabal
+++ b/skeleton/frontend/frontend.cabal
@@ -9,6 +9,7 @@ library
                , common
                , obelisk-frontend
                , obelisk-route
+               , jsaddle
                , reflex-dom
                , obelisk-executable-config-lookup
                , obelisk-generated-static
--- a/skeleton/frontend/src/Frontend.hs
+++ b/skeleton/frontend/src/Frontend.hs
@@ -4,8 +4,10 @@
 
 module Frontend where
 
+import Control.Monad
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
+import Language.Javascript.JSaddle (eval, liftJSM)
 
 import Obelisk.Frontend
 import Obelisk.Configs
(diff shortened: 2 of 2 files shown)


skeleton/frontend/src/Frontend.hs
```diff
@@ -4,8 +4,11 @@
 
 module Frontend where
 
+import Control.Monad
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
+import GHCJS.DOM.Types (liftJSM)
```
3noch: I expect these will require dependencies to `frontend.cabal`. CI should confirm :smile:

skeleton/frontend/src/Frontend.hs
```diff
@@ -26,6 +29,7 @@ frontend = Frontend
   , _frontend_body = do
       el "h1" $ text "Welcome to Obelisk!"
       el "p" $ text $ T.pack commonStuff
+      prerender_ blank $ liftJSM $ void $ eval ("console.log('Hello, World!')" :: T.Text)
```
3noch: Let's add a comment explaining that this is an *example* of using prerender specifically. And maybe alert really would be a better example since `print` and `putStrLn` can technically also be used for logging and we don't want to imply that this is *the* way to log.

skeleton/frontend/frontend.cabal
```diff
@@ -24,6 +26,8 @@ executable frontend
                , common
                , obelisk-frontend
                , obelisk-route
+               , ghcjs-dom-jsffi
+               , jsaddle
```
3noch: Shouldn't need them in the executable. In fact ... we probably can prune these even further...

skeleton/frontend/frontend.cabal
```diff
@@ -9,6 +9,8 @@ library
                , common
                , obelisk-frontend
                , obelisk-route
+               , ghcjs-dom-jsffi
```
3noch: Can you try `ghcjs-dom` instead?

skeleton/frontend/src/Frontend.hs
```diff
@@ -30,9 +30,9 @@ frontend = Frontend
       el "h1" $ text "Welcome to Obelisk!"
       el "p" $ text $ T.pack commonStuff
       
-      -- 'prerender' and 'prerender_' let you choose a widget to run on the server
+      -- `prerender` and `prerender_` let you choose a widget to run on the server
       -- during prerendering and a different widget to run on the client with
-      -- JavaScript. The following will generate a 'blank' widget on the server and
+      -- JavaScript. The following will generate a `blank` widget on the server and
```
3noch: Using `'` is the "Haddock" way to refer to a function in scope. This doesn't get built by haddock though... still I kinda prefer the haddock-friendliness. Just FYI really.

PR #625, opened 2020-01-23
Remove COMPLETE pragma on tuple due to GHC bug

3noch:
See https://gitlab.haskell.org/ghc/ghc/issues/17729

cc @tomsmalley who found the bug in GHC!

I have:


3noch:
Oh this actually causes incomplete pattern match warnings on *all* pattern matches on 2-tuples. That's unacceptable. We'll have to go a different route.

3noch:
No I'm wrong. It only causes incomplete pattern match warnings on uses of the pattern synonym. But at this point the pattern synonym isn't useful.

PR #632, opened 2020-01-29
Fix override of Android static assets in ob deploy

kmicklas:
`staticSrc` is what the iOS builder calls it but this gets silently ignored for Android.

[Provide a clear overview of your changes.]

I have:


3noch:
Thanks! What's the best way to test this change and continue checking for regressions?

kmicklas:
No idea how to test it. I tried the nix fragment in my project and had to make this change for it to work. The best way to prevent this would probably be if the interface exposed in `passthru` were simpler (e.g. just take the config path and return the app derivation).

3noch:
How did you observe the bug to begin with? Was something clearly broken in your app?

kmicklas:
I copied out this nix fragment and used it to try to build an APK with a different set of config files than the default ones used for `ob run`. The override got silently ignored and the default config files were used. Replacing `staticSrc` with `assets` it works as intended. I haven't empirically verified that the same issue would apply here in `ob deploy` but it seems pretty clear looking at `__androidWithConfig` in `default.nix`.

PR #641, opened 2020-02-04
Don't hardcode 'Frontend' module

alexfmpe:
[Provide a clear overview of your changes.]

I have:

  - [x] Based work on latest `qa` branch


3noch:
Despite this being partly my idea: I think something *like* this may be where we end up, but exporting `frontend` from `Backend` is a bit too odd as is.

Closing until we make it better.

PR #654, opened 2020-02-11
Add obelisk profiling feature

matthewbauer:
Adds obelisk profile command. This dumps some profiling information into the profile/ directory that can give you some idea of what your program is spending resources on. Currently outputs 3 profiles:

- Heap profiling (.hp)
- Time profiling (.prof)
- Reflex profiling (.rprof)

I have:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -399,7 +399,7 @@ in rec {
       in nixpkgs.runCommand "ob-run" {
            buildInputs = [ (profiled.ghc.ghcWithPackages (p: [ p.backend p.frontend])) ];
       } ''
-        mkdir -p $out/bin/ob-run
+        mkdir -p $out/bin/
         ghc -x hs -prof -fno-prof-auto -threaded ${exeSource} -o $out/bin/ob-run
       '';
 
--- a/default.nix
+++ b/default.nix
@@ -17,10 +17,8 @@ let
     openssh
   ];
 
-  getReflexPlatform = sys: reflex-platform-func {
-    inherit iosSdkVersion config;
-    system = sys;
-    enableLibraryProfiling = profiling;
+  getReflexPlatform = { system, enableLibraryProfiling ? profiling }: reflex-platform-func {
+    inherit iosSdkVersion config system enableLibraryProfiling;
 
     nixpkgsOverlays = [
       (self: super: {
@@ -102,7 +100,7 @@ let
     ];
   };
 
-  reflex-platform = getReflexPlatform system;
+  reflex-platform = getReflexPlatform { inherit system; };
   inherit (reflex-platform) hackGet nixpkgs;
   pkgs = nixpkgs;
 
@@ -261,7 +259,7 @@ in rec {
   # An Obelisk project is a reflex-platform project with a predefined layout and role for each component
   project = base': projectDefinition:
     let
-      projectOut = sys: let reflexPlatformProject = (getReflexPlatform sys).project; in reflexPlatformProject (args@{ nixpkgs, ... }:
+      projectOut = { system, enableLibraryProfiling ? profiling }: let reflexPlatformProject = (getReflexPlatform { inherit system enableLibraryProfiling; }).project; in reflexPlatformProject (args@{ nixpkgs, ... }:
         let
           inherit (lib.strings) hasPrefix;
           mkProject =
@@ -366,7 +364,7 @@ in rec {
               });
             in allConfig;
         in (mkProject (projectDefinition args)).projectConfig);
-      mainProjectOut = projectOut system;
+      mainProjectOut = projectOut { inherit system; };
       serverOn = projectInst: version: serverExe
         projectInst.ghc.backend
         mainProjectOut.ghcjs.frontend
@@ -376,6 +374,7 @@ in rec {
       linuxExe = serverOn (projectOut "x86_64-linux");
       dummyVersion = "Version number is only available for deployments";
     in mainProjectOut // {
+      profiled = projectOut { inherit system; enableLibraryProfiling = true; };
       linuxExeConfigurable = linuxExe;
       linuxExe = linuxExe dummyVersion;
       exe = serverOn mainProjectOut dummyVersion;
--- a/default.nix
+++ b/default.nix
@@ -375,7 +375,7 @@ in rec {
       dummyVersion = "Version number is only available for deployments";
     in mainProjectOut // {
 
-      profiledObRun = let
+      __unstable__.profiledObRun = let
         profiled = projectOut { inherit system; enableLibraryProfiling = true; };
         exeSource = builtins.toFile "ob-run.hs" ''
           module Main where
--- a/default.nix
+++ b/default.nix
@@ -398,7 +398,10 @@ in rec {
         '';
       in nixpkgs.runCommand "ob-run" {
            buildInputs = [ (profiled.ghc.ghcWithPackages (p: [ p.backend p.frontend])) ];
-      } "ghc -x hs -prof -fno-prof-auto -threaded ${exeSource} -o $out";
+      } ''
+        mkdir -p $out/bin/ob-run
+        ghc -x hs -prof -fno-prof-auto -threaded ${exeSource} -o $out/bin/ob-run
+      '';
 
       linuxExeConfigurable = linuxExe;
       linuxExe = linuxExe dummyVersion;
--- a/default.nix
+++ b/default.nix
@@ -374,7 +374,32 @@ in rec {
       linuxExe = serverOn (projectOut "x86_64-linux");
       dummyVersion = "Version number is only available for deployments";
     in mainProjectOut // {
-      profiled = projectOut { inherit system; enableLibraryProfiling = true; };
+
+      profiledObRun = let
+        profiled = projectOut { inherit system; enableLibraryProfiling = true; };
+        exeSource = builtins.toFile "ob-run.hs" ''
+          module Main where
+
+          import Control.Exception
+          import Reflex.Profiled
+          import System.Environment
+
+          import qualified Obelisk.Run
+          import qualified Frontend
+          import qualified Backend
+
+          main :: IO ()
+          main = do
+            args <- getArgs
+            let port = read $ args !! 0
+                assets = args !! 1
+                profileFile = (args !! 2) <> ".rprof"
+            Obelisk.Run.run port (Obelisk.Run.runServeAsset assets) Backend.backend Frontend.frontend `finally` writeProfilingData profileFile
+        '';
+      in nixpkgs.runCommand "ob-run" {
+           buildInputs = [ (profiled.ghc.ghcWithPackages (p: [ p.backend p.frontend])) ];
+      } "ghc -x hs -prof -fno-prof-auto -threaded ${exeSource} -o $out";
+
       linuxExeConfigurable = linuxExe;
       linuxExe = linuxExe dummyVersion;
       exe = serverOn mainProjectOut dummyVersion;
--- a/dep/reflex-platform/github.json
+++ b/dep/reflex-platform/github.json
@@ -1,8 +1,8 @@
 {
   "owner": "reflex-frp",
   "repo": "reflex-platform",
-  "branch": "master",
+  "branch": "bump-reflex-0-6-4-1",
   "private": false,
-  "rev": "5429278830e1555a577f2550e045ce7f7164aa65",
-  "sha256": "1lp86cgccmim573rarsjny5vh0ygkfp5afq7006li0k9w2sw2d4c"
+  "rev": "38060894cb377160d1217305334f7f11202042d4",
+  "sha256": "0n47fcwj9szy3aqip0smmxdsp9fr78985rdslpmx2zdlkswmm5r1"
 }
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -53,8 +53,8 @@ newtype ArgsConfig = ArgsConfig
   { _argsConfig_enableVmBuilderByDefault :: Bool
   }
 
-args :: ArgsConfig -> UTCTime -> Maybe FilePath -> Parser Args
-args cfg time root = Args <$> noHandoff <*> verbose <*> obCommand cfg time root
+args :: ArgsConfig -> Parser Args
+args cfg = Args <$> noHandoff <*> verbose <*> obCommand cfg
 
 noHandoff :: Parser Bool
 noHandoff = flag False True $ mconcat
@@ -70,8 +70,8 @@ verbose = flag False True $ mconcat
   , help "Be more verbose"
   ]
 
-argsInfo :: ArgsConfig -> UTCTime -> Maybe FilePath -> ParserInfo Args
-argsInfo cfg time root = info (args cfg time root <**> helper) $ mconcat
+argsInfo :: ArgsConfig -> ParserInfo Args
+argsInfo cfg = info (args cfg <**> helper) $ mconcat
   [ fullDesc
   , progDesc "Manage Obelisk projects"
   ]
@@ -90,7 +90,7 @@ data ObCommand
    = ObCommand_Init InitSource Bool
    | ObCommand_Deploy DeployCommand
    | ObCommand_Run
-   | ObCommand_Profile FilePath
+   | ObCommand_Profile (Maybe FilePath)
    | ObCommand_Thunk ThunkCommand
    | ObCommand_Repl
    | ObCommand_Watch
@@ -106,13 +106,13 @@ data ObInternal
    = ObInternal_ApplyPackages String String String [String]
    deriving Show
 
-obCommand :: ArgsConfig -> UTCTime -> Maybe FilePath -> Parser ObCommand
-obCommand cfg time root = hsubparser
+obCommand :: ArgsConfig -> Parser ObCommand
+obCommand cfg = hsubparser
   (mconcat
     [ command "init" $ info (ObCommand_Init <$> initSource <*> initForce) $ progDesc "Initialize an Obelisk project"
     , command "deploy" $ info (ObCommand_Deploy <$> deployCommand cfg) $ progDesc "Prepare a deployment for an Obelisk project"
     , command "run" $ info (pure ObCommand_Run) $ progDesc "Run current project in development mode"
-    , command "profile" $ info (ObCommand_Profile <$> option auto (long "output" <> short 'o' <> help "Base output to use for profiling output. Suffixes are added to this based on the profiling type. Defaults to a timestamped path in the profile/ directory in the project's root." <> showDefault <> value ((fromMaybe "" root) </> "profile" </> formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S" time) <> metavar "PATH")) $ progDesc "Run current project with profiling enabled"
+    , command "profile" $ info (ObCommand_Profile <$> option auto (long "output" <> short 'o' <> help "Base output to use for profiling output. Suffixes are added to this based on the profiling type. Defaults to a timestamped path in the profile/ directory in the project's root." <> metavar "PATH")) $ progDesc "Run current project with profiling enabled"
     , command "thunk" $ info (ObCommand_Thunk <$> thunkCommand) $ progDesc "Manipulate thunk directories"
     , command "repl" $ info (pure ObCommand_Repl) $ progDesc "Open an interactive interpreter"
     , command "watch" $ info (pure ObCommand_Watch) $ progDesc "Watch current project for errors and warnings"
@@ -253,8 +253,8 @@ parserPrefs = defaultPrefs
   { prefShowHelpOnEmpty = True
   }
 
-parseCLIArgs :: ArgsConfig -> UTCTime -> Maybe FilePath -> [String] -> IO Args
-parseCLIArgs cfg time root = handleParseResult . execParserPure parserPrefs (argsInfo cfg time root)
+parseCLIArgs :: ArgsConfig -> [String] -> IO Args
+parseCLIArgs cfg = handleParseResult . execParserPure parserPrefs (argsInfo cfg)
 
 -- | Create an Obelisk config for the current process.
 mkObeliskConfig :: IO Obelisk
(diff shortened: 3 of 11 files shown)


lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -307,3 +308,26 @@ projectShell root isPure shellName command = do
...
   void $ liftIO $ waitForProcess ph
+
+findProjectAssets :: MonadObelisk m => FilePath -> m Text
+findProjectAssets root = do
+  let importableRoot = toNixPath root
+  isDerivation <- readProcessAndLogStderr Debug $
+    proc nixExePath
+      [ "eval"
+      , "-f"
+      , root
```
3noch: Might as well `setCwd` (in the Cli module) and just use constant `"."` here.

3noch: Actually, is there a reason to use `-f` *and* `import ...`?

matthewbauer: no, I think it's a no-op actually.

3noch: Note that I applied this improvement already to other places in #653

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -307,3 +308,24 @@ projectShell root isPure shellName command = do
...
     & nixShellConfig_common . nixCmdConfig_target . target_attr ?~ ("shells." <> shellName)
   void $ liftIO $ waitForProcess ph
+
+findProjectAssets :: MonadObelisk m => FilePath -> m Text
+findProjectAssets root = do
+  let importableRoot = toNixPath root
+  isDerivation <- readProcessAndLogStderr Debug $
+    proc nixExePath
+      [ "eval"
+      , "(let a = import " <> importableRoot <> " {}; in toString (a.reflex.nixpkgs.lib.isDerivation a.passthru.staticFilesImpure))"
```
3noch: Here you can still do `setCwd (Just root) $ proc nixExePath ...` and then `importableRoot` can just constant.

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -307,3 +308,24 @@ projectShell root isPure shellName command = do
...
+      -- `--raw` is not available with old nix-instantiate. It drops quotation
+      -- marks and trailing newline, so is very convenient for shelling out.
+      ]
+  -- Check whether the impure static files are a derivation (and so must be built)
+  if isDerivation == "1"
+    then fmap T.strip $ readProcessAndLogStderr Debug $ -- Strip whitespace here because nix-build has no --raw option
+      proc nixBuildExePath
+        [ "--no-out-link"
+        , "-E", "(import " <> importableRoot <> "{}).passthru.staticFilesImpure"
+        ]
```
3noch: This looks like duplicated code. Is there a way to share it?

3noch: I realize now this *is* sharing it. :D

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -69,40 +74,63 @@ data CabalPackageInfo = CabalPackageInfo
...
+              [ "module Main where"
+              , "import Control.Exception"
+              , "import Reflex.Profiled" ]
+              <> obRunImports <>
+              [ "main :: IO ()"
+              , "main = " <> obRunExpr <> " `finally` writeProfilingData \"" <> profileBaseName <> ".rprof\"" ]
+          -- Sane flags to enable by default, enable time profiling +
+          -- closure heap profiling.
+          rtsFlags = [ "+RTS", "-p", "-po" <> profileBaseName, "-hc", "-RTS" ]
+          profileDirectory = root </> "profile"
```
3noch: Seems like using `root` here is a bit odd. I'd expect the output to be in the directory where I ran the command.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -69,40 +74,63 @@ data CabalPackageInfo = CabalPackageInfo
...
+      putLog Debug "Using profiled build of project."
+      time <- liftIO $ formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S" <$> getCurrentTime
+      let exeSource =
+            unlines $
+              [ "module Main where"
+              , "import Control.Exception"
+              , "import Reflex.Profiled" ]
+              <> obRunImports <>
+              [ "main :: IO ()"
+              , "main = " <> obRunExpr <> " `finally` writeProfilingData \"" <> profileBaseName <> ".rprof\"" ]
```
3noch: Might as well play it safe and use `(" <> obRunExpr <> ")` in case that ever ends up using something like `$`.

lib/command/src/Obelisk/Command/Run.hs
matthewbauer: I would think you'd want it in a predictable place. That way you don't have to remember what directory you were in when you ran it.

By default, the profiling will put it in the cwd of where the executable was run. For things like config/ lookups to work correctly, this has to be the root directory. I guess if we want it to be where `ob profile` was run, we could save the cwd above and put it here.

3noch: I can very much imagine doing 
```
mkdir test-run1; cd test-run1; ob profile
mkdir test-run2; cd test-run2; ob profile
```

Possibly even in parallel (given the right circumstances).

3noch: Perhaps best of all would be to take an option like `-o` to actually let me specify instead. That can have a default.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -69,40 +74,63 @@ data CabalPackageInfo = CabalPackageInfo
...
+          -- Sane flags to enable by default, enable time profiling +
+          -- closure heap profiling.
+          rtsFlags = [ "+RTS", "-p", "-po" <> profileBaseName, "-hc", "-RTS" ]
+          profileDirectory = root </> "profile"
+          profileBaseName = profileDirectory </> time
+      liftIO $ createDirectoryIfMissing False profileDirectory
+      withSystemTempFile "ob-run-profiled.hs" $ \hsFname hsHandle -> withSystemTempFile "ob-run" $ \exeFname exeHandle -> do
+        liftIO $ hPutStr hsHandle exeSource
+        liftIO $ hFlush hsHandle
+        (_, _, _, ph1) <- createProcess_ "nixGhcWithProfiling" $ setCwd (Just root) $ proc "nix-shell" [ "-p", "((import ./. {}).profiled.ghc.ghcWithPackages (p: [ p.backend p.frontend]))", "--run", unwords [ "ghc", "-x", "hs", "-prof", "-fno-prof-auto", hsFname, "-o", exeFname ] ]
```
3noch: I wonder if we should actually build this with nix to
  1) allow caching
  2) allow us to actually pass the path to nix path to the user in case they want to copy a closure somewhere else and not depend on 'ob profile' in that environment

lib/command/src/Obelisk/Command/Run.hs
matthewbauer: Yeah a -o to configure profileBaseName sounds good. At least it makes this use case possible.

The idea with how this is set up is so to pick profileBaseName based on timestamps so multiple runs don't overwrite each other. Ideally this would have more information as well, so you would know what commit you were on, any special actions or notes, and maybe any additional changes from HEAD.

lib/command/src/Obelisk/Command/Run.hs
matthewbauer: Caching is hard here because freePort & assets will be different each time. I guess we can make both of those an option that is passed in. The linking can be a little slow, but since backend & frontend are already there it is usually comparable to the time it takes to setup ob run.

3noch: Yeah I definitely don't think we should be building one-off exes that embed port numbers, etc. Since we're running it ourselves anyway we might as well just pass that in...

lib/command/src/Obelisk/Command.hs
```diff
@@ -366,6 +368,7 @@ ob = \case
     DeployCommand_Update -> deployUpdate "."
     DeployCommand_Test (platform, extraArgs) -> deployMobile platform extraArgs
   ObCommand_Run -> run
+  ObCommand_Profile basePath -> profile basePath
```
3noch: Would be cleaner to resolve the default for `basePath` here instead of inlining it later.

lib/command/src/Obelisk/Command.hs
```diff
@@ -108,6 +109,7 @@ obCommand cfg = hsubparser
     [ command "init" $ info (ObCommand_Init <$> initSource <*> initForce) $ progDesc "Initialize an Obelisk project"
     , command "deploy" $ info (ObCommand_Deploy <$> deployCommand cfg) $ progDesc "Prepare a deployment for an Obelisk project"
     , command "run" $ info (pure ObCommand_Run) $ progDesc "Run current project in development mode"
+    , command "profile" $ info (ObCommand_Profile <$> optional (strOption ( long "output" <> short 'o' <> metavar "BASEPATH" ))) $ progDesc "Run current project with profiling enabled"
```
3noch: Add docs for what the default is when not specified

lib/command/src/Obelisk/Command.hs
matthewbauer: Unfortunately I don't think we know the root path yet. We also need a timestamp which isn't possible in the applicative.

lib/command/src/Obelisk/Command.hs
matthewbauer: yeah this would be fixed by the above, since optparse-applicative passes that.

lib/command/src/Obelisk/Command.hs
3noch: I mean just describe it in words

lib/command/src/Obelisk/Command.hs
3noch: I mean just describe it in words.

lib/command/src/Obelisk/Command.hs
3noch: See how `DeployCommand_Init` is handled as an example of how to do this.

matthewbauer: Done in 9ec5d2b6

matthewbauer:
> I'm still pretty suspicious of bothering to find root here. What's the problem if we just use CWD again?

It seems safer to use an absolute path here. I originally assumed it would save it to the executable's directory, not the CWD. That's not the case, but the documentation leaves that behavior unspecified (https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/profiling.html#rts-flag--po%20%E2%9F%A8stem%E2%9F%A9), and I'd prefer to not have to debug it if it ever changes.

lib/command/src/Obelisk/Command.hs
```diff
@@ -374,7 +371,14 @@ ob = \case
...
     DeployCommand_Test (platform, extraArgs) -> deployMobile platform extraArgs
   ObCommand_Run -> run
-  ObCommand_Profile basePath -> profile basePath
+  ObCommand_Profile mBasePath -> do
+    basePath <- case mBasePath of
+      Just path -> pure path
+      Nothing -> do
+        time <- liftIO $ getCurrentTime
+        root <- findProjectRoot "."
+        pure $ (fromMaybe "" root) </> "profile" </> formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S" time
```
3noch: You can use `withProjecRoot` here to protect against the case when the root isn't found.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -70,37 +74,67 @@ data CabalPackageInfo = CabalPackageInfo
...
-  withGhciScript pkgs root $ \dotGhciPath -> do
-    assets <- do
-      isDerivation <- readProcessAndLogStderr Debug $ setCwd (Just root) $
-        proc nixExePath
-          [ "eval"
-          , "(let a = import ./. {}; in toString (a.reflex.nixpkgs.lib.isDerivation a.passthru.staticFilesImpure))"
-          , "--raw"
-          -- `--raw` is not available with old nix-instantiate. It drops quotation
-          -- marks and trailing newline, so is very convenient for shelling out.
+  freePort <- getFreePort
```
3noch: I made a fix to `develop` for this that we should preserve here. The call to `getFreePort` is definitely a race condition waiting to happen so we need to *at least* run it *immediately* before binding on it.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -70,37 +74,67 @@ data CabalPackageInfo = CabalPackageInfo
...
+        <> obRunImports <>
+        [ "main :: IO ()"
+        , "main = getArgs >>= \\args -> (Obelisk.Run.run (read (args !! 0)) (Obelisk.Run.runServeAsset (args !! 1)) Backend.backend Frontend.frontend) `finally` writeProfilingData \"" <> profileBaseName <> ".rprof\""
+        , "  '';"
+        , "in obelisk.nixpkgs.runCommand \"ob-run\" {"
+        , "  buildInputs = [ (profiled.ghc.ghcWithPackages (p: [ p.backend p.frontend])) ];"
+        , "} \"ghc -x hs -prof -fno-prof-auto ${exeSource} -o $out\")" ]
+  exePath <- nixCmd $ NixCmd_Build $ def
+      Lens.& nixBuildConfig_outLink Lens..~ OutLink_None
+      Lens.& nixCmdConfig_target Lens..~ Target
```
3noch: Just import what you need here. We use `lens` a lot in this code base...

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -70,37 +74,67 @@ data CabalPackageInfo = CabalPackageInfo
...
+profile profileBaseName = withProjectRoot "." $ \root -> do
+  freePort <- getFreePort
+  assets <- findProjectAssets root
+  putLog Debug $ "Assets impurely loaded from: " <> assets
+  putLog Debug "Using profiled build of project."
+  liftIO $ createDirectoryIfMissing False $ takeDirectory profileBaseName
+  let -- Sane flags to enable by default, enable time profiling +
+      -- closure heap profiling.
+      rtsFlags = [ "+RTS", "-p", "-po" <> profileBaseName, "-hc", "-RTS" ]
+      nixBuildExpr = unlines $
```
3noch: We actually already depend on `here` in this project. You can use the `[iTrim|...|]` quasiquoter to make this much prettier.

3noch:
> > I'm still pretty suspicious of bothering to find root here. What's the problem if we just use CWD again?
> 
> It seems safer to use an absolute path here. I originally assumed it would save it to the executable's directory, not the CWD. That's not the case, but the documentation leaves that behavior unspecified (https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/profiling.html#rts-flag--po%20%E2%9F%A8stem%E2%9F%A9), and I'd prefer to not have to debug it if it ever changes.

I would have been very surprised if a built exe ran `getExecutablePath` in the RTS in order to figure out where to put files. The most obvious default is to use CWD. What's more, `root` isn't an absolute path and we can't make it one without opening up the can of worms that is relative path resolution. I.e. using `root` is *already* going to rely on CWD.

3noch:
Also `getExecutablePath` will very likely be write-protected because you typically don't want exes writing files into `/usr/bin` let alone `/nix/store`.

lib/command/src/Obelisk/Command/Run.hs
matthewbauer: thanks! that makes things much cleaner.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -89,21 +85,32 @@ profile profileBaseName = withProjectRoot "." $ \root -> do
...
+  exeSource = obelisk.nixpkgs.writeText "ob-run" ''
+module Main where
+
+import Control.Exception
+import Reflex.Profiled
+import System.Environment
+
+import qualified Obelisk.Run
+import qualified Frontend
+import qualified Backend
```
3noch: You can still interpolate `obRunImports` if you use `iTrim` instead as the QQ.

matthewbauer: oh okay

My thought on this was that it didn't make sense to share these since the `ob run` and `ob profile` main expressions are now substantially different. Ideally these would be closer so that we don't accidentally break ob profile when ob run changes or vice versa.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -70,30 +75,64 @@ data CabalPackageInfo = CabalPackageInfo
...
+      & nixBuildConfig_outLink .~ OutLink_None
+      & nixCmdConfig_target .~ Target
+        { _target_path = Nothing
+        , _target_attr = Nothing
+        , _target_expr = Just nixBuildExpr }
+  assets <- findProjectAssets root
+  putLog Debug $ "Assets impurely loaded from: " <> assets
+  freePort <- getFreePort
+  (_, _, _, ph) <- createProcess_ "runProfExe" $ setCwd (Just root) $ setDelegateCtlc True $ proc exePath ([show freePort, T.unpack assets, profileBaseName] <> rtsFlags)
+  _ <- liftIO $ waitForProcess ph
```
3noch: I think we have a wrapper around this that you should use instead.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -45,11 +48,13 @@ import System.Environment (getExecutablePath)
 import System.FilePath
 import qualified System.Info
 import System.IO.Temp (withSystemTempDirectory)
+import System.Process (waitForProcess)
```
3noch: We should never be importing `System.Process` directly since we wrap the interface. See previous comment for the fix.

lib/command/src/Obelisk/Command/Run.hs
matthewbauer: It looks like the wrappers (readProcessAndLogOutput & callProcessAndLogOutput) hide stdout from the user. The default ob run can provide some helpful information like what URL to visit. I guess we could do `readProcessAndLogOutput (Error, Error)` to show this, but that seems a little bit hacky. Using createProcess_ matches the behavior of `ob run` and `ob repl` with the nixShellWithPkgs function.

3noch: I mean just for `waitForProcess`

lib/command/src/Obelisk/Command.hs
```diff
@@ -111,7 +109,7 @@ obCommand cfg = hsubparser
...
     , command "deploy" $ info (ObCommand_Deploy <$> deployCommand cfg) $ progDesc "Prepare a deployment for an Obelisk project"
     , command "run" $ info (pure ObCommand_Run) $ progDesc "Run current project in development mode"
-    , command "profile" $ info (ObCommand_Profile <$> optional (strOption (long "output" <> short 'o' <> help "Base output to use for profiling output. Suffixes are added to this based on the profiling type. Defaults to a timestamped path in the profile/ directory in the project's root." <> metavar "PATH"))) $ progDesc "Run current project with profiling enabled"
+    , command "profile" $ info (ObCommand_Profile <$> (strOption (long "output" <> short 'o' <> help "Base output to use for profiling output. Suffixes are added to this based on the profiling type. Defaults to a timestamped path in the profile/ directory in the project's root." <> metavar "PATH" <> value "profile/%Y-%m-%dT%H:%M:%S" <> showDefault)) <*> (words <$> strOption (long "rts-flags" <> help "RTS Flags to pass to the executable." <> value "-p -hc" <> metavar "FLAGS" <> showDefault))) $ progDesc "Run current project with profiling enabled"
```
3noch: Let's not inline this massive thing now :D

dep/reflex-platform/github.json
```diff
@@ -1,8 +1,8 @@
 {
   "owner": "reflex-frp",
   "repo": "reflex-platform",
-  "branch": "master",
+  "branch": "bump-reflex-0-6-4-1",
```
3noch: Once this is merged we should bump and then we can merge here.

lib/command/src/Obelisk/Command/Run.hs
3noch: I see... I was wrong. We don't have a wrapper for just `waitForProcess`

matthewbauer: I don't see one for waitForProcess. There is something called withProcess which might be usable though? But it ends up supplying "CreatePipe" like readProcessAndLogOutput.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -71,30 +79,76 @@ data CabalPackageInfo = CabalPackageInfo
...
+    [ show freePort
+    , T.unpack assets
+    , profileBaseName
+    , "+RTS"
+    , "-po" <> profileBaseName
+    ] <> rtsFlags
+      <> [ "-RTS" ]
+  void $ waitForProcess ph
+  where
+    nixBuildExpr = [hereLit|
```
3noch: As discussed, this should be moved to `default.nix` so we instantiate it outside of `ob profile`.

default.nix
```diff
@@ -374,7 +374,32 @@ in rec {
       linuxExe = serverOn (projectOut "x86_64-linux");
       dummyVersion = "Version number is only available for deployments";
     in mainProjectOut // {
-      profiled = projectOut { inherit system; enableLibraryProfiling = true; };
+
+      profiledObRun = let
```
3noch: I'd feel better if this were in the `internal` attrset or similar since I don't really want to provide any guarantees about this API.

3noch: We need to make sure this is added to the cache outputs in `release.nix`.

matthewbauer: It is added to all-builds.nix for skeleton at least

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -71,30 +78,49 @@ data CabalPackageInfo = CabalPackageInfo
...
+  -> m ()
+profile profileBasePattern rtsFlags = withProjectRoot "." $ \root -> do
+  putLog Debug "Using profiled build of project."
+
+  time <- liftIO getCurrentTime
+  let profileBaseName = formatTime defaultTimeLocale profileBasePattern time
+
+  putLog Debug $ T.pack $ "Storing profiled data under base name of " <> profileBaseName
+
+  liftIO $ createDirectoryIfMissing False $ takeDirectory profileBaseName
```
3noch: This should be `True` to create parents too. That way patterns like `profile/%Y-%m-%dT%H:%M:%S/run` will work.

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -71,30 +78,49 @@ data CabalPackageInfo = CabalPackageInfo
...
+
+  outPath <- nixCmd $ NixCmd_Build $ def
```
3noch: Hm, `nixCmd` doesn't seem to properly `setCwd (Just root)` here so this command fails in anything but the root dir.

PR #662, opened 2020-02-21
Recognize loaderless thunk types in ob thunk

matthewbauer:
Some thunks like those in
reflex-platform/haskell-overlays/reflex-packages/dep/ don’t have a
default.nix loader. This is because the target repo doesn’t have a
default.nix, so import’ing doesn’t work. As a consequence, these
loaders only work with hackGet.

This makes ThunkType accept no loader, and not give an unrecognized
files error.

See https://github.com/reflex-frp/reflex-platform/tree/45a8885e2fe404060722cb5101297a30d308d88b/haskell-overlays/reflex-packages/dep/reflex

I have:


3noch:
I need to think about this. This seems like part of #223 which we haven't designed a solution for yet.

3noch:
We have a plan for how to fix this differently.

Issue #660, opened 2020-02-18
ghcide integration ?


JosephLucas:
Hi, 
Thank you for developing this great obelisk framework, I am found of the overall structure.
After creating a new project with `obelisk init` what can I do to start VScode with ghcide features ?
VSCode + ghcide works well for projects created with `stack new` but with obelisk projects it seems that imported modules cannot be found. I saw the [reflex-ghcide](https://github.com/mpickering/reflex-ghcide) tool but I cannot find out how to use it properly.
Cheers

-----
Edit : Author of `reflex-ghcide` explained that his project has nothing to do with obelisk projects and ghcide.

3noch:
Great idea! We'd love to have *some* sort of real IDE support. It's been difficult to justify investing heavily in any one idea since they tend to shift so much, but things seem to be on a good trajectory with ghcide and hie joining forces. Adding support for ghcide shouldn't be *that* hard. If you want to give it a stab I can provide pointers. Otherwise, we hope to revisit this idea but I can't say when.

JosephLucas:
Thanks. I will focus on learning and using the obelisk ecosystem first. If one day I get enough knowledge about it, I might develop tools for it :wink:

3noch:
Keep the ideas coming!

meditans:
@JosephLucas if you want to thinker around, take a look at https://github.com/mpickering/hie-bios/issues/54 for how I got to use ghcide with obelisk skeleton. Unfortunately ghcide can't follow definitions across packages, so the usefulness is less than I originally expected, and I got back to using `ob watch` till things mature a bit

JosephLucas:
> ghcide can't follow definitions across packages

[An issue is now open for this feature](https://github.com/digital-asset/ghcide/issues/448#issuecomment-589760022).

mpickering:
`reflex-ghcide` is nothing to do with obelisk projects and ghcide. @meditans has pointed out that there previously was support in `hie-bios` for obelisk projects but seeing as nobody was using it and maintaining it, I removed it.

JosephLucas:
Ooops. So IIUC reflex-ghcide uses reflex to run itself but is not meant to handle reflex projects.

maralorn:
You can use `ghcide` with obelisk but you have to compile ghcide with the patched ghc used by obelisk. (Thanks to @mpickering for pointing that out to me.)

Steps:

1a) Compile ghcide with e.g. `cabal new-install ghcide`

[...]

PR #663, opened 2020-03-04
Experimental ghcide support

3noch:
See #660 

To experiment:

```
cd skeleton
ob shell ghcide
```

For vscode, etc: `ob shell 'code .'`

Todo:

  - [x] Integrate nix to provide ghcide
  - [x] Add `ob shell --for-unpacked` to provide a nix-shell exactly like `ob run` uses
  - [x] Write wrapper script for `ghcide` that proxies to `ob shell --for-unpacked ghcide` so editors don't need to run inside a special shell

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
   - [Running over HTTPS](#running-over-https)
   - [IDE Support](#ide-support)
     - [Terminal-based feedback](#terminal-based-feedback)
-    - [**Experimental** `ghcide` support](#experimental-ghcide-support)
+    - [Experimental `ghcide` support](#experimental-ghcide-support)
 - [Deploying](#deploying)
   - [Locally](#locally)
   - [Default EC2 Deployment](#default-ec2-deployment)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
 - [Developing an Obelisk project](#developing-an-obelisk-project)
   - [Adding Packages](#adding-packages)
   - [Adding Package Overrides](#adding-package-overrides)
-  - [Running over https](#running-over-https)
+  - [Running over HTTPS](#running-over-https)
 - [Deploying](#deploying)
   - [Locally](#locally)
   - [Default EC2 Deployment](#default-ec2-deployment)
@@ -169,9 +169,10 @@ project ./. ({ pkgs, ... }: {
 
 ### Running over HTTPS
 
-To run your app locally over https, update the protocol in `config/common/route` to `https`, and then use `ob run` as normal.
+To run your app locally over HTTPS, update the protocol in `config/common/route` to `https`, and then use `ob run` as normal.
 
-Since Obelisk generates a self-signed certificate for running https, the browser will issue a warning about using an invalid certificate. On Chrome, you can go to `chrome://flags/#allow-insecure-localhost` to enable invalid certificates for localhost.
+Since Obelisk generates a self-signed certificate for running HTTPS, the browser will issue a warning about using an invalid certificate. On Chrome, you can go to `chrome://flags/#allow-insecure-localhost` to enable invalid certificates for localhost.
+
 
 ## Deploying
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
   - [Adding Packages](#adding-packages)
   - [Adding Package Overrides](#adding-package-overrides)
   - [Running over HTTPS](#running-over-https)
+  - [IDE Support](#ide-support)
+    - [Terminal-based feedback](#terminal-based-feedback)
+    - [**Experimental** `ghcide` support](#experimental-ghcide-support)
 - [Deploying](#deploying)
   - [Locally](#locally)
   - [Default EC2 Deployment](#default-ec2-deployment)
@@ -173,6 +176,42 @@ To run your app locally over HTTPS, update the protocol in `config/common/route`
 
 Since Obelisk generates a self-signed certificate for running HTTPS, the browser will issue a warning about using an invalid certificate. On Chrome, you can go to `chrome://flags/#allow-insecure-localhost` to enable invalid certificates for localhost.
 
+### IDE Support
+
+#### Terminal-based feedback
+
+Obelisk officially supports terminal-based feedback (akin to [`ghcid`](https://github.com/ndmitchell/ghcid)) in `ob run` and `ob watch`.
+
+#### **Experimental** `ghcide` support
+
+> **NOTE:** `ghcide` support is strictly **experimental** and **not officially supported**. There is no guarantee that it will work for you or that this support will be maintained over time.
+
+To try out the **experimental** [`ghcide`](https://github.com/digital-asset/ghcide) support:
+
+  * Install a `ghcide` plugin for your editor.
+  * Add `__withGhcide = true;` to your Obelisk project's `default.nix` (inside the call to `project`). With this setting, `ob shell` will provide `ghcide` as well.
+  * Add `hie-bios.sh` to the root of your project:
+      ```shell
+      cat > hie-bios.sh <<'EOF'
+      #!/usr/bin/env bash
+      source "$HOME/.bash_profile"; # NOTE: Some editors need help finding 'ob' and this assumes that `ob` is made available on your `$PATH` in `.bash_profile`
+      ob internal export-ghci-configuration > "$HIE_BIOS_OUTPUT"
+      EOF
+      chmod +x hie-bios.sh
+      ```
+  * Add `hie.yaml` to the root of your project:
+      ```shell
+      echo 'cradle: { bios: { program: hie-bios.sh } }' > hie.yaml
+      ```
+  * Test the configuration with `ob shell ghcide`. Assuming your project compiles and works with `ghcide` otherwise, this should succeed.
+  * Start your editor in a way that it can access the appropriate `ghcide`:
+      * This can sometimes be achieved with `ob shell 'start my editor'`.
+      * If you can't easily start your editor in a way that it inherits your shell, you may need to customize how your `ghcide` plugin finds `ghcide` or create a custom `ghcide` replacement script that you install globally. For example, for Visual Studio Code on macOS adding a script called `ghcide` to your `PATH` like the one below will cause `ghcide` from each project's `ob shell` to be used:
+          ```shell
+          #!/usr/bin/env bash
+          source "$HOME/.bash_profile"  # NOTE: This assumes that `ob` is made available on your `$PATH` in `.bash_profile`
+          ob shell "ghcide $@"
+          ```
 
 ## Deploying
 
--- a/all-builds.nix
+++ b/all-builds.nix
@@ -53,19 +53,28 @@ let
         obelisk.haskellPackageSets.ghcjs
         obeliskPackagesCommon;
       command = obelisk.command;
-      skeleton = import ./skeleton { inherit obelisk; };
-      serverSkeletonExe = skeleton.exe;
+
+      withSkeletonOptions = skel: options: (skel.passthru.__unstable__.self.extend (self: super: {
+        userSettings = super.userSettings // options;
+      })).project;
+      rawSkeleton = import ./skeleton { inherit obelisk; };
+      skeleton = withSkeletonOptions rawSkeleton {
+        withHoogle = true;  # cache the Hoogle database for the skeleton
+        __withGhcide = true; # cache the ghcide build for the skeleton
+      };
+
+      serverSkeletonExe = rawSkeleton.exe;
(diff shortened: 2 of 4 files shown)


3noch:
cc @mpickering

mpickering:
@3noch This is similar to what `hie-wrapper` does. Perhaps worth coordinating with the `hls` guys as I am sure they will implement something similar.

maralorn:
I think putting the overrides to make ghcide work with the current nixpkgs in the skeleton is a bad idea. People will have obelisk projects with very fast bitrot.

3noch:
@maralorn Oh yes I definitely agree. I plan to move that into obelisk's nix, or possibly reflex-platform's even.

maralorn:
@3noch I will play around with the expression for `ghcide` to see if I can make it a bit less obnoxious.

maralorn:
The problem is that this will quite probably break on every nixpkgs bump in r-p. We would need a sensible workflow for that. I wouldn‘t want this to prevent r-p from keeping up-to-date.

3noch:
I'm inclined to keep it in obelisk at first to see how onerous it is to keep it up-to-date. We won't release any "official" ghcide support in order to keep the commitment low. But we can try to provide it as long as it's not a bear. One convenience is that we can override for ghcide only and not affect the package set used by obelisk apps.

3noch:
Another option would be to figure out how to get ghcide-nix to use reflex-platform's patched GHC. Then we could piggyback on the "stack to nix package set" magic they're doing and not worry about it ourselves.

maralorn:
> One convenience is that we can override for ghcide only and not affect the package set used by obelisk apps.

Yeah, took me a while to figure out, that I should do it that way.

maralorn:
> Another option would be to figure out how to get ghcide-nix to use reflex-platform's patched GHC. Then we could piggyback on the "stack to nix package set" magic they're doing and not worry about it ourselves.

I guess that should be easy with a minor tweak to [ghcide-nix](https://github.com/cachix/ghcide-nix/blob/e840090fca63758ca55a2f3b74f3e13d93365e97
/nix/default.nix#L22) to expose `mkHieCore` as an attribute. Then we can just call that with our own ghc. (If haskell.nix can work with a none haskell.nix ghc, which I am not sure about.)

@domenkozar Would you accept a PR in that direction? The direction being: Let the caller to ghcide-nix provide the ghc themselves. Would that make sense for our purpose?

maralorn:
@3noch 

```
 shellToolOverrides = ghc: _:
    let
      overrides = ghc: _: {
        haddock-library = pkgs.haskell.lib.dontCheck
          (ghc.callHackage "haddock-library" "1.8.0" { });
        haskell-lsp = ghc.callHackage "haskell-lsp" "0.19.0.0" { };
        haskell-lsp-types = ghc.callHackage "haskell-lsp-types" "0.19.0.0" { };
        regex-posix = ghc.callHackage "regex-posix" "0.96.0.0" { };
        test-framework = pkgs.haskell.lib.dontCheck
          (ghc.callHackage "test-framework" "0.8.2.0" { });
        regex-base = ghc.callHackage "regex-base" "0.94.0.0" { };
        regex-tdfa = ghc.callHackage "regex-tdfa" "1.3.1.0" { };
        shake =
          pkgs.haskell.lib.dontCheck (ghc.callHackage "shake" "0.18.4" { });
        hie-bios = pkgs.haskell.lib.dontCheck (ghc.callHackageDirect {
          pkg = "hie-bios";
          ver = "0.4.0";
          sha256 = "19lpg9ymd9656cy17vna8wr1hvzfal94gpm2d3xpnw1d5qr37z7x";
        } { });
        ghcide = pkgs.haskell.lib.dontCheck (ghc.callHackageDirect {
          pkg = "ghcide";
          ver = "0.1.0";
          sha256 = "0vwaaqb74dzsvx5xdfkzbi8zzvbd5w9l1wdhl3rhvi8ibnrchgfs";
        } { });
      };
      ghcidePkgSet = ghc.override { inherit overrides; };
    in { inherit (ghcidePkgSet) ghcide; };
```

That‘s the nicest I can come up with. I have manually verified that every dontCheck still present is necessary. :-/

3noch:
@maralorn I appreciate it! I had basically given up on this idea until you handed us that nix :D I couldn't figure out magical set of bumps to use to make ghcide build with our package set.

maralorn:
> @maralorn I appreciate it! I had basically given up on this idea until you handed us that nix :D I couldn't figure out magical set of bumps to use to make ghcide build with our package set.

What I did: I built ghcide with cabal. And thus I had a list of compatible package versions. Then I overrode every error with the package version I new which should work. (So basically the haskell.nix cabal-buildplan-thing by hand.) The really annoying thing is that the job is even easier for cabal because cabal does not run the tests of all dependencies by default. nix does. And tests break easily and have more dependencies …

domenkozar:
> @domenkozar Would you accept a PR in that direction? 

Absolutely, passing in your own ghc makes a lot of sense.

maralorn:
@3noch Then I suggest we do the ghcide-nix thing and include it in reflex-platform. Of course r-p needs to expose it to obelisk in a way, that obelisk can provide it‘s patched ghc to it.
It probably should be gated behind a Boolean in a way that the ghcide-nix repo is only pulled in, when it is actually used. I think the most comfortable for users would be if we then also cache the built ghcide on the reflex-frp cachix.

3noch:
@maralorn Agreed! It may take me a bit to polish this off so feel free to beat me to it. But I do plan to take your suggested approach.

3noch:
Note that the patched ghc comes from reflex-platform anyway, so obelisk won't have to do anything fancy once it can get its paws on a working ghcide.

maralorn:
Aaaand down the rabbit-hole we go:

[ghcide-nix](https://github.com/cachix/ghcide-nix) now exposes a `mkGhcide` function which takes an `{ ghc = ...; }` attrset. (It has a problem with default values right now, but this will be fixed once [this fixup PR into ghcide-nix](https://github.com/cachix/ghcide-nix/pull/42) is merged as well.)

But playing around with that and obelisk led to this:

> It is likely you used `haskell.compiler.X` instead of `haskell-nix.compiler.X` for ghcide-0.1.0

Searching for that error led me [to this PR witch introduced that error](https://cardanoupdates.com/commits/6efcabacaf91018f5e624139b919c0e998986ea6). So apparently what we want to do, is not supported.

Possible ways forward:

* Ditch the ghcide-nix approach.
* Start using a haskell.nix ghc in reflex-platform.
* Try faking that our ghc is a haskell.nix compatible compiler. (Faking seems trivial if you look at what they check, but I guess the check is there for a reason? :-/)

*insert rambling about haskell+nix here*

3noch:
Yikes. That's no fun. I would not have expected ghc to be so deeply integrated into haskell.nix, but alas. Even if we were able to simply apply our patch to their ghc, I'll bet the nix paths would be different enough that linking during TH would still fail.

cgibbard:
Maybe it'd be worth getting Hamish's input on what the actual constraints
on the GHC in use are -- he mentioned at least a Cabal library patch that
was needed. If they're not too onerous, perhaps we can make
reflex-platform's GHC satisfy them.

3noch:
Good idea. @hamishmack do you think we could apply necessary changes to reflex-platform's GHC to work with haskell.nix?

maralorn:
> I'll bet the nix paths would be different enough that linking during TH would still fail.

Just out of curiosity. Can you elaborate on that point? Of course using a different ghc will change all nix-store hashes. But why would that affect anything to do with Template Haskell?

3noch:
When I tried to use ghcide-nix directly it worked except for failing to link to dependencies needed during TH.

3noch:
@maralorn @JosephLucas Please refer to new README section about `ghcide` support and let me know what kind of experience you have.

JosephLucas:
- `hie-bios.sh` assumes a `~/.bash_profile` file exists.
-  Otherwise I am stuck with this error
```
Step 4/6, Cradle 1/1: Loading GHC Session
> Invalid argument `internal'
> 
> Usage: ob [-v|--verbose] COMMAND
>   Manage Obelisk projects
ghcide: CradleError (ExitFailure 1) ["Invalid argument `internal'","","Usage: ob [-v|--verbose] COMMAND","  Manage Obelisk projects"]
```
Still in `hie-bios.sh`, what means `ob internal`  ? in
```
ob internal export-ghci-configuration > "$HIE_BIOS_OUTPUT"
```

maralorn:
@3noch That looks fine to me. Although I don‘t use the hie-bios.sh setup method. I just have something like
```
❯ cat frontend/hie.yaml
cradle:
   cabal:
      component: "frontend"
```
in every subpackage. That works out of the box when in the obelisk nix-shell.

It feels easier to me, but it might also have disadvantages.

3noch:
@maralorn Interesting. I'll play around with that. I tried to make `ghcide` behave identically to `ob run`, but it's not quite right.

3noch:
@JosephLucas Yes that assumption is noted in the comments. The README isn't (yet) a full tutorial. More like "Something along these lines should work". Your project needs to be using this branch of obelisk. `ob thunk update --branch eac@ghcide .obelisk/impl` will update it but be warned that will also put you on an unstable version of obelisk.

mpickering:
@maralorn Using separate hie.yaml files won't work well with the multi-component branch.

mpickering:
@3noch If you have any comments about your experience setting up a custom cradle or any questions then please email me or message me on IRC so we can chat about it.

3noch:
@maralorn IIRC using multiple hie.yaml files actually requires that the *other* dependencies be fully built before you get feedback on the one you're editing. I.e. if I edit `common` and then open `frontend` it will need to fully build `common` before showing me anything in `frontend`.

3noch:
@mpickering Thanks very much. Now that I have a baseline working option (despite the preprocessor not working) I will try out your multi-component branch of ghcide.

maralorn:
@3noch That is true and a small annoyance. I restart my ghcide when the API of one subcomponent changes …

If your setup does not have that problem, I definitely have to test it.

tek:
tried it on a project of mine with `ob shell nvim` and it works partially – it sees obelisk types and local deps, but nothing from cabal files, like default-extensions or nonlocal deps.

3noch:
@tek I noticed the same thing. I'm trying to figure out if that's a problem with the setup or a problem with ghcide/hie-bios.

3noch:
@tek What do you mean by `non-local` deps? I was able to see type info for things not defined in my local packages (like things from `Prelude` or `reflex` etc).

tek:
> @tek What do you mean by "non-local" deps? I was able to see type info for things not defined in my local packages (like things from `Prelude` or `reflex` etc).

I get "unknown module" errors for dependencies that are not in the project directory. But after putting the `hie.yaml` into the subpackages, I external dependencies are visible, but local ones aren't :smile:

tek:
btw I have to do this:
```
ob internal export-ghci-configuration | grep -v Warning > "$HIE_BIOS_OUTPUT"
```
otherwise the warnings confuse ghcide

maralorn:
Do you have a cabal.project file which lists your different component packages?

tek:
@maralorn just found that out myself :smile:

tek:
everything works now, see my updated comment :tada:

tek:
ghcide with stack is able to resolve all components in subprojects without any help from `hie.yaml`, any idea if this is a stack feature or if it's related to how `ob` or nix set up the environment?

3noch:
stack is a built-in cradle for hie-core. Obelisk isn't. Maybe we'll get that far eventually.

tek:
ah.

3noch:
@tek @maralorn @mpickering I found a bug in our preprocessor (#709) that was preventing `ghcide` from working as well as `ob run`. There must be a difference between how `ob run` and `ghcide` pass paths to the preprocessor. But regardless, if you update this branch you should get `ghcide` support that as just as good as `ob run`. This means packages should be aware of their respective cabal configurations (like `default-extensions`) and all your local packages will be in the same session (no recompilation between libs).

tek:
I removed the subprojects' `hie.yaml`s and the `cabal.project`, and `ob shell ghcide` seems to work, and much faster! Is that a consequence of your fix or could this be some cached configuration?

3noch:
It should be much faster than having separate `hie.yaml` files because it loads all of your packages into ghcide at once (interpreted) instead of forcing `ghcide` to do full builds of your libraries whenever you change them and then switch to a different package.

tek:
wonderful!

README.md
```diff
@@ -10,7 +10,10 @@ Obelisk provides an easy way to develop and deploy your [Reflex](https://github.
 - [Developing an Obelisk project](#developing-an-obelisk-project)
   - [Adding Packages](#adding-packages)
   - [Adding Package Overrides](#adding-package-overrides)
-  - [Running over https](#running-over-https)
+  - [Running over HTTPS](#running-over-https)
+  - [IDE Support](#ide-support)
+    - [Terminal-based feedback](#terminal-based-feedback)
+    - [**Experimental** `ghcide` support](#experimental-ghcide-support)
```
tomsmalley, line 16: This shouldn't be bold, it makes it stand out more than other items

3noch: I'll fix. Thanks.

PR #666, opened 2020-03-10
Put gitignoreSource under cleanSource

tomsmalley:
When these two are the other way around, `ob run` (or any ghcid using
obelisk's repl) is much slower.

I have:


ryantrinkle:
I've heard that filtering things in the nix store doesn't work properly; could this be a symptom of that?  CC @Ericson2314

PR #668, opened 2020-03-12
Allow customisation of how all.js is loaded

tomsmalley:
If your page is mostly static, running the all.js blob immediately is not much of a concern.
When it does run, it will often cause the UI thread to stop for a significant chunk of time, meaning any animations which may be running at page load will be janky.
This change allows loading all.js at a later time, for example, after most of the animations have already finished.

I have:

### DIFF
--- a/lib/backend/src/Obelisk/Backend.hs
+++ b/lib/backend/src/Obelisk/Backend.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
@@ -12,9 +13,15 @@ module Obelisk.Backend
   -- * Running a backend
   , runBackend
   , runBackendWith
+  -- * Configuration of backend
+  , GhcjsWidgets
+  , GhcjsWidgetsImpl(..)
+  , defaultGhcjsWidgets
   -- * all.js script loading functions
   , deferredGhcjsScript
   , delayedGhcjsScript
+  -- * all.js preload functions
+  , preloadGhcjs
   -- * Re-exports
   , Default (def)
   , getPageName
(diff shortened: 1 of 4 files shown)


tomsmalley:
This will certainly conflict with #644, but should be easy to resolve

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -193,16 +208,37 @@ runBackend backend frontend = case checkEncoder $ _backend_routeEncoder backend
 renderGhcjsFrontend
   :: (MonadSnap m, HasCookies m)
   => (route -> Text)
-  -> Text -- ^ URL to GHCJS app JavaScript
+  -> Text
+  -> (Text -> FrontendWidgetT route ()) -- ^ GHCJS all.js script widget
```
3noch: Maybe it makes more sense to not pass both `allJsUrl` and this script everywhere. Why not "close over" the URL and pass just the script?

tomsmalley: We use the path in the preload widget, which isn't customisable here. I thought about introducing a type to wrap this up, but didn't come to anything that felt "right".

3noch: Fair.

tomsmalley: I wonder if instead we should allow customising both preload widget and script widget? That shifts the problem slightly while giving the user a little more control.

tomsmalley: The two widgets would fit fairly sensibly in a config record, I think.

3noch: I like it.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -58,19 +70,37 @@ data GhcjsApp route = GhcjsApp
   , _ghcjsApp_value :: !(Frontend route)
   }
 
+type GhcjsWidgets r = GhcjsWidgetsImpl (Text -> FrontendWidgetT r ())
+
+-- | Widgets used to load all.js on the frontend
+data GhcjsWidgetsImpl a = GhcjsWidgets
```
3noch: Hm, I'm inclined to suggest that we drop the type alias and just make the ctor consistent name.

PR #687, opened 2020-04-01
Add routeLinkAttr function and handle the target=_blank case

meditans:
We had no way of providing custom attributes for the link in `routeLink`, and adding `target=_blank` is sometimes useful if we want to open the link in another tab, like we do in the marketing site.

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -145,7 +145,7 @@ deployPush deployPath getNixBuilders = do
         }
       & nixBuildConfig_outLink .~ OutLink_None
       & nixCmdConfig_args .~ (
-        [ strArg "hostName" host
+        [ strArg "hostName" $ fmap (\c -> if c == '.' then '_' else c) host
         , strArg "adminEmail" adminEmail
         , strArg "routeHost" routeHost
         , strArg "version" version


lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -530,14 +550,17 @@ routeLinkImpl
...
-routeLinkImpl r w = do
+routeLinkImpl attrs r w = do
   enc <- askRouteToUrl
   let cfg = (def :: ElementConfig EventResult t (DomBuilderSpace m))
-        & elementConfig_eventSpec %~ addEventSpecFlags (Proxy :: Proxy (DomBuilderSpace m)) Click (\_ -> preventDefault)
-        & elementConfig_initialAttributes .~ "href" =: enc r
+          & elementConfig_initialAttributes .~ ("href" =: enc r <> attrs)
+          & (if Map.lookup "target" attrs == Just "_blank"
+             then id
+             else elementConfig_eventSpec %~ addEventSpecFlags (Proxy :: Proxy (DomBuilderSpace m)) Click (\_ -> preventDefault))
```
3noch, line 563: In the case of _blank we actually want to not even `setRoute` below either.

3noch: Actually we *also* don't want to `scrollToTop`

meditans: Thanks, done

lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -519,27 +521,55 @@ routeLink
...
+    -- setRoute.
+    targetBlank = Map.lookup "target" attrs == Just "_blank"
+    cfg = (def :: ElementConfig EventResult t (DomBuilderSpace m))
+        & elementConfig_initialAttributes .~ ("href" =: enc r <> attrs)
+        & (if targetBlank
+           then id
+           else elementConfig_eventSpec %~ addEventSpecFlags (Proxy :: Proxy (DomBuilderSpace m)) Click (\_ -> preventDefault))
   (e, a) <- element "a" cfg w
-  setRoute $ r <$ domEvent Click e
+  when (not targetBlank) $ setRoute $ r <$ domEvent Click e
```
3noch, line 572: Is there a way to factor out the `targetBlank` check so we only have to do it once? Perhaps `scrollToTop` could be moved into `routeLinkImpl`?

ali-abrar:
Is there ever a case when target blank would be used with a route-altering
link? Wouldn’t you just use a plain <a href> with the target attribute and
not worry about routing on the page you’re leaving?

3noch:
@ali-abrar I think the answer is "yes...but". We have a policy for `routeLink` in that "It must be indistinguishable from `<a href>`. And here it definitely has *surprising* behavior in that it just completely ignores the `target="_blank"` *but only after hydration*! So I think the only reasonable thing to do is make it *not* weird....even though that itself is weird for different reasons. But at least it's consistent.

3noch:
To put it differently: `routeLink` should not force you to think that hard about your links. It should "Just Work".

lib/route/src/Obelisk/Route/Frontend.hs
tomsmalley: `routeLinkImpl` was used for the no-scroll-to-top version too. That function seems to have gone though, so I'm fine with moving the scrollToTop call into here. Otherwise we can just return `targetBlank` so we don't duplicate the check - this function isn't exported.

tomsmalley: We also need to do this check for the dynamic attr version.

PR #689, opened 2020-04-03
wip: Wasm support (run on worker thread)

dfordivam:
Wasm support.
Currently full of hacks!
Works for the skeleton app after disabling threaded for frontend

I have:

### DIFF
--- a/lib/backend/src/Obelisk/Backend.hs
+++ b/lib/backend/src/Obelisk/Backend.hs
@@ -279,7 +279,31 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
 preloadGhcjs :: GhcjsWasmAssets -> FrontendWidgetT r ()
 preloadGhcjs (GhcjsWasmAssets allJsUrl mWasm) = case mWasm of
   Nothing -> elAttr "link" ("rel" =: "preload" <> "as" =: "script" <> "href" =: allJsUrl) blank
-  Just _ -> pure ()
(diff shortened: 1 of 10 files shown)


lib/backend/src/Obelisk/Backend.hs
```diff
@@ -277,29 +279,73 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+        , "                   , [jsaddleVals.jsaddleListener]);"
+        ]
+
   elAttr "script" ("type" =: "text/javascript") $ text $ mconcat $
-    [ "var jsaddleVals = jsaddleJsInit();"
-    , "const worker = new Worker(\"" <> workerRunnerJs <> "\");"
-    , "worker.postMessage({ url: \"" <> wasmUrl <> "\""
-    , "                   , jsaddleVals: jsaddleVals }"
-    , "                   , [jsaddleVals.jsaddleListener]);"
+    [ "if (typeof(SharedArrayBuffer) === 'undefined') {           "
```
3noch, line 329: You can use `Data.String.Here.Interpolated` to make this much less noisy.

3noch: Either that or just use multiline strings since you don't need to manually include the `\n` in JS because it's not whitespace dependent.

3noch: In fact, multiline strings would be an ideal way to make this readable but still achieve some form of "minimization" on the output.

dfordivam: Done

PR #692, opened 2020-04-08
Clean up the skeleton

3noch:
I have:


skeleton/.gitignore
```diff
@@ -1,9 +1,16 @@
-dist-newstyle
-result
-result-*
-result-android
-result-ios
-result-exe
 .attr-cache
+.cabal-sandbox
+.ghc.environment.*
```
kmicklas, line 3: Having `.ghc.environment.*` in `.gitignore` is liable to cause a lot of people a lot of trouble I think.

alexfmpe: I suspect it's fine for nix-based use as long as `gitIgnoreSource` is called. 
I do that on several single-package projects, but it might be hard to enforce when having several unpacked thunks?

3noch: @kmicklas Is your reasoning that `git status` won't complain about it being around? I did think of this and it concerns me as well....

3noch: @kmicklas Can you elaborate your concern here?

kmicklas: @3noch, Yeah, just that `git status` wouldn't show them. You may *think* you're just on master, but GHC has other ideas...

3noch: Fair enough. I think we can disable them in cabal.project too which should eliminate any need to ignore them.

3noch:
```suggestion
```

srid:
@3noch Why was `skeleton/shell.nix` removed?

PR #697, opened 2020-04-11
Implement a dockerTools.buildLayeredImage target for obelisk projects

fabianhjr:
This is a rebase (solving conflicts) and squash of #512

I have:
  - [X] (Optional) Run CI tests locally: `nix-build release.nix -A build.x86_64-linux --no-out-link` (or `x86_64-darwin` on macOS)

Changes from #512:
 - ChangeLog Next entry on merge
 - version became optional with default "latest" docker tag
 - moved docker build along deployment in readme
 - changed buildImage to buildLayeredImage for image efficiency/reusability.
 - use an optional internalPort argument

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Functional reactive web and mobile applications, with batteries included. Obelis
   - [Custom Non-EC2 Deployment](#custom-non-ec2-deployment)
     - [VirtualBox Deployment](#virtualbox-deployment)
   - [Locally](#locally)
+  - [Docker](#docker)
   - [From macOS](#from-macos)
   - [Deploying an updated version](#deploying-an-updated-version)
 - [Mobile](#mobile)
@@ -185,6 +186,60 @@ cp -r config test-app
 (cd test-app && ./backend)
 ```
 
+### Docker
+
+#### Variables
+
+Setting the following environment variables will be helpful for keeping build and run parameters consistent:
+
+```bash
+export NAME=app
+export VERSION=$(git rev-parse HEAD)
+export CONFIGDIR=/path/to/obelisk/config
+```
+
+#### Building
+
+To build a Docker image using a default configuration, run:
+
+```bash
+docker load < $(nix-build -A dockerImage --no-out-link --argstr name $NAME --argstr version $VERSION)
+```
+
+To build with a custom configuration, Obelisk's default configuration can be imported and overriden in a `release.nix` file added to your top-level directory:
+
+```nix
+{ }:
+let
+  project = import ./. {};
+  obelisk = import ./.obelisk/impl {
+    system = builtins.currentSystem;
+  };
+  mkBaseDockerImageConfig = { version }: obelisk.dockerImageConfig {
+    name = "app";
+    exe = project.linuxExe;
+    version = version;
+  };
+in {
+  customDockerImage = args@{ version ? "latest" }:
+    obelisk.nixpkgs.dockerTools.buildImage (mkBaseDockerImageConfig args // {
+      config.expose = 3000; # overriding default port 8000
+    });
+}
+```
+
+The custom image can then be built with: `nix-build release.nix -A customDockerImage --no-out-link --argstr version $VERSION`
+
+#### Running
+
+Using the default configuration, the image can then be run with:
+
+```bash
+docker run -p 127.0.0.1:8000:8000/tcp --volume "$CONFIGDIR:/var/run/backend/config" "${NAME}:${VERSION}"
+```
+
+And the application should be accessible at `http://localhost:8000`.
+
 ### From macOS
 
 Deploying from macOS requires some extra setup:
@@ -312,12 +367,11 @@ This should copy over and install the application on your device (if you see a  
 
 After having configured signing for your app, you may proceed to build a release version of the app. This is no different to how you build the non-release version, so consult the section [Android](#android) further above for exact instructions on building and deploying to your device.
 
-
 ## Frequently Asked Questions (FAQ)
 
 Refer to [FAQ](FAQ.md).
 
-
 ## Contributing
 
 Contributions and issue reports are encouraged and appreciated! Refer to the [Contributing](CONTRIBUTING.md) guide for information about getting started.
+
--- a/default.nix
+++ b/default.nix
@@ -173,6 +173,42 @@ in rec {
 
   inherit mkAssets;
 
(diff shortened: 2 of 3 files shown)


fabianhjr:
Pending comments from #512:
 - version is required - would it make sense to make it optional and default to latest ?
 - I would suggest avoiding runAsRoot if at all possible, because using it means constructing the Docker image with QEMU, which is slower than using extraCommands or something along those lines.

3noch:
Awesome! We'll try to get this in!

fabianhjr:
Hi @3noch, good to hear. Currently doing some testing with some of the changes made from the original merge request. :3

Though, it will take me some time to really get a feel about this. First time doing anything docker/k8s related. I can say with confidence that buildLayeredImage, nix and obelisk have been really interesting to learn about.

Anyways, here are the layers created for the project I am currently working on (https://gitlab.com/LibreCybernetics/LaTiendita/-/tree/1df40b5d8b984e55588b4ada1ec70db336cb40c5):
Screenshot from 2020-04-11 00-00-42

fabianhjr:
From initial tests buildImage created a single ~550 MB layer, buildLayeredImage created several images and a ~450 MB one.
Screenshot from 2020-04-11 00-07-00

fabianhjr:
Interestingly enough, seems like the compressed image is closer to 160MB.
Screenshot from 2020-04-11 00-09-45

fabianhjr:
And it is live :')

http://latiendita.coop

fabianhjr:
Hi @3noch, on some further testing it seems like nix is packaging obelisk projects alongside thunks (like obelisk/obvelisk-oauth) as a single package and that stops some reusability from happening in the layered docker image.

I tried reading some of the nix expressions but wouldn't know how to separate them into packages. :C

3noch:
Obelisk executables are statically linked. Perhaps that's what you're seeing? However, it's uncommon to layer multiple *nix* produced Docker images. It's more common, I think, to do your layering *in* nix and then produce a single Docker image at the end. Nix is better at sharing correctly than Docker.

fabianhjr:
> Nix is better at sharing correctly than Docker.

Yes, however this is for deployment from nix, Docker doesn't build the layers nix does as per documentation:

>  pkgs.dockerTools is a set of functions for creating and manipulating Docker images according to the Docker Image Specification v1.2.0 . Docker itself is not used to perform any of the operations done by these functions.

It would be really neat if for this small project the difference (the project itself) could get to about 50 MB or so. (Though I am not really sure how small or not the resulting single project layer can get)

3noch:
If the docker image size is the problem then what we probably need to do is figure out how to minimize our nix closure for obelisk projects. I've looked into this before and I suspect we can make the closure much smaller. That will produce smaller Docker images.

default.nix
```diff
@@ -167,6 +167,35 @@ in rec {
...
+      config = {
+        Env = [
+          ("PATH=" + builtins.concatStringsSep(":")(extraPaths ++ [
+            "/var/lib/backend" # put the obelisk project on the path.
+            "/bin" # put contents on path
+          ] ++ map (pkg: "${pkg}/bin") pkgs.stdenv.initialPath # put common tools in path so docker exec is useful
+          ))
+        ];
+        Expose = 8000;
+        Entrypoint = ["/var/lib/backend/backend"];
```
3noch, line 191: We shouldn't have hardcoded this. We'll probably have to move this nix into the `mkObelisk` module function so that one can get a docker image for obelisk without assuming default configuration.

fabianhjr:
I think the closure doesn't need to be smaller. Maybe dynamically building and having thunks be their own nixpkg would allow nix to split them into different docker layers. (However that is a future nice to have)

default.nix
fabianhjr: Will look into it

fabianhjr: If I understood this correctly then any changes on internalPort should propagete to the docker target. It should now not be hardcoded.

fabianhjr: As per the entrypoint, if I understand this correctly that location is being grabbed from nix and symlinked in line 172

default.nix
```diff
@@ -167,6 +167,35 @@ in rec {
...
 
+  dockerImageConfig = args@{exe, name, internalPort ? 8000, version ? "latest", created ? "now",extraContents ? [], extraPaths ? []}:
+    let
+      appDirSetupScript = nixpkgs.runCommand "appDirSetupScript.sh" {} ''
+        mkdir -p    $out/var/lib/backend
+        ln -sft $out/var/lib/backend '${exe}'/*
+        ${nixpkgs.findutils}/bin/find $out/var/lib/backend
+        '';
+    in {
+      name = name;
```
endgame, line 178: `inherit name created;`

default.nix
```diff
@@ -167,6 +167,35 @@ in rec {
...
+      tag = version;
+      created = created;
+      contents = [ nixpkgs.iana-etc nixpkgs.cacert appDirSetupScript ] ++ extraContents;
+      config = {
+        Env = [
+          ("PATH=" + builtins.concatStringsSep(":")(extraPaths ++ [
```
endgame, line 184: I think parens around `":"` not required

README.md
```diff
@@ -192,6 +193,60 @@ cp -r config test-app
...
+To build a Docker image using a default configuration, run:
+
+```bash
+docker load < $(nix-build -A dockerImage --no-out-link --argstr name $NAME --argstr version $VERSION)
+```
+
+To build with a custom configuration, Obelisk's default configuration can be imported and overriden in a `release.nix` file added to your top-level directory:
+
+```nix
+{ }:
```
endgame, line 219: A function from the empty set seems not useful. Is there something I'm missing with this?

fabianhjr: Don't think so, not that I could find from #512

default.nix
fabianhjr: This seems necesary, got the following error when testing:

```
error: syntax error, unexpected ':', at /home/fabian/Development/Open/obelisk/default.nix:191:48
```

default.nix
fabianhjr: Hi @3noch, do the most recent version address this concern?

augyg:
@fabianhjr is this still being worked on and wanted?

fabianhjr:
I haven't used obelisk in a while, it is something I wanted though. Feel free to rebase and reopen a merge request.

PR #700, opened 2020-04-16
Fix init unpack bug

tomsmalley:
* Handoff to the obelisk we are `init`ing to repack the impl thunk
* Add a test for this

I have:


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -93,6 +93,11 @@ data ThunkSource
    | ThunkSource_Git GitSource
    deriving (Show, Eq, Ord)
 
+thunkSourceToGitSource :: ThunkSource -> GitSource
+thunkSourceToGitSource = \case
+  ThunkSource_GitHub s -> forgetGithub False s
+  ThunkSource_Git s -> s
```
3noch, line 99: I feel like I've seen this before. Is it already written somewhere else?

tomsmalley: I found a few similar cases. Switched them to use this function

3noch:
Ah yes thank you.

PR #708, opened 2020-04-20
Wasm support

dfordivam:
Cleaned up (rebased) the work done for wasm support.

The feature is working fine now, though there are still nix related things to be fixed, especially in reflex-platform.

I have:


3noch:
@dfordivam  Does this obviate #702 ?

dfordivam:
> 
> 
> @dfordivam Does this obviate #702 ?

Yes, and #689 also

default.nix
```diff
@@ -167,13 +195,13 @@ in rec {
 
   inherit mkAssets;
 
-  serverExe = backend: frontend: assets: optimizationLevel: version:
+  serverExe = backend: frontendJs: assets: optimizationLevel: frontendWasm: version:
```
3noch, line 198: This should probably get turned into an attrset now

default.nix
```diff
@@ -289,6 +317,7 @@ in rec {
                       lib.filter (x: lib.hasAttr x self.combinedPackages) self.shells-ghcSavedSplices;
                     ghc = lib.filter (x: lib.hasAttr x self.combinedPackages) self.shells-ghc;
                     ghcjs = lib.filter (x: lib.hasAttr x self.combinedPackages) self.shells-ghcjs;
+                    wasm = lib.filter (x: lib.hasAttr x self.combinedPackages) self.shells-ghcjs;
```
3noch, line 320: `self.shells-ghcjs` should be renamed to `self.shells-frontend` or similar since it now applies to both ghcjs and wasm.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -73,8 +78,9 @@ data Backend backendRoute frontendRoute = Backend
 data BackendConfig frontendRoute = BackendConfig
   { _backendConfig_runSnap :: !(Snap () -> IO ()) -- ^ Function to run the snap server
   , _backendConfig_staticAssets :: !StaticAssets -- ^ Static assets
-  , _backendConfig_ghcjsWidgets :: !(GhcjsWidgets (Text -> FrontendWidgetT (R frontendRoute) ()))
-    -- ^ Given the URL of all.js, return the widgets which are responsible for
+  , _backendConfig_ghcjsAssets :: !StaticAssets -- ^ Frontend ghcjs (and wasm) assets
```
3noch, line 81: We should rename this but it can be in a different PR.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -92,10 +98,14 @@ data GhcjsWidgets a = GhcjsWidgets
   -- ^ A script widget, placed in the document body
   } deriving (Functor, Generic)
 
+data GhcjsWasmAssets = GhcjsWasmAssets
+  { _ghcjsWasmAssets_allJs :: Text
```
3noch, line 102: Let's document what this `Text` is.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -92,10 +98,14 @@ data GhcjsWidgets a = GhcjsWidgets
...
+  , _ghcjsWasmAssets_wasmPathRoot :: Maybe Text
```
3noch, line 103: Document what this is.

lib/backend/src/Obelisk/Backend.hs
3noch: I think it's a relative URL.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -233,21 +250,34 @@ runBackendWith
...
+                widgets = ($ GhcjsWasmAssets allJsUrl mWasmUrls) <$> ghcjsWidgets
+                frontendApp = GhcjsApp
+                  { _ghcjsApp_compiled = ghcjsAssets
+                  , _ghcjsApp_value = frontend
+                  }
+  where
+    checkWasmFile = do
+      let base = _staticAssets_processed ghcjsAssets
+          p = "frontend.wasm"
+      assetType :: (Either IOError ByteString) <- liftIO $ try $ BSC8.readFile $ base </> p </> "type"
```
3noch, line 279: Is there a way we can get this sort of check *directly* from the Assets library? It seems like internal logic that should be encapsulated there.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -233,21 +250,34 @@ runBackendWith
...
+                mWasmUrls = if enableWasm then Just (renderWasmPathRoot validFullEncoder) else Nothing
+                widgets = ($ GhcjsWasmAssets allJsUrl mWasmUrls) <$> ghcjsWidgets
+                frontendApp = GhcjsApp
+                  { _ghcjsApp_compiled = ghcjsAssets
+                  , _ghcjsApp_value = frontend
+                  }
+  where
+    checkWasmFile = do
+      let base = _staticAssets_processed ghcjsAssets
+          p = "frontend.wasm"
```
3noch, line 278: Can we find a way not to hardcode this?

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \  document.head.appendChild(link_tag);                  \
+      \};                                                      \
+      \if (typeof(WebAssembly) === 'undefined') {              \
+      \  add_preload_tag('" <> allJsUrl <> "', 'script');      \
+      \} else {                                                \
+      \  add_preload_tag('" <> wasmUrl <> "', 'script');       \
+      \  add_preload_tag('" <> jsaddleJs <> "', 'script');     \
+      \  add_preload_tag('" <> interfaceJs <> "', 'script');   \
+      \  add_preload_tag('" <> runnerJs <> "', 'script');      \
+      \}"
```
3noch, line 334: Can we get rid of the trailing space on these?

3noch: Since we're not including `\n` we might as well get rid of leading space too.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+scriptTag :: DomBuilder t m => Text -> m ()
+scriptTag t = elAttr "script" ("type" =: "text/javascript") $ text t
+
+wasmScripts :: Text -> Text -> (Text, Text, (Int -> Text))
+wasmScripts allJsUrl wasmRoot = (preloadScript, runJsScript, delayedJsScript)
+  where
+    jsaddleJs = wasmRoot <> "/jsaddle_core.js"
+    interfaceJs = wasmRoot <> "/jsaddle_mainthread_interface.js"
+    runnerJs = wasmRoot <> "/mainthread_runner.js"
+    wasmUrl = wasmRoot <> "/frontend.wasm"
```
3noch, line 317: Note: Think about how to put these names somewhere else.

3noch: Probably we just need to make this a record of names somewhere.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
-deferredGhcjsScript allJsUrl = elAttr "script" ("type" =: "text/javascript" <> "src" =: allJsUrl <> "defer" =: "defer") blank
+deferredGhcjsScript :: GhcjsWasmAssets -> FrontendWidgetT r ()
+deferredGhcjsScript (GhcjsWasmAssets allJsUrl mWasm) = case mWasm of
+  Nothing -> elAttr "script" ("type" =: "text/javascript" <> "src" =: allJsUrl <> "defer" =: "defer") blank
+  Just wasmAssets -> scriptTag $ view _2 (wasmScripts allJsUrl wasmAssets)
+
+scriptTag :: DomBuilder t m => Text -> m ()
+scriptTag t = elAttr "script" ("type" =: "text/javascript") $ text t
+
+wasmScripts :: Text -> Text -> (Text, Text, (Int -> Text))
```
3noch, line 311: Let's document all of this.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+wasmScripts allJsUrl wasmRoot = (preloadScript, runJsScript, delayedJsScript)
+  where
+    jsaddleJs = wasmRoot <> "/jsaddle_core.js"
+    interfaceJs = wasmRoot <> "/jsaddle_mainthread_interface.js"
+    runnerJs = wasmRoot <> "/mainthread_runner.js"
+    wasmUrl = wasmRoot <> "/frontend.wasm"
+
+    preloadScript =
```
3noch, line 319: Preload tags should probably be *optimistic* in the sense that we shouldn't do anything fancy to get them. They should just *always* be present and load as many things as might make the page go faster.

3noch: Hm, actually want to think about this more.

3noch: After discussion I take this back. This is ok.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      "add_preload_tag = function (docSrc, docType) {          \
+      \  var link_tag = document.createElement('link');        \
+      \  link_tag.rel = 'preload';                             \
+      \  link_tag.as = docType;                                \
```
3noch, line 323: Why not hardcode `script` since it's the same value every time we call it.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \  link_tag.href = docSrc;                               \
+      \  document.head.appendChild(link_tag);                  \
+      \};                                                      \
+      \if (typeof(WebAssembly) === 'undefined') {              \
+      \  add_preload_tag('" <> allJsUrl <> "', 'script');      \
+      \} else {                                                \
+      \  add_preload_tag('" <> wasmUrl <> "', 'script');       \
+      \  add_preload_tag('" <> jsaddleJs <> "', 'script');     \
+      \  add_preload_tag('" <> interfaceJs <> "', 'script');   \
+      \  add_preload_tag('" <> runnerJs <> "', 'script');      \
```
3noch, line 333: Splicing strings directly into JS code always makes me nervous. Is there a rudimentary escaping we can do. It just so happens that all of these strings are *also* parts of a URL which means the rules should be even *more* stringent. This is mostly a TODO for us to think about.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \  add_deferload_tag('" <> allJsUrl <> "');    \
+      \} else {                                      \
+      \  var wasmFile = '" <> wasmUrl <> "';         \
+      \  add_deferload_tag('" <> jsaddleJs <> "');   \
+      \  add_deferload_tag('" <> interfaceJs <> "'); \
+      \  add_deferload_tag('" <> runnerJs <> "');    \
+      \}"
+
+    delayedJsScript n =
+      "var wasmFile = '" <> wasmUrl <> "';            \
```
3noch, line 354: This appears to be unused?

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \  var tag = document.createElement('script'); \
+      \  tag.type = 'text/javascript';               \
+      \  tag.src = docSrc;                           \
+      \  tag.setAttribute = ('defer', 'defer');      \
+      \  document.body.appendChild(tag);             \
+      \};                                            \
+      \if (typeof(WebAssembly) === 'undefined') {    \
+      \  add_deferload_tag('" <> allJsUrl <> "');    \
+      \} else {                                      \
+      \  var wasmFile = '" <> wasmUrl <> "';         \
```
3noch, line 347: This appears to be unused?

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +293,100 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \  add_deferload_tag('" <> jsaddleJs <> "');   \
+      \  add_deferload_tag('" <> interfaceJs <> "'); \
+      \  add_deferload_tag('" <> runnerJs <> "');    \
```
3noch, line 350: Why does this one only load 3 files but the preload loads 4?

lib/backend/src/Obelisk/Backend.hs
dfordivam: added a comment of why we need this

lib/backend/src/Obelisk/Backend.hs
dfordivam: Added comment to explain this

default.nix
```diff
@@ -261,7 +292,7 @@ in rec {
 
                 shells-ghc = builtins.attrNames (self.predefinedPackages // self.shellPackages);
 
-                shells-ghcjs = [
+                shells-frontend = [
```
3noch, line 295: Ryan pointed out that it's possible to have a ghc frontend so my suggestion of `shells-frontend` isn't quite right. Maybe `shells-web`?

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -233,21 +268,35 @@ runBackendWith
...
+                widgets = ($ GhcjsWasmAssets allJsUrl mWasmAssets) <$> ghcjsWidgets
+                frontendApp = GhcjsApp
+                  { _ghcjsApp_compiled = ghcjsAssets
+                  , _ghcjsApp_value = frontend
+                  }
+  where
+    checkWasmFile = do
+      let base = _staticAssets_processed ghcjsAssets
+          p = T.unpack $ _wasmAssets_wasmFile defaultWasmAssets
+      isJust <$> getAssetPath base p
```
3noch, line 299: nit: Binding to name `p` doesn't communicate much so might as well inline that.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +312,111 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+        \add_deferload_tag(" <> allJsUrl <> ");\
+      \} else {\
+        \var wasmFile = " <> wasmUrl <> ";\
+        \add_deferload_tag(" <> jsaddleJs <> ");\
+        \add_deferload_tag(" <> interfaceJs <> ");\
+        \add_deferload_tag(" <> runnerJs <> ");\
+      \}"
+
+    delayedJsScript n =
+      "var wasmFile = " <> wasmUrl <> ";\
```
3noch, line 384: `var WASM_APP_URL` or something and even a comment explaining what it's for.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +312,111 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+      \}"
+
+    -- The 'wasmFile' variable is needed here because export API is not working
+    -- in the webabi ts code. The wasm file is fetched by the runnerJs.
+    runJsScript =
+      "add_deferload_tag = function (docSrc) {\
+        \var tag = document.createElement('script');\
+        \tag.type = 'text/javascript';\
+        \tag.src = docSrc;\
+        \tag.setAttribute = ('defer', 'defer');\
```
3noch, line 371: I fear for backwards compatibility on this syntax. Let's stick with the function call. 
```suggestion
        \tag.setAttribute('defer', 'defer');\
```

lib/backend/src/Obelisk/Backend.hs
3noch: There's enough "logic" in these drivers that it makes me want to templatize them so we only have to express the meaningful bits once.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -263,29 +312,111 @@ renderGhcjsFrontend urlEnc ghcjsWidgets route configs f = do
...
+  :: Text
+  -- ^ URL (could be relative) of "all.js"
+  -> (Text, WasmAssets)
+  -- ^ Root URL (could be relative) of wasm assets along with
+  -- the record of wasm assets.
+  -> (Text, Text, (Int -> Text))
+  -- ^ (preload script, deferred run script, delayed run script (Int input is the delay))
+wasmScripts allJsUrl' (wasmRoot, wAssets) = (preloadScript, runJsScript, delayedJsScript)
+  where
+    wrapName f = "'" <> wasmRoot <> "/" <> f wAssets <> "'"
```
3noch, line 340: Since this stuff is user-supplied we should do at least basic sanitizing on it. Maybe replacing `'` with `\'` at least.

ryantrinkle: Isn't there some library (jmacro?) that actually knows how to do this properly?  We should not be implementing escaping ourselves, and we certainly shouldn't be doing it inline in other code.

lib/backend/src/Obelisk/Backend.hs
ryantrinkle: Agreed

bidigo:
Just a quick question: any idea if this is going to be merged anytime soon?

dfordivam:
@bidigo this will take a while. It needs more testing and performance improvements. There is a more recent work done on this branch https://github.com/obsidiansystems/obelisk/tree/dn-jsaddle-core2-2-wasm

bidigo:
@dfordivam Thx for the info (and thanks for the great work!)

PR #711, opened 2020-04-21
command: Unpack latest thunk type inplace instead of in local

3noch:
I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -14,13 +14,14 @@ import Control.Applicative
 import Control.Exception (displayException, try)
 import Control.Lens (ifor, ifor_, (.~))
 import Control.Monad
-import Control.Monad.Extra (findM, unlessM, whenM)
+import Control.Monad.Extra (findM)
 import Control.Monad.Catch (MonadCatch, handle)
 import Control.Monad.Except
 import Data.Aeson ((.=))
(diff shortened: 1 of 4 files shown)


3noch:
Still one failing test in selftest...

lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -555,7 +541,7 @@ let fetch = { private ? false, fetchSubmodules ? false, owner, repo, rev, sha256
     inherit owner repo rev sha256 fetchSubmodules private;
   };
   json = builtins.fromJSON (builtins.readFile ./github.json);
-in if builtins.pathExists ./local then ./local else fetch json
+in fetch json
```
kmicklas, line 544: This causes `Thunk update: ReadThunkError_UnrecognizedThunk` when trying to update a thunk from before this change. I think `ThunkSpec`s need to be immutable.

ryantrinkle:
@kmicklas I think the ThunkSpecs are meant to be immutable once released, but I don't think the one with `local` was actually released.  If it was, we should definitely add support for it.  I do think it's probably a bit tricky to try to maintain support for anything that was ever on develop, since we could find bugs in the ThunkSpec itself after that.

@3noch Any thoughts?

3noch:
@kmicklas Yes that format was never released. It made it to `develop` but never `master`. Sadly there are probably some of us that accidentally ended up with the broken intermediate version.

PR #715, opened 2020-04-23
Rename prismEncoder and isoEncoder to prismDecoder and isoDecoder to signal contravariance

cardenaso11:
prismEncoder and isoEncoder have also been deprecated as part of this


ryantrinkle:
They're still Encoders, though.  Clarifying the contravariance, and possibly changing the names, seems like a good idea, but I think we should pick different new names.

Also, while the Prism actually does need to be contravariant, I can't see why the Iso one would need to be.  Maybe we should fix this?

3noch:
Yeah sadly we stole the good names for the wrong impls. Any good colors for the bikeshed? Seems we need 3 names: 2 contravariant-clarifying names and one for a covariant isoEncoder.

ryantrinkle:
What are the names in `lens` for actually *using* these things?  `view`?  Also: do we even need both?  Isos are valid prisms, no?

xplat:
Isos are valid prisms indeed, so we really only need `prismEncoder` ... `isoEncoder` might be able to get away with fewer constraints though since an Iso can't fail in either direction.  Isos are used with `view` and `review` to directly get out the two directions, and prisms with `preview` and `review`.  Isos can use `from` to reverse them as an Iso and prisms can be reversed with `re` but only a Getter comes out so we can't expect people to pre-reverse them ...

(Some optics library variants have an `Unprism` type for that but lens doesn't--the hybrid profunctor/classic van Laarhoven encoding of optics in lens exploits the use of `->` as the profunctor whenever representable profunctors are sufficient, while not doing the same for corepresentable, but this twists the way the optics subtyping lattice is expressed as constraints, hiding symmetries in a way that makes it difficult to write a generic optic-reverser that works correctly.)

cardenaso11:
Lens has view and review for using these, where view will look inside of them and review will bury them. Isos are valid prisms, so really there's no reason to actually /use/ isoEncoder over prismEncoder I don't think. In that case, is just prismDecoder an appropriate name?

The names for using these things in lens are just view to "parse" and review to "unparse", but I'm not entirely sure how to make a good name from that...

xplat:
it's actually `view` that is only for isos ... prisms have to use `preview` because the `view` direction can fail.

cardenaso11:
eek, right, typo on my part

3noch:
Even just for the sake of code clarity I like having `isoEncoder` just because it communicates a stronger guarantee. I.e. I'd want to use that when possible even if `prismEncoder` would have worked.

cardenaso11:
Okay, would isoEncoder (contravariant), isoDecoder (covariant), prismDecoder (contravariant) work?

ryantrinkle:
@cardenaso11 No.  Nothing can be named `*Decoder`: regardless of their input, they are all still producing a value of type `Encoder`.

- Let's call this thing `reviewEncoder` - it is the `review` arrow in the `Encoder` category
- More generally, if we have a function `f :: a -> b` and there is a unique most-sensible `Encoder` for it, the `Encoder` should be named `fEncoder`
- Even more generally, this module is designed to encourage users to think in terms of how they translate their route datastructure into a raw URL, not how they decode it.
  - If we go the other direction, the `case` statements are more awkward - they *must* scrutinize the decoded values, not the encoded ones, because otherwise the completeness checking does not work.
  - It does make it a bit awkward to think about what happens in the presence of non-canonical encodings of routes - e.g. `someStrictEncoder` (which only accepts the exact canonical encoding when it decodes things) and `someLenientEncoder` (which accepts more than that) makes less sense to me than `someStrictDecoder` and `someLenientDecoder` would.
  - Regardless of whether "think of everything in terms of encoding" is the right choice or not, it's absolutely critical that we maintain consistency.  Thinking about *some* things in terms of encoding and *other* things in terms of decoding would be hellish.
- Let's find a way to integrate these explanations into the documentation of the library itself, so that contributors in the future have a way of understanding how these decisions are made.

3noch:
@ryantrinkle I've linked to your comment on our backlog regarding docs for this library.

lib/route/src/Obelisk/Route.hs
```diff
@@ -816,14 +823,20 @@ _R
...
 -- | An encoder that only works on the items available via the prism. An error will be thrown in the parse monad
 -- if the prism doesn't match.
 prismEncoder :: (Applicative check, MonadError Text parse) => Prism' b a -> Encoder check parse a b
-prismEncoder p = unsafeMkEncoder $ EncoderImpl
-  { _encoderImpl_encode = (^. re p)
+prismEncoder = reviewEncoder
+
+-- | An encoder that only works on the items available via the prism. An error will be thrown in the parse monad
+-- if the prism doesn't match.
+reviewEncoder :: (Applicative check, MonadError Text parse) => Prism' b a -> Encoder check parse a b
```
3noch, line 834: Let's call out that this is contravariant just like `review` from lens. Otherwise everything looks great.

cardenaso11: Okie, added documentation, is that good?

3noch: Whoops just forgot the `f` at the end in both cases.

cardenaso11: eek, thank you!

lib/route/src/Obelisk/Route.hs
```diff
@@ -825,14 +832,22 @@ _R
   -> Prism' (R f) a
 _R variant = dSumGEqPrism variant . iso runIdentity Identity
 
+{-# DEPRECATED prismEncoder "Use reviewEncoder instead" #-}
```
ryantrinkle, line 835: As with the other deprecation:

1. Make the replacement rule blatantly clear in the deprecation warning; the user should not need to understand the deprecated function OR the new function in order to fix the deprecation warning.  (Imagine someone assigned you to bump a dep for a library you've never heard of in a codebase you're not familiar with - this is not uncommon!)
2. Move the deprecated function to a "deprecated" section at the end of the file.

lib/route/src/Obelisk/Route.hs
```diff
@@ -825,14 +832,22 @@ _R
...
 -- | An encoder that only works on the items available via the prism. An error will be thrown in the parse monad
 -- if the prism doesn't match.
+-- This encoder is contravariant like 'review': @prismEncoder (f . g) = prismEncoder g . prismEncoder f@
```
ryantrinkle, line 838: Let's down-level the language here.  Why rely on the user understanding the concept of contravariance?

> `Note that a 'Prism' from @a@ to @b@ will produce an 'Encoder' from @b@ to @a@ (i.e. 'prismEncoder' is a contravariant functor from the category of prisms to the category of encoders), just like `review` produces a function @b -> a@.  This is because 'Prism's extract values, in a way that might fail, in their forward direction and inject values, in a way that cannot fail, in their reverse direction; whereas Encoders encode, which cannot fail, in their forward direction, and decode, which can fail, in their reverse direction.`

This way, we're not relying on the user already knowing the terminology, but we are teaching it to them.  Also, we're explaining the intuition of *why* this is the case.

lib/route/src/Obelisk/Route.hs
```diff
@@ -825,14 +832,22 @@ _R
...
-prismEncoder p = unsafeMkEncoder $ EncoderImpl
-  { _encoderImpl_encode = (^. re p)
+prismEncoder = reviewEncoder
+
+-- | An encoder that only works on the items available via the prism. An error will be thrown in the parse monad
+-- if the prism doesn't match.
+-- This encoder is contravariant like 'review': @reviewEncoder (f . g) = reviewEncoder g . reviewEncoder f@
+reviewEncoder :: (Applicative check, MonadError Text parse) => Prism' b a -> Encoder check parse a b
+reviewEncoder p = unsafeMkEncoder $ EncoderImpl
+  { _encoderImpl_encode= view (re p)
```
ryantrinkle, line 847: Missing space

ryantrinkle: Is there any particular reason to use `view . re` instead of `review` here?

lib/route/src/Obelisk/Route.hs
```diff
@@ -433,18 +435,23 @@ instance (Applicative check, Monad parse) => Monoidal (Encoder check parse) (,)
...
+{-# DEPRECATED isoEncoder "Use viewEncoder instead" #-}
 -- | Given a valid 'Iso' from lens, construct an 'Encoder'
 isoEncoder :: (Applicative check, Applicative parse) => Iso' a b -> Encoder check parse a b
-isoEncoder f = unsafeMkEncoder $ EncoderImpl
-  { _encoderImpl_encode = view f
-  , _encoderImpl_decode = pure . view (from f)
+isoEncoder p = viewEncoder (from p)
+
+-- | Given a valid 'Iso' from lens, construct an 'Encoder'
+viewEncoder :: (Applicative check, Applicative parse) => Iso' a b -> Encoder check parse b a
```
ryantrinkle, line 444: Why is this contravariant?  [`view`](https://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Getter.html#v:view) is covariant.

lib/route/src/Obelisk/Route.hs
3noch: This is a good idea. I was going to address this later when we actually did a full-scale attack on the docs of this module in general, but no harm in starting now.

3noch:
@ryantrinkle Should be ready for your review

PR #720, opened 2020-05-04
run: Set preferred host based on route

### LINKED ISSUE #719: ob run should not serve at 0.0.0.0
When developing an obelisk app locally, the server started by `ob run` is exposed to everyone on the network. 

Incidentally using `127.0.0.1` in common/route has no effect (`Run.hs:runWidget` sets only port, not host, from the route URI).

3noch:
Fixes #719 

I have:


3noch:
@srid Does this solve your issue?

srid:
@3noch It does, yes!

3noch:
Awesome. Thanks for the report.

3noch:
:thinking: It appears we no longer detect when two `ob run`s are using the same port. @tomsmalley can you confirm? I'm pretty sure that feature doesn't work on a mac anyway so this may only be reproducible on Linux.

srid:
@3noch I can reproduce that on my NixOS machine. Seems strange to me that this particular change would cause that regression ...

3noch:
There's something about binding to a particular host that causes the binding to no longer *fail*.

PR #721, opened 2020-05-06
Add nix attribute for qemu/virtualbox image

alexfmpe:
I have:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -167,29 +167,37 @@ in rec {
 
   inherit mkAssets;
 
-  serverExe = backend: frontend: assets: optimizationLevel: version:
+  serverExe = backend: frontend: assets: optimizationLevel: bundledConfig: version:
     pkgs.runCommand "serverExe" {} ''
       mkdir $out
       set -eux
       ln -s "${if profiling then backend else haskellLib.justStaticExecutables backend}"/bin/* $out/
       ln -s "${mkAssets assets}" $out/static.assets
       ln -s ${mkAssets (compressedJs frontend optimizationLevel)} $out/frontend.jsexe.assets
+      ${pkgs.lib.optionalString (bundledConfig != null) "ln -s ${bundledConfig} $out/config"}
       echo ${version} > $out/version
     '';
 
+  mkNixOS = imports: import (pkgs.path + /nixos) {
+    system = "x86_64-linux";
+    configuration = { inherit imports; };
+  };
+
+  serverImports = args: [
+    (serverModules.mkDefaultNetworking args)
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -348,7 +362,10 @@ in rec {
 
       linuxExeConfigurable = linuxExe;
       linuxExe = linuxExe dummyVersion;
-      exe = serverOn mainProjectOut dummyVersion;
+      exe = serverOn mainProjectOut null dummyVersion;
+      inherit vm;
+      ova = args@{ hostName, adminEmail, routeHost, enableHttps, version, module ? (_: _: {}) }:
+        (vm args).config.system.build.virtualBoxOVA;
```
3noch, line 368: I think we should put these behind an `__unstable__` attr for now until we actually officially use them inside obelisk. Would that mess up your ability to use it?

alexfmpe: Shouldn't hinder me.
Do you mean the same [`__unstable__` where `profiledObRun` is](https://github.com/obsidiansystems/obelisk/pull/721/files#diff-5712e736e0de6ba170577f8472c398e9R335)?

3noch: Yes

default.nix
```diff
@@ -167,29 +167,37 @@ in rec {
 
   inherit mkAssets;
 
-  serverExe = backend: frontend: assets: optimizationLevel: version:
+  serverExe = backend: frontend: assets: optimizationLevel: bundledConfig: version:
```
matthewbauer, line 170: might want to switch this to a struct so that we can provide defaults for things like bundledConfig

3noch: Agreed.

alexfmpe: In this PR?

3noch: Sure.

default.nix
```diff
@@ -317,17 +317,23 @@ in rec {
...
-        projectInst.passthru.staticFiles
-        projectInst.passthru.__closureCompilerOptimizationLevel
-        bundledConfig
-        version;
-      linuxExe = serverOn (projectOut { system = "x86_64-linux"; }) null;
-      ovaExe = serverOn (projectOut { system = "x86_64-linux"; }) (base' + /config);
+      serverOn = { projectInst, bundledConfig ? null, version } : serverExe {
+        inherit bundledConfig version;
+        inherit (projectInst.ghc) backend;
+        inherit (mainProjectOut.ghcjs) frontend;
```
alexfmpe, line 323: `mainProjectOut != projectInst` in `linuxExe` (and consequently in `ovaExe`). Is the extra indirection undesirable only for this attribute?

PR #728, opened 2020-05-12
cli-app: expose createProcess

jmininger:
Expose the `createProcess` function written in `lib/cliapp/src/Obelisk/CliApp/Process.hs`

Note that the Process.hs module exports it, but it is not on the list of functions exposed by the package.


3noch:
Can we rebase on `develop`?

PR #730, opened 2020-05-12
Remove COMPLETE pragma and add warning for (:.) on GHC < 8.10

tomsmalley:
I have:


lib/route/src/Obelisk/Route.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
```
3noch, line 3: This should always be first because it can affect other pragmas.

PR #733, opened 2020-05-14
Add initial tutorial for obelisk-routes

mankyKitty:
This adds an initial tutorial for obelisk-routes, along with some rationale and a birds-eye view of the types involved.

I have:

### DIFF
--- a/lib/route/ROUTING.md
+++ b/lib/route/ROUTING.md
@@ -1,0 +1,202 @@
+# Obelisk Routing
+
+This document contains examples and introductory material for using the
+`obelisk-routes` package in your application.
+
+The `obelisk-routes` package is designed to help with managing the paths and
+parameters for the various pages in your application. These routes are for
+organising and structuring the navigation of your application.
+
+This is distinct from packages like `servant` that are designed for building
+endpoints for a REST API. When building these routes you're not going to be
+handling headers or HTTP methods.
+
+One of the goals of `obelisk-routes` is to allow for complex routing
+structures to be partitioned into manageable chunks that are then combined in
+a straight-forward manner. Allowing for re-use of various components and to
+help the type-system provide as many guarantees as possible.
+
+## Building blocks
+
+There are a few pieces required to build routes in Obelisk and there is plenty to dig into
+if you are curious. However for our purposes we can survive with a surface level
+understanding.
+
+#### `SegmentResult`
+
+This is the primary building block of individual routes, it has a list-like
+structure of; a recursive component, and a terminating case.
+
+The `PathEnd ...` constructor may be used to indicate that there is only this
+single segment in that route, or to terminate a longer route.
+
+Or the `PathSegment "user" ...` to encode "user" as a literal segment in the
+path, followed by more `Encoder`s.
+
+#### `Encoder`
+
+An `Encoder` describes how the either an individual segment of a route is to
+be encoded. Such as how to read the number from a route like `/user/33`.
+
+It can also describe how several routes are encoded. So you may have an
+`Encoder` that describes how all of the routes pertaining to 'blog posts' are
+encoded.
+
+In particular note that `Encoder`s may be composed using the composition
+operator from `Control.Category`. This means that complex route structures
+can be broken down into manageable pieces and then composed together to
+encode the entire structure.
+
+#### `BackendRoute` & `FrontendRoute`
+
+Obelisk has two distinct sets of routes for backend and frontend. These GADTs
+are called `BackendRoute` and `FrontendRoute`. These are provided for you to
+start you off, and we will create our own later on.
+
+In a fresh Obelisk application these live in the `Common.Route` module.
+
+`BackendRoute` is for serving static things or deferring to other backend endpoints. Such as:
+- "static/files/img/:imgName:"
+- "api/..." where the "..." portion is managed by a REST endpoint package of your choice.
+
+`FrontendRoute` is for managing the frontend routing of your application. Such as:
+- "blog/:userId:/:postId:"
+- "dashboard?optionA=true".
+
+Building routes for either one is the same process. Keeping the types separate enforces
+the distinction at compile time and helps to prevent errors.
+
+The routing system in Obelisk is quite powerful and we'll only touch on some
+basics here, but this should be enough to start with and we can dive into the
+gnarlier routing problems later.
+
+## An example route
+
+Consider the following `Frontend` request that we want our Obelisk
+application to be able to handle:
+
+```
+GET /user/42
+```
+
+It has a static portion "user" and some number that we want to verify and
+have available on our page for us to use.
+
+We could use a bare `Int` in our types but we're responsible Haskellers so
+define a `newtype` in `Common.Api` to wrap the number for greater type safety
+and to be more descriptive:
+
+```haskell
+newtype UserId = UserId { unUserId :: Int }
+```
+
+For Obelisk to be aware of our routes existence, we add a constructor to
+`FrontendRoute` GADT in `Common.Route`. The purpose of this constructor is to
+give us something to match on when we're deciding which code to run for which
+page. As well as declare what are the types of things that will be pulled
+from the route and available on the resulting page.
+
+Add the following constructor to `FrontendRoute`:
+
+```haskell
+  FrontendRoute_User :: FrontendRoute UserId
+```
+
+This says that we have route that will provide a `UserId` to our frontend.
+
+> The naming of the route 'FrontendRoute_User' is an Obelisk convention of including the
+> type name in individual constructor names. This helps disambiguate them in the code at
+> the cost of a few extra keystrokes. As an example:
+>
+> ```haskell
+> data Foo
+>   = Foo_ConstructorA
+>   | Foo_ConstructorB
+> ```
+
+Next we need to build the `SegmentResult` that will tell Obelisk how this route is constructed.
+
+Obelisk router breaks down incoming routes into smaller segments that are then given to
+our `Encoder` to try to find a match. We define these segments using the `SegmentResult`
+type and add them to the case expression in `Common.Route` to be matched against.
+
+Looking at our route:
+
+```
+/user/42
+```
+
+We're able to identify two segments:
+
+```
+/ "user" / (UserId 42)
+```
+
+So we need to describe the static "user" segment, and the dynamic segment that takes the
+raw input of "42" and turns it into a `UserId`.
+
+We need to tell Obelisk how to create a `UserId` from `Text` and back again. To do this
+we're going to write two functions and put them together into a `Prism'`. A `Prism'` is a
+condensed way of describing how to go from one type to another, and maybe back the other
+way.
+
+Turning our `UserId` into `Text` is easy so let's do that one first:
+
+```haskell
+userIdTextPrism :: Prism' Text UserId
+userIdTextPrism = prism toText _fromText
+  where
+    -- Unwrap the UserId, use 'show' on the 'Int' value, and then turn that into 'Text'
+    toText = T.pack . show . unUserId
+```
+
+Now we encode `Text` as a `UserId`. This means we have to check that this input is
+valid. At this stage we only care that the input is a number greater than zero:
+
+```haskell
+userIdTextPrism :: Prism' Text UserId
+userIdTextPrism = prism toText _fromText
+  where
+    -- Unwrap the UserId, use 'show' on the 'Int' value, and then turn that into 'Text'
+    toText :: UserId -> Text
+    toText = Text.pack . show . unUserId
+
+    fromText :: Text -> Either Text UserId
+    fromText t = case readMaybe $ Text.unpack x of
+      Just i -> Right $ UserId i
+      Nothing -> Left x
+```
+
+Our `Prism'` can now be used to create the rest of `SegmentResult` like so:
+
+```haskell
+PathSegment "user" $ singlePathSegmentEncoder . prismEncoder userIdTextPrism
+```
+
+We combine our `userIdTextPrism` using `prismEncoder` from `Obelisk.Route` to do the heavy
+lifting for us, and compose that using `(.)` with `singlePathSegmentEncoder` to tell
+Obelisk that this part of the route only has a single piece.
+
+It should be a rare event that you will need to write your own `Encoder` directly. The
+provided `Encoder`s in `Obelisk.Route` are designed to be composed and combined to build more
+complicated encoders and should be sufficient for most requirements.
+
+#### 'Wrapped' typeclass and Template Haskell.
+
+If you're comfortable with Template Haskell, you can use the `makeWrapped` splice from
+`Control.Lens.TH` to derive the `Wrapped` typeclass. This would let you avoid having to
+write a `Prism'`:
+
+```haskell
+PathSegment "user" $ readShowEncoder . unwrappedEncoder
+```
+
+This is handy if you already have your newtypes defined and don't want to hand write
+`Prism'`s. This uses the `_Wrapped` prism from the `lens` library and then the encoder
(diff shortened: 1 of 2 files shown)


lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
+# Obelisk Routing
+
+The `obelisk-routes` package is designed to help with managing the paths and parameters for
+routing in your application.
+
+This is distinct from packages like `servant` that are designed for specifying endpoints for a
+REST API. When building routes with `obelisk-routes` you're not going to be handling headers or
+HTTP methods.
```
3noch, line 8: Backend route handlers do need to think about HTTP methods/codes. But it's not strictly part of routing system. So I guess this is fine but it's maybe a bit surprising to a "normie" understanding of routing.

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+
+One of the goals of `obelisk-routes` is to allow for complex routing structures to be partitioned
+into manageable chunks that are then combined in a straight-forward manner. Allowing for re-use
+of various components and to help the type-system provide as many guarantees as possible. We'll
```
3noch, line 12: "Allowing for ..." isn't a full sentence. Maybe "This allows for..."

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+
+Add this to the end of our `case` expression in `Common.Route`:
+
+```haskell
+FrontendRoute_User -> PathSegment "user" $ _todo . prismEncoder userIdTextPrism
+```
+
+Next is to tell Obelisk that we have only a single path segment and once it has matched the
+static part and the `UserId` this route is complete. This is done by composing our `UserId`
+encoder with the `singlePathSegmentEncoder` using `(.)` imported from `Control.Category`:
```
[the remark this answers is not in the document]
ali-abrar: Relatedly, perhaps a note on how to get hackage docs or hoogle for one's local obelisk?

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+```
+
+We can now move onto building the `Encoder` for our `UserId`.
+
+#### Wait a minute, shouldn't I be matching on the route on the left?
+
+In many routing systems you match directly on the route input and break it down as required. This
+process can be fragile and has limitations. Additionally you are left on your own when comes to
+_creating_ links in your application. Because there is no way to relate the structure of a route
+to anything. You have to manually build up routes again, and if those routes change it can be
```
3noch, line 178:
```suggestion
to anything, you have to manually build up routes again, and if those routes change it can be
```

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+for greater type safety and to be more descriptive:
+
+```haskell
+newtype UserId = UserId { unUserId :: Int }
+```
+
+To create the route itself, we add a constructor to the `FrontendRoute` GADT in `Common.Route`.
+
+The purpose of this constructor is to give us something to match on when we're deciding which
+code to run, it will also be used to help build type-safe links within our application. As well
```
3noch, line 114:
```suggestion
code to run. It will also be used to help build type-safe links within our application. As well
```

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+- etc...
+
+In particular note that `Encoder`s may be composed, thereby creating more elaborate routing
+structures from simpler pieces. It also means that scary looking routing hierarchies can be
+partitioned and made easier to work with.
+
+#### `BackendRoute` & `FrontendRoute`
+
+An Obelisk application begins with two distinct sets of routes for backend and frontend. These
+are defined as GADTs called `BackendRoute` and `FrontendRoute`, respectively. These are provided
```
3noch, line 53: Link to some juicy GADT docs.

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+routing in your application.
+
+This is distinct from packages like `servant` that are designed for specifying endpoints for a
+REST API. When building routes with `obelisk-routes` you're not going to be handling headers or
+HTTP methods.
+
+One of the goals of `obelisk-routes` is to allow for complex routing structures to be partitioned
+into manageable chunks that are then combined in a straight-forward manner. Allowing for re-use
+of various components and to help the type-system provide as many guarantees as possible. We'll
+see this come into play when we start adding more complex routes.
```
ali-abrar, line 13: Should we mention somewhere what some of the guarantees are? E.g., non-overlapping

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+  )
+```
+
+Our route has multiple segments so we need to use the `PathSegment` constructor:
+
+```haskell
+  | PathSegment Text (Encoder check parse a PageName)
+```
+
+Matching on the first segment is done by using the literal value "user" as the first argument to
```
ali-abrar, line 164: I know there's a note below, but somehow "Matching" still doesn't feel like it's the whole story here. This is also where you *declare* the path segment.

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+are defined as GADTs called `BackendRoute` and `FrontendRoute`, respectively. These are provided
+for you to start you off.
+
+In a fresh Obelisk application these live in the `Common.Route` module.
+
+`BackendRoute` is for serving static things or deferring to other backend endpoints. Such as:
+- "static/files/img/:imgName:"
+- "api/..." where the "..." portion is managed by a REST endpoint package of your choice.
+
+`FrontendRoute` is for managing the frontend routing of your application. Such as:
```
cgibbard, line 62: Even if it's slightly redundant, it might be helpful to add a note around here something like "These are all the routes for which the backend will serve up the application frontend."

lib/route/ROUTING.md
ryantrinkle: We should provide the reason here, rather than just stating the fact, because most readers will probably think it's weird.  The reason is that our current incarnation of routes is designed to support only features that are usable on *both* the frontend and the backend.  That's because there are some laws we're trying to preserve here, like: backend pre-rendering produces a page that looks the same as what the frontend will render.  This rules out the use of the hash string, because it doesn't get sent to the backend.  (We could support the hash string if we could guarantee that the frontend doesn't use it in any way other than going to the anchor, because that's what will happen with the pre-rendered version!)

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
+# Obelisk Routing
+
+The `obelisk-routes` package is designed to help with managing the paths and parameters for
+routing in your application.
+
+This is distinct from packages like `servant` that are designed for specifying endpoints for a
+REST API. When building routes with `obelisk-routes` you're not going to be handling headers or
+HTTP methods.
+
+One of the goals of `obelisk-routes` is to allow for complex routing structures to be partitioned
```
ryantrinkle, line 10: While this is certainly a goal, most of the complexity of obelisk-routes isn't related to this goal.  The big thing we are trying to do here is provide several important guarantees:

- *Every* route value can be encoded to a URL (i.e. encoding can't fail)
- `decode . encode = pure` (i.e.: encoding roundtrips, which also implies that there's no ambiguity in the routes)
- When you make common modifications to routes, such as adding or removing a route, you get "incomplete case" and similar warnings everywhere you need to update your app - this makes it much harder to miss things when you add routes.

When I was initially investigating the options for route parsing, I couldn't find anything that achieved all three of these, and most options (e.g. boomerang) didn't achieve *any* of them.

One thing to keep in mind here is that this library asks a lot from the user, e.g. not allowing arbitrary shadowing.  This means we need to give them a lot of value in return.  Personally, I consider these guarantees to be really valuable, and I think we should probably make sure they're mentioned prominently, especially when they help explain limitations or difficult stuff.

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+    -- Unwrap the UserId, use 'show' on the 'Int' value, and then turn that into 'Text'
+    toText = T.pack . show . unUserId
+```
+
+Next to encode `Text` from the route input as a `UserId`. This means we have to check that this
+input is valid. Our requirement is that the input is a number greater than zero:
+
+```haskell
+userIdTextPrism :: Prism' Text UserId
+userIdTextPrism = prism toText fromText
```
ryantrinkle, line 234: This isn't a good approach to implementing this function.  What we should do (and teach users to do) is to break down their encoders into small pieces, and then compose those.

What this is doing right now is 3 steps: unwrapping the UserId, showing the contents, and packing the resulting string as text.  The *combination* of those three things had better round-trip with the *combination* of the three parsing steps that are listed below.  This is dangerous - and, in fact, as a result, you've written a bug!  `UserId (-1)` does not roundtrip, which means that this prism, and the encoder based on it, are both invalid.

Instead, the correct way to implement it is to write each tiny piece of roundtripping logic separately, and then compose the resulting `Encoders`.  For instance: `userIdTextEncoder = tshowEncoder . unwrappedEncoder`.  This would also have probably made it clearer that `greaterThanZeroEncoder` doesn't make sense.

lib/route/ROUTING.md
mankyKitty: resolved by deletion as part of previous comment

lib/route/ROUTING.md
mankyKitty: > We declare the first segment by using the literal value "user" as the first argument to the
constructor that we added to the `case` expression:

Changed it up as we don't match at this point, we are _declaring_ the segment here. :+1:

lib/route/ROUTING.md
mankyKitty: I wanted to expand that sentence a bit as I wasn't quite able to parse it without a bit more description. :)

> These backend routes can run their own code or hand off to a different request handling package
> to a finer grained set of endpoints. But should none of them match the frontend of the
> application is served and the frontend router will continue parsing the route.

lib/route/ROUTING.md
mankyKitty: I've overhauled the introduction to try to convey this point and a few others. Probably still rough edges...

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+>
+> ```haskell
+> data Foo
+>   = Foo_ConstructorA
+>   | Foo_ConstructorB
+> ```
+
+### Building our route from segments
+
+Next we need to build the `SegmentResult` that will tell Obelisk how this route is to be encoded
```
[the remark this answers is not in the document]
mankyKitty: I added some wording for that

lib/route/ROUTING.md
mankyKitty: The introduction has been overhauled to include a better description of the guarantees and motivation, let me know if what you've mentioned here is sufficiently covered.

lib/route/ROUTING.md
mankyKitty: Intro has been rewritten as a 'motivation' section that contains these points. :+1: It may need to be fleshed out a bit with some descriptive examples ?

lib/route/ROUTING.md
mankyKitty: But `Prism`s are partial in one direction so there's never a guarantee of being able to roundtrip?

Exposing the constructor seems to have been my mistake, because you can create an invalid `UserId` and therefore an invalid route, I think. One function to create a valid `UserId` separate to any of the routing encoders might be enough.

I didn't want to use the `unwrappedEncoder` for my initial example as using that single line of template haskell to derive the `Wrapped` code involves turning on about six language extensions. Also using the `tshowEncoder` doesn't provide any control over what is an acceptable number for a `UserId`. 

I know it's a tutorial example but I'm trying to keep the complexity budget down and avoid the Haskell meme of "just trust me and turn on all these extensions". I'll come up with a cleaner and not-busted example. :)

lib/route/ROUTING.md
```diff
@@ -0,0 +1,446 @@
...
+are curious. It may seem daunting and there is certainly a lot of power on offer. However for our
+purposes we can survive with a surface level understanding.
+
+Below we will cover a few of the types that we need to interact with to add routes to our
+application. Again you don't have to have a deep understanding of these types, a high level view
+will suffice for now.
+
+#### `SegmentResult`
+
+This is the primary building block of individual routes, it has a list-like structure with a
```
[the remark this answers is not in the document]
mankyKitty: Moved the introduction of this route to before the types overview

lib/route/ROUTING.md
ryantrinkle: The roundtripping law is that each value you can encode must decode successfully, and to the same value that was originally encoded.  That does *not* imply (and it is not the case) that decoding must be total.  Instead, this creates a requirement that anything which fails to decode must also not be a possible output of `encode`.  It is also fine for multiple values may decode to the same value; exactly one of them will be the *canonical* encoding, and so we can always canonicalize encoded values (if they are valid) by doing `fmap encode . decode`.

Keeping complexity down is a great goal, but we need to be mindful of the other goals as well, one of which is correctness.  The reason this library exists is that it is hard to write correct route handling code, and this library is able to make it somewhat easier.  The value of `unwrappedEncoder` is that it is known to be actually correct.

`tshowEncoder` not providing control over what is an acceptable number for a UserId is exactly the point - it only does the read/show part, and doesn't need to understand anything else.  It's safe as long as read/show roundtrips.

It's important here to ensure we convey to the reader what the actual purpose of this library is.  If it were a good idea for them to build their route encoder and their route decoder separately, and then turn them into an Encoder, this library would be unnecessary.  The purpose of using the Encoder abstraction is that if you have two Encoders that you know are valid, you can put them together in a way that is *also* guaranteed to be valid.

ryantrinkle: One decent heuristic to use here: `prism` should probably not appear in this tutorial, except possibly towards the end in a section about "how to unsafely build an encoder".

lib/route/ROUTING.md
```diff
@@ -0,0 +1,491 @@
...
+
+Update the `FrontendRoute_User` branch of our `case` expression to the following:
+
+```haskell
+  , _frontend_body = subRoute_ $ \case
+    FrontendRoute_User -> do
+      text "We're on the user page! But which user?"
+      dUserId <- askRoute
+      dyn_ $ ffor dUserId $ \(UserId uidVal) ->
+        text "This user : " <> T.pack (show uidVal)
```
srid, line 457: `text $`

mankyKitty:
Closing this PR for deeper discussions and a bit of an overhaul. 

Don't worry, the route docs are coming!

lib/route/ROUTING.md
```diff
@@ -0,0 +1,491 @@
...
+```
+
+It should be a rare event that you will need to write your own `Encoder` function like
+`reviewEncoder` or `singlePathSegmentEncoder` directly. The provided `Encoder`s in `Obelisk.Route`
+are designed to be composed and combined to build more complicated encoders and should be
+sufficient for most requirements. If you find this is not the case and there is a pattern that
+keeps appearing and is not accounted for, then that is probably a bug and we'd like to know about
+it!
+
+#### Aside: 'Wrapped' typeclass and Template Haskell.
```
imuli, line 347: Note: the `Wrapped` class doesn't require TemplateHaskell, if you use `DeriveGeneric` (or a `Generic` instance) the default class implementation will use that.

AJChapman:
This document as it is is better than nothing. Could I suggest that this be merged as-is and then perfected later? That way you can e.g. link to it from the Obelisk README. Currently people are linking to this PR as one of the best Obelisk docs, e.g. from the comments in https://www.reddit.com/r/haskell/comments/gki4ge/obelisk_tutorial_for_web_app/ , so I think it deserves a better place to live.

PR #740, opened 2020-05-18
Use unix-compat instead of unix for windows compat

mightybyte:
Beginning of a better Windows compatibility story for Obelisk.

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -22,7 +22,7 @@ import System.Environment
 import System.FilePath
 import qualified System.Info
 import System.IO (hIsTerminalDevice, stdout)
-import System.Posix.Process (executeFile)
+import System.Process (rawSystem)
 
 import Obelisk.App
 import Obelisk.CliApp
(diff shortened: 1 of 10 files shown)


3noch:
@mightybyte have you tested this *on* windows?

mightybyte:
Not yet.  (It's building now.)  I anticipate more obstacles before I can actually use an Obelisk app on Windows.  Things like jsaddle-wkwebview, etc.  My goal here is pretty narrow--to build an Obelisk app as a Windows desktop app with GHC.  Note that this doesn't necessarily include making the `ob` command line app work on Windows.  All I need is for the obelisk-provided libraries to build.  I also don't intend to tackle the problem of getting ghcjs to build on Windows at all.  So while this in no way gets to what's needed for full-blown Obelisk Windows support, it is a necessary precondition.

3noch:
Makes sense. As long as there's no expectation that I will start actively testing obelisk on Windows I don't mind accepting this provided nothing else suffers as a result.

mightybyte:
Oh yes, I definitely don't expect any Windows testing.

lib/command/src/Obelisk/Command.hs
```diff
@@ -363,7 +363,9 @@ main' argsCfg = do
         Just impl -> do
           -- Invoke the real implementation, using --no-handoff to prevent infinite recursion
           putLog Debug $ "Handing off to " <> T.pack impl
-          liftIO $ executeFile impl False ("--no-handoff" : myArgs) Nothing
+          --liftIO $ executeFile impl False ("--no-handoff" : myArgs) Nothing
```
3noch, line 366:
```suggestion
```

Issue #747, opened 2020-05-26
thunk unpacking does not preserve semantics of ob run


maralorn:
Thanks to your recent fixes my `ob run` is now working again.

But after

    ob thunk unpack .obelisk/impl

I get

    common/src/Prelude.hs:73:1: error:
        Ambiguous module name ‘Data.String.Interpolate’:
          it was found in multiple packages:
          interpolate-0.2.0 string-interpolate-0.1.0.1
       |
    73 | import           Data.String.Interpolate        ( i )
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

I guess I can work around that myself. Probably your choice of interpolation lib is better than mine. But still I think this not something that should happen, should it?

madeline-os:
There is a work-around for this, package-qualified imports: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/package_qualified_imports.html

Since this is surprising, I will document it in the FAQ.

PR #748, opened 2020-05-26
Don‘t interpret obelisk with language-server

maralorn:
Actually I would suggest making `--no-interpret .obelisk/impl` a default for `ob run` and friends. Does anyone ever need it differently? That would likely address #747, too?


3noch:
Ideally we want ob run and ghcide to agree moat of the time so I'd rather think about how to do that. Any ideas?

maralorn:
Well I don't know your code that good, but I guess at some point you define defaults for which dirs to interpret and which not?
Setting that option as global default could work, couldn't it?

3noch:
I have different plans for this for now.

PR #752, opened 2020-05-29
cliapp, command: Preserve exit code for subprocesses

3noch:
This fixes a long-standing TODO to use `bracket`ing in cliapp when running processes. It normalizes how failed processes get turned into exceptions. A new function `runProcess_` was added to fix several places that were not propagating underlying exit failure codes.

I have:

### DIFF
--- a/lib/cliapp/src/Obelisk/CliApp/Process.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Process.hs
@@ -17,6 +17,7 @@ module Obelisk.CliApp.Process
   , callProcessAndLogOutput
   , createProcess
   , createProcess_
+  , exitCodeToException
   , overCreateProcess
   , proc
   , readCreateProcessWithExitCode
@@ -24,6 +25,7 @@ module Obelisk.CliApp.Process
   , readProcessAndLogStderr
   , readProcessJSONAndLogStderr
   , reconstructCommand
+  , runProcess_
   , setCwd
   , setDelegateCtlc
   , setEnvOverride
@@ -32,6 +34,7 @@ module Obelisk.CliApp.Process
   ) where
 
 import Control.Monad ((<=<), join, void)
+import Control.Monad.Catch (MonadMask, bracketOnError)
 import Control.Monad.Except (throwError)
 import Control.Monad.Fail
 import Control.Monad.IO.Class (MonadIO, liftIO)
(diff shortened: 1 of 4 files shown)


lib/cliapp/src/Obelisk/CliApp/Process.hs
```diff
@@ -212,18 +212,28 @@ callCommand cmd = do
...
-  waitForProcess p >>= \case
-    ExitSuccess -> return (out, err)
-    ExitFailure code -> throwError $ review asProcessFailure $ ProcessFailure (Process.cmdspec $ _processSpec_createProcess process) code
+withProcess process f =
+  bracketOnError
+    (createProcess $ overCreateProcess
+      (\x -> x { std_out = CreatePipe , std_err = CreatePipe }) process
+    )
+    (liftIO . Process.cleanupProcess)
+    (\(_, Just out, Just err, p) -> do
```
mankyKitty, line 223: If we don't get two `Just ..` values then this will fail with an ugly pattern failure exception, we should handle that even if it's "impossible".

lib/cliapp/src/Obelisk/CliApp/Process.hs
```diff
@@ -242,6 +252,12 @@ streamToLog stream = fix $ \loop -> do
 waitForProcess :: MonadIO m => ProcessHandle -> m ExitCode
 waitForProcess = liftIO . Process.waitForProcess
 
+-- | Converts an 'ExitCode' to an exception when it's non-zero.
+exitCodeToException :: (CliThrow e m, AsProcessFailure e) => ProcessSpec -> ExitCode -> m ()
+exitCodeToException spec = \case
+  ExitSuccess -> pure ()
+  ExitFailure code -> throwError $ review asProcessFailure $ ProcessFailure (Process.cmdspec $ _processSpec_createProcess spec) code
```
mankyKitty, line 259: Does this resolve the `TODO` on line 233 ?

3noch: No this just turns `ExitCode` into exception but any actual exceptions pass through as normal. Now that I think about it that TODO is probably just wrong and I should delete it.

lib/cliapp/src/Obelisk/CliApp/Process.hs
3noch: Fixed

PR #774, opened 2020-06-29
Include cpp-options in OPTIONS_GHC with preprocessor

3noch:
I thought this would help but it actually doesn't. The CPP pass happens *before* our preprocessor runs so we're too late to inject any new defines/etc. Instead we need to additionally wrap the cpp and insert our fields in first using `-pgmP`. But I still think these changes are worth keeping since the code is cleaner and we'll need some of this anyway when we go to fix it correctly.

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Preprocessor.hs
+++ b/lib/command/src/Obelisk/Command/Preprocessor.hs
@@ -54,42 +54,42 @@ applyPackages origPath inPath outPath packagePaths' = withFile outPath WriteMode
   where
     hPutTextBuilder h = BU.hPutBuilder h . TL.encodeUtf8Builder . TL.toLazyText
 
-
- -- I'm pretty sure there's a certain amount of oversimplification in CabalPackageInfo, so I doubt this is fully robust.
-
 generateHeader :: FilePath -> CabalPackageInfo -> TL.Builder
 generateHeader origPath packageInfo =
     hsExtensions <> ghcOptions <> lineNumberPragma origPath
   where
     hsExtensions =
-      if not (null extList)
-      then TL.fromText "{-# LANGUAGE "
-        <> mconcat (intersperse (TL.fromText ", ") extList)
-        <> TL.fromText " #-}\n"
-      else mempty
+      if null extList
+        then mempty
+        else pragma $ TL.fromText "LANGUAGE " <> mconcat (intersperse (TL.fromText ", ") extList)
     extList = addDefaultLanguage $ concatMap showExt $ _cabalPackageInfo_defaultExtensions packageInfo
     addDefaultLanguage =
       case _cabalPackageInfo_defaultLanguage packageInfo of
         Nothing -> id
         Just x -> case x of
-          UnknownLanguage ext -> ( TL.fromString ext :)
-          ext -> ( TL.fromString (show ext) :)
+          UnknownLanguage ext -> (TL.fromString ext :)
+          ext -> (TL.fromString (show ext) :)
     showExt = \case
       EnableExtension ext -> [TL.fromString (show ext)]
       DisableExtension _ -> []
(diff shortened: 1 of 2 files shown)


lib/command/src/Obelisk/Command/Preprocessor.hs
```diff
@@ -17,40 +17,39 @@ import System.FilePath (hasTrailingPathSeparator, joinPath, splitPath)
...
-        Left err ->
-          hPutStrLn stderr $ "Error: Unable to parse cabal package " <> packagePath <> "; Skipping preprocessor on " <> origPath <> ". Error: " <> show err
-        Right (_warnings, packageInfo) ->
-          hPutTextBuilder outFile (generateHeader origPath packageInfo)
-
-  BL.readFile inPath >>= BL.hPut outFile
+    [] -> hPutStrLn stderr $ "Error: Unable to find cabal information for " <> origPath <> "; Skipping preprocessor."
+    packagePath:_ -> parseCabalPackage' packagePath >>= \case
+      Left err ->
+        hPutStrLn stderr $ "Error: Unable to parse cabal package " <> packagePath <> "; Skipping preprocessor on " <> origPath <> ". Error: " <> show err
```
lpsmith, line 45: Note that there used to be a `BL.readFile origPath >>= BL.hPut hOut` here that doesn't exist after factoring out `writeOutput`:  this was intentional in the original implementation,  otherwise I don't think GHC will see anything.  

I'm a little unclear on the exact semantics of the `-pgm*` options,  but if this change is also intentional,  then I'm guessing that

1.  the preprocessor should probably return something other than `0` as the process error code   (maybe it should do this in any case?  I dunno, it might be worth some additional experimentation/reading the code regarding how GHC actually responds to this)
2.  The messages above should be probably remove the "skipping preprocessor" language.  (Though, technically previously they were more akin to warnings than errors.)

3noch: No this was unintentional. Good catch.

lib/command/src/Obelisk/Command/Preprocessor.hs
```diff
@@ -41,14 +42,14 @@ applyPackages origPath inPath outPath packagePaths' = do
   case matches of
     [] -> hPutStrLn stderr $ "Error: Unable to find cabal information for " <> origPath <> "; Skipping preprocessor."
```
lpsmith, line 43: There was also essentially a `writeOutput Nothing inPath outPath` here as well,  before these changes.

lib/command/src/Obelisk/Command/Preprocessor.hs
```diff
@@ -17,79 +18,78 @@ import System.FilePath (hasTrailingPathSeparator, joinPath, splitPath)
...
-        Right (_warnings, packageInfo) ->
-          hPutTextBuilder outFile (generateHeader origPath packageInfo)
-
-  BL.readFile inPath >>= BL.hPut outFile
+    [] -> hPutStrLn stderr $ "Error: Unable to find cabal information for " <> origPath <> "; Skipping preprocessor."
+    packagePath:_ -> parseCabalPackage' packagePath >>= \case
+      Left err -> do
+        hPutStrLn stderr $ "Error: Unable to parse cabal package " <> packagePath <> "; Skipping preprocessor on " <> origPath <> ". Error: " <> show err
+        writeOutput Nothing inPath outPath
+      Right (_warnings, packageInfo) -> writeOutput (Just packageInfo) inPath outPath
```
lpsmith, line 48: Now I'm wondering if we shouldn't be printing `_warnings` out to stderr as well.

3noch: Yeah we do that in ob run anyway though so it's probably superfluous.

PR #776, opened 2020-07-02
Fix asset creation when asset file names aren't valid nix drv names

3noch:
I have:

### DIFF
--- a/lib/asset/assets.nix
+++ b/lib/asset/assets.nix
@@ -20,7 +20,7 @@ zopfliEncodings = file:
   nixpkgs.stdenv.mkDerivation {
     name = "encodings";
 
-    input = file;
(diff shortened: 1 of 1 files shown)


matthewbauer:
It looks like something similar was added to nixpkgs, recently:

https://github.com/NixOS/nixpkgs/blob/3b4df94aeb6e215085d08e3d5b0edc1313b9f584/lib/strings.nix#L698-L726

I don't think we have access to it, but maybe we should leave a note to use that once it's available.

3noch:
Aha thanks. I did ask on IRC if such a thing existed already. Thanks for locating it.

3noch:
FTR These names don't case path collisions because they are appended to the hash of the file in the nix store path.

PR #780, opened 2020-07-09
Make project out extensible

3noch:
This allows power users to extend `exeConfigurable` and get modified versions of `server` and `linuxExe`, etc. that reflect their modifications.

I have:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -328,40 +328,39 @@ in rec {
               });
             in allConfig;
         in (mkProject (projectDefinition args)).projectConfig);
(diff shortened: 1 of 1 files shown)


madeline-os:
I have converted this to a draft while we work out some finer details. There are attributes of `server` and `linuxExe` that are not extensible, such as `version`, and we do not want to confuse users who venture to override them. Once we have a clean separation of attributes that obelisk must control, and attributes that the user may override, we can revisit this.

PR #785, opened 2020-07-10
Add --preserve-symlinks option to thunk operations

lpsmith:
From the `ob thunk pack --help` documentation:

```
Usage: ob thunk pack THUNKDIRS... [--preserve-symlinks] [-f|--force]
                     ([--private] | [--public])
  Pack git checkout or unpacked thunk into thunk that points at the current
  branch's upstream

Available options:
  THUNKDIRS...             Paths to directories containing thunk data
  --preserve-symlinks      when a THUNKDIR is a symlink, overwrite the target of
                           the symlink, instead of overwriting the symlink
                           itself
  -f,--force               Force packing thunks even if there are branches not
                           pushed upstream, uncommitted changes, stashes. This
                           will cause changes that have not been pushed upstream
                           to be lost; use with care.
  --private                Mark thunks as pointing to a private repository
  --public                 Mark thunks as pointing to a public repository
  -h,--help                Show this help text
```

Though, it occurred to me this morning that there is another implementation with a different set of tradeoffs,  namely, if a `THUNKDIR` is a symlink we resolve the symlink ourselves using the `readlink(2)` system call,  which means we have to then check if the target is also a symlink and also detect symlink loops.   Then we pass the resolved symlink as the target of the thunk operation.   (Actually, in hindsight, one can get this behavior simply by `ob thunk pack $(readlink -f THUNKDIR)`, though maybe this option is still worthwhile)

Advantages versus this proposed `--preserve-symlink` implementation:
  + less impact on the implementation 
  + more efficient when `thunk unpack`'ing when the symlink points to another filesystem,  as the current implementation creates a temp dir in same directory the symlink is,  then uses `renamePath` to move the contents.  
Disadvantages
  + this implementation only needs write permissions inside the symlink target:  whereas resolving the symlink beforehand would require write permissions to the target's parent directory.

Would also be nice to codify this behavior in the test suite.

I have:

  - [*] Based work on latest `develop` branch
  - [*] Looked for lint in my changes with `hlint .` (lint found code you did not write can be left alone)
  - [*] Run the test suite: `$(nix-build -A selftest --no-out-link)`


lib/command/src/Obelisk/Command/Thunk.hs
```diff
@@ -349,17 +371,32 @@ parseGitSource v = do
...
 
+removeThunkPath :: MonadIO m => Bool -> FilePath -> m ()
+removeThunkPath preserveSymlinks target = liftIO $ do
+  if preserveSymlinks
+    then do
+      ps <- listDirectory target
+      forM_ ps $ \p -> do
+        removePathForcibly (target </> p)
+    else do
+      putStrLn ("removing " ++ show target)
```
3noch, line 397: We need to use obelisk's logging facility here and capitalize log messages.

lpsmith: oops, that's a wayward debugging statement.  It can be removed altogether as far as I'm concerned.

mankyKitty:
@ali-abrar Should be moved to the `nix-thunk` repo ?

PR #787, opened 2020-07-10
Add immutable directive to Cache-Control header

lpsmith:
When serving up a Content-Addressed static asset,  include immutable on the cache-control header

However, I did notice that when serving up a asset that is Content-Addressed and redirects are disabled, we are disabling all caching.   We could modify [`Obelisk.Asset.Serve.Snap.serveAsset'`](https://github.com/obsidiansystems/obelisk/blob/4f2d3f1c1312d833ae814b7a86cf7b65ea0614ec/lib/asset/serve-snap/src/Obelisk/Asset/Serve/Snap.hs#L74) to be less conservative and more precise on this regard:  in doing so, we'd need to remember the whether or not the _first_ call to `serveAsset'` before it recurses is an `immutable` or a `redirect`,  and then use that instead of `doRedirect` on [line 94](https://github.com/obsidiansystems/obelisk/blob/4f2d3f1c1312d833ae814b7a86cf7b65ea0614ec/lib/asset/serve-snap/src/Obelisk/Asset/Serve/Snap.hs#L74).

I have:

### DIFF
--- a/lib/snap-extras/src/Obelisk/Snap/Extras.hs
+++ b/lib/snap-extras/src/Obelisk/Snap/Extras.hs
@@ -17,7 +17,7 @@ import System.Directory
 -- | Set response header for "permanent" caching
 cachePermanently :: MonadSnap m => m ()
 cachePermanently = do
-  modifyResponse $ setHeader "Cache-Control" "public, max-age=315360000"
+  modifyResponse $ setHeader "Cache-Control" "public, max-age=315360000, immutable"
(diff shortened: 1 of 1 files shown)


3noch:
> and redirects are disabled

How does this happen?

lpsmith:
> > and redirects are disabled
> 
> How does this happen?

By calling either `serveAssetsInPlace` or `serveAssetInPlace`.  I didn't look at how those particular routines are used in typical codebases, though.

ali-abrar (approved):

PR #788, opened 2020-07-10
Add staticWhich around most executable paths in obelisk-command

lpsmith:
I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -184,10 +184,10 @@ deployPush deployPath getNixBuilders = do
   isClean <- checkGitCleanStatus deployPath True
   when (not isClean) $ do
     withSpinner "Committing changes to Git" $ do
-      callProcessAndLogOutput (Debug, Error) $ proc gitPath
-        ["-C", deployPath, "add", "."]
-      callProcessAndLogOutput (Debug, Error) $ proc gitPath
-        ["-C", deployPath, "commit", "-m", "New deployment"]
+      callProcessAndLogOutput (Debug, Error) $
+        gitProc deployPath ["add", "."]
+      callProcessAndLogOutput (Debug, Error) $
+        gitProc deployPath ["commit", "-m", "New deployment"]
   putLog Notice $ "Deployed => " <> T.pack route
   where
     callProcess' envMap cmd args = do
--- a/lib/command/src/Obelisk/Command/Deploy.hs
+++ b/lib/command/src/Obelisk/Command/Deploy.hs
@@ -87,11 +87,12 @@ deployInit' thunkPtr (DeployInitOpts deployDir sshKeyPath hostnames route adminE
   --production data backed up to a dev environment.
   withSpinner "Creating project configuration directories" $ do
     callProcessAndLogOutput (Notice, Error) $
-      proc "mkdir" [ "-p"
-                   , deployDir </> "config" </> "backend"
-                   , deployDir </> "config" </> "common"
-                   , deployDir </> "config" </> "frontend"
-                   ]
+      proc mkdirPath
+        [ "-p"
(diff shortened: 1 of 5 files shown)


lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -182,9 +184,9 @@ deployPush deployPath getNixBuilders = do
   isClean <- checkGitCleanStatus deployPath True
   when (not isClean) $ do
     withSpinner "Committing changes to Git" $ do
-      callProcessAndLogOutput (Debug, Error) $ proc "git"
+      callProcessAndLogOutput (Debug, Error) $ proc gitPath
         ["-C", deployPath, "add", "."]
-      callProcessAndLogOutput (Debug, Error) $ proc "git"
+      callProcessAndLogOutput (Debug, Error) $ proc gitPath
```
3noch, line 189: This should use `gitProc`

lib/command/src/Obelisk/Command/VmBuilder.hs
```diff
@@ -98,12 +98,12 @@ setupNixDocker stateDir = withSpinner ("Creating Docker container named " <> con
...
   callProcessAndLogOutput (Debug, Error) $
-    proc "ssh-keygen" ["-t", "ed25519", "-f", stateDir </> sshKeyFileName, "-P", ""]
+    proc sshKeygenPath ["-t", "ed25519", "-f", stateDir </> sshKeyFileName, "-P", ""]
 
   -- Build the docker container (which uses the SSH keys in the 'ssh' folder)
   containerId <- fmap T.strip $ readProcessAndLogStderr Error $
-    proc "docker" ["build", stateDir, "--quiet"]
-  callProcessAndLogOutput (Debug, Error) $ proc "docker"
+    proc dockerPath ["build", stateDir, "--quiet"]
+  callProcessAndLogOutput (Debug, Error) $ proc dockerPath
```
3noch, line 106: Let's leave docker out of this for now. None of these commands will work unless they have docker installed anyway and this is only for mac.

lpsmith: docker has been left out of it.

lib/command/src/Obelisk/Command/Deploy.hs
lpsmith: Fixed

lib/command/src/Obelisk/Command/VmBuilder.hs
3noch: Excellent thanks.

lib/command/src/Obelisk/Command/Utils.hs
```diff
@@ -39,6 +39,9 @@ mvPath = $(staticWhich "mv")
 rmPath :: FilePath
 rmPath = $(staticWhich "rm")
 
+mkdirPath :: FilePath
+mkdirPath = $(staticWhich "mkdir")
```
3noch, line 43: Hm. Why wouldn't we just use haskell to do this ...

lpsmith: Well, we'd have to emulate the `mkdir -p` option in this case, as the `mkdir` command line utility is slightly more than a thin wrapper around the `mkdir(2)` syscall.

That's very doable, but would have to be done.  And I didn't really treat this PR as more than a mechanical transformation + should I apply said mechanical transformation in a particular context.

Honestly, I didn't consider _not_ `staticWhich`'ing `mkdir` in this case.  `which` also was a pretty easy call.  I debated briefly regarding `rsync`, `git`, `ssh`, and `ssh-agent`.

lpsmith: Oh, I forgot `System.Directory` already has emulation for the `-p` option.

lpsmith: Done

PR #789, opened 2020-07-14
Use system temp dir for temporary thunk clone

alexfmpe:
Obelisk seems not to handle the ssh prompt correctly so when that kicks in I have to cancel `ob thunk unpack` to get the ssh-agent going, and end up with cruft in my local directory, hence this PR.

I have:

  - [x] (Optional) Run CI tests locally: `nix-build release.nix -A build.x86_64-linux --no-out-link` (or `x86_64-darwin` on macOS)

### DIFF
--- a/lib/command/src/Obelisk/Command/Thunk.hs
+++ b/lib/command/src/Obelisk/Command/Thunk.hs
@@ -782,8 +782,8 @@ unpackThunk' noTrail thunkDir = checkThunkDirectory thunkDir *> readThunk thunkD
   --TODO: Overwrite option that rechecks out thunk; force option to do so even if working directory is dirty
   Right ThunkData_Checkout -> failWith [i|Thunk at ${thunkDir} is already unpacked|]
   Right (ThunkData_Packed _ tptr) -> do
-    let (thunkParent, thunkName) = splitFileName thunkDir
-    withTempDirectory thunkParent thunkName $ \tmpThunk -> do
+    let thunkName = takeFileName thunkDir
+    withSystemTempDirectory thunkName $ \tmpThunk -> do
       let
         gitSrc = thunkSourceToGitSource $ _thunkPtr_source tptr
         newSpec = case _thunkPtr_source tptr of


lpsmith:
Actually this does have a problem:  I for one tend to put `/tmp` on `tmpfs`, (though I might also set up a user-specific `~/tmp` tmpfs and set `TMPDIR` accordingly)

As it turns out, this will cause `renameDirectory` to fail, as the temporary directory will be on a different block device/filesystem than the unpack location.

There's a similar-ish problem with #785.   We could say,  shell out to `mv`,  (or maybe emulate `mv` ourselves?  meh)

3noch:
Hm, shouldn't unpacking be bracketed so that if you kill it it deletes it's temporary dir?

lpsmith:
> Hm, shouldn't unpacking be bracketed so that if you kill it it deletes it's temporary dir?

You can't entirely rely on `bracket` to clean up temporary files:  `kill -9` will prevent the clean up code from running;  not sure about GHC's default `SIGTERM` signal handler.

But yes, we could at least make obelisk clean up after itself after a `SIGTERM`.

3noch:
Yeah I'm not worried about SIGKILL. If you have to do that something is very broken and you should not be surprised when cruft is left behind. But that should happen basically never.

alexfmpe:
Isn't it bracketed already? `withSystemTempDirectory` delegates to [withTempDirectory](https://hackage.haskell.org/package/temporary-1.3/docs/src/System.IO.Temp.html#withTempDirectory) which is a `bracket`

alexfmpe:
Oh, but that's what was there before and it was still leaving cruft around on ctrl-c. 
Maybe something we're doing interferes with the cleanup

3noch:
Yeah that's my concern. Why isn't bracket handling this for us??

alexfmpe:
I wonder if the git process doing the cloning ends up locking the directory while it's hanging on the ssh prompt. 
I'll try to test that when I have some time

mankyKitty:
@ali-abrar move to `nix-thunk` ?

ryantrinkle:
This should be handled with bracketing.  Also we can make the temp directory hidden if we aren't already.  It **should** be in the same directory as where it will end up so that it is on the same filesystem.

alexfmpe:
> This should be handled with bracketing. 


> Isn't it bracketed already? withSystemTempDirectory delegates to [withTempDirectory](https://hackage.haskell.org/package/temporary-1.3/docs/src/System.IO.Temp.html#withTempDirectory) which is a bracket

madeline-os:
The bracketing is broken and the ticket I made is supposed to be about fixing the bracketing. I'll make that more clear :sweat_smile:

PR #791, opened 2020-07-21
Disable nginx's access log for obelisk server

3noch:
Otherwise nginx will save access logs to disk which can fill up the disk

I have:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -137,6 +137,9 @@ in rec {
           locations.${baseUrl} = {
             proxyPass = "http://127.0.0.1:" + toString internalPort;
             proxyWebsockets = true;
+            extraConfig = ''
+              access_log stdout;
+            '';
           };
         };
       };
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -138,6 +138,9 @@ in rec {
             proxyPass = "http://127.0.0.1:" + toString internalPort;
             proxyWebsockets = true;
           };
+          extraConfig = ''
+            access_log stdout;
+          '';
```
3noch, line 143: Maybe should put this inside location block instead.

default.nix
```diff
@@ -137,6 +137,9 @@ in rec {
           locations.${baseUrl} = {
             proxyPass = "http://127.0.0.1:" + toString internalPort;
             proxyWebsockets = true;
+            extraConfig = ''
+              access_log stdout;
```
matthewbauer, line 141: not sure how much this will spam your syslog, but i think it will just be info severity so that's okay

matthewbauer: let's just set it to off, since obelisk has its own access log

3noch:
```suggestion
              access_log off;
```

Issue #790, opened 2020-07-19
error loading static css on ios


bxrt:
Following along the steps outlined in https://github.com/obsidiansystems/obelisk#ios I can build, deploy and run the scaffolding app on an iPhone with iOS 13.5.1.
However, the app does not load the static css that is linked. This is evident because the line "Here is a string defined in Common.Api" is not coloured red. Debugging the app while the phone is connected to a Mac shows "An error occurred trying to load the resource." for the css file in Safari's Web Inspector.

Similarly with my own app, an image placed in the static directory and linked in the html is not displayed in the app on iOS.

Xcode: 11.5
MacOS: 10.15.5

relevant output of `xcodebuild -showsdks`

```
2020-07-19 17:16:10.356 xcodebuild[71739:4196397] [MT] DVTSDK: Skipped SDK /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.2.sdk; its version (10.2) is below required minimum (13.5) for the iphoneos platform.
iOS SDKs:
	iOS 13.5                      	-sdk iphoneos13.5
[...]
```

Noteworthy: the deploy script places `[...]-obelisk.jpg` in directory: `/private/var/containers/Bundle/Application/[BA9B...B8C]/frontend.app/static`, whereas `[...]-main.css` is placed in `/static`
I have no clue if that is supposed to happen or not.

Edit: In "Noteworthy" above I assumed that the deploy script places the files in these directories, based on the error I saw in Safari's Web Inspector. After some thought, I now believe that the error means to say that Safari expects the files in those locations based on the html, but doesn't find the css file there.

3noch:
Thanks for the detailed report!

bxrt:
This is the generated HTML according to Safari's Web Inspector:
```
<!DOCTYPE html>
<html>
	<head>
		<base href="/">
		<title>Obelisk Minimal Example</title>
		<link href="static/1sjgs70sb2mpmx4k6608n42f0d1y5xxyaf7f4pl5hl961xzjzy7n-main.css" rel="stylesheet" type="text/css">
	</head>
	<body>
		<h1>Welcome to Obelisk!</h1>
		<p>Here is a string defined in Common.Api</p>
		<img src="static/0dspiyay7ldlnxdk7c7q3wjc9b2m4sy348piy8c6qq4dbs00d9x6-obelisk.jpg">
		<div>This string comes from config/common/example</div>
	</body>
[...]

PR #793, opened 2020-07-24
Frontend: fix loading css file by ios app

bxrt:
A \<base href="/"\> element in the html string makes the ios app
look for (some of) its assets in `/static/` instead of where they
are:
`/private/var/containers/Bundle/Application/[BA9B...B8C]/frontend.app/static/`

There should probably be no base element in the html in a WKWebView
since the baseURL is set separately in a WKWebView.

Resolves: #790

I have:

### DIFF
--- a/lib/frontend/src/Obelisk/Frontend.hs
+++ b/lib/frontend/src/Obelisk/Frontend.hs
@@ -211,9 +211,7 @@ runFrontendWithConfigsAndCurrentRoute mode configs validFullEncoder frontend = d
       w appendHead appendBody = do
         rec switchover <- runRouteViewT ve switchover (_frontendMode_adjustRoute mode) $ do
               (switchover'', fire) <- newTriggerEvent
-              mapRoutedT (mapSetRouteT (mapRouteToUrlT (appendHead . runConfigsT configs))) $ do
-                -- The order here is important - baseTag has to be before headWidget!
-                baseTag
+              mapRoutedT (mapSetRouteT (mapRouteToUrlT (appendHead . runConfigsT configs))) $
                 _frontend_head frontend
               mapRoutedT (mapSetRouteT (mapRouteToUrlT (appendBody . runConfigsT configs))) $ do
                 _frontend_body frontend


bxrt:
Remote CI build failed. I am not sure what I can do about that. Please help.

3noch:
Thank you! CI failure was spurious. I'm retrying it.

matthewbauer:
We have another baseTag at https://github.com/obsidiansystems/obelisk/blob/master/lib/frontend/src/Obelisk/Frontend.hs#L247. This is important because hydration requires the DOM to match.

I think the best solution here is to just disable it on wkwebview since we already have a base url set. Something like this:

```
baseTag :: forall route js t m. ObeliskWidget js t route m => RoutedT t route m ()
#ifdef APPLE_IOS
baseTag = blank
#else
baseTag = elAttr "base" ("href" =: "/") blank --TODO: Figure out the base URL from the routes
#endif
```

3noch:
We shouldn't have two *different* places that introduce the base tag though.

o1lo01ol1o:
Any update on this?

bxrt:
It turns out that changing
```baseTag = elAttr "base" ("href" =: "/") blank --TODO: Figure out the base URL from the routes```
to 
```baseTag = elAttr "base" ("href" =: "") blank --TODO: Figure out the base URL from the routes```
fixes the issue for ios and keeps the minimal obelisk application functional on the other platforms that I tried.
However, @matthewbauer mentions hydration. [I learned from Reddit](https://www.reddit.com/r/reflexfrp/comments/htonag/obeliskgeneratedstatic/fykcexf?utm_source=share&utm_medium=web2x&context=3) what hydration is, but I don't know how it is supposed to work and how much it depends on the value of the html base element.

I have also tried Matthew's proposal with the `#ifdef`, `#else` and `#endif`, but that didn't work. I assume because these are validated at compile time by the C-preprocessor and APPLE_IOS is not defined then.

bxrt:
@3noch, @cgibbard, is there anything I am expected to do for this fix at this point?

o1lo01ol1o:
ping

ali-abrar:
@bxrt Thanks for this. Sorry for the wait.

PR #801, opened 2020-08-23
Remove errors and warning for local packages without a library component

kmicklas:
Currently `ob shell` and derivative commands will print errors like this if you have a local package without a library component:

```
Haskell package has no library component
Failed to find buildable packages in $path_to_cabal_file
```

I don't see any reason why this should be considered an erroneous case.

This removes the `Failed to find buildable packages in ...` warning, but I don't believe it's necessary since any actual errors or warnings will still be reported by `parseCabalPackage`.

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Run.hs
+++ b/lib/command/src/Obelisk/Command/Run.hs
@@ -14,7 +14,7 @@ module Obelisk.Command.Run where
 import Control.Arrow ((&&&))
 import Control.Exception (Exception, bracket)
 import Control.Lens (ifor, (.~), (&))
-import Control.Monad (filterM, unless)
+import Control.Monad (filterM)
 import Control.Monad.Except (runExceptT, throwError)
 import Control.Monad.IO.Class (liftIO)
 import Control.Monad.Reader (MonadIO)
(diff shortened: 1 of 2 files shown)


kmicklas:
@3noch Anything blocking merging this?

lib/command/src/Obelisk/Command/Run.hs
```diff
@@ -386,9 +387,6 @@ parsePackagesOrFail dirs' = do
       "No valid, buildable packages found" <> (if null dirs then "" else " in " <> intercalate ", " dirs)
     Just xs -> pure xs
 
-  unless (null pkgDirErrs) $
-    putLog Warning $ T.pack $ "Failed to find buildable packages in " <> intercalate ", " pkgDirErrs
```
Ericson2314, line 390: Is this change related?

kmicklas: It's related in that you also get this warning for a package with no library component, in addition to the error above. I don't think either of them should be considered out of the ordinary.

Ericson2314: Ah and there is no more `Either`.

Ericson2314 (approved):

ali-abrar:
Thanks, Kenny

Issue #795, opened 2020-08-01
Unexpected implicit imports


rprije:
The `ob` command suite seems to implicitly load modules I don't ask for in my `.cabal` files.

For example, if I fail to add `jsaddle-dom` or `ghcjs-dom` to `frontend.cabal` and add this import to `frontend/src/Frontend.hs`:

```
import GHCJS.DOM.Document (getCookie, setCookie)
```

I get the following error in `ob run` and `ob repl`:

```
frontend/src/Frontend.hs:19:1: error:
    Ambiguous module name ‘GHCJS.DOM.Document’:
      it is bound as JSDOM.Generated.Document by a reexport in package jsaddle-dom-0.9.4.0
      it is bound as ghcjs-dom-0.9.4.0:GHCJS.DOM.Document by package ghcjs-dom-0.9.4.0
   |
19 | import           GHCJS.DOM.Document          (getCookie, setCookie)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, three modules loaded.
```

For comparison, if I go into a nix shell with `ob shell` and run `cabal new-repl frontend` I get the sort of error I'm expecting:

```
src/Frontend.hs:19:1: error:
    Could not load module ‘GHCJS.DOM.Document’
    It is a member of the hidden package ‘jsaddle-dom-0.9.4.0’.
    Perhaps you need to add ‘jsaddle-dom’ to the build-depends in your .cabal file.
    It is a member of the hidden package ‘ghcjs-dom-0.9.4.0’.
    Perhaps you need to add ‘ghcjs-dom’ to the build-depends in your .cabal file.
    Use -v to see a list of the files searched for.
   |
19 | import           GHCJS.DOM.Document          (getCookie, setCookie)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, no modules loaded.
```

In this case I seem to have no way of importing `GHCJS.DOM.Document` that will satisfy tools such as `ob run` due to the conflicting implicit imports.

Another problem is that in cases where obelisk loads non-conflicting packages my failure to add the needed package to my `.cabal` file is hidden from me by tools such as `ob run` and I won't learn of my mistake until I try to build my package and encounter failures.

3noch:
[...]

PR #804, opened 2020-09-08
Fix package deps

rprije:
Fixes obsidiansystems/obelisk#795

The problem in the above issue is that the `ghci` launched by the `ob` command suite by default exposes all packages in the project's GHC package database. The full list of packages can be seen by using `ghc-pkg list` in a nix shell within an initialised Obelisk project. This list contains both `jsaddle-dom` and `ghcjs-dom` giving rise to the error seen in obsidiansystems/obelisk#795 as well as other problems such as concealing a failure to include needed modules in a project's `.cabal` file.

`cabal new-repl` deals with this problem by passing `-hide-all-packages` to `ghci`. However, this means that all package dependencies must be explicitly exposed with `-package-id`  arguments which `cabal` takes care of.

This PR imitates `cabal`'s behaviour by adding `-hide-all-packages`, extracting build dependencies from the `.cabal` files, and adding them as `-package-id` arguments.

`ob` seems to overlap with `cabal` in a lot of its scope. I wonder if a better approach than further work akin to what I'm doing here might be to backport to `cabal` the things the `ob` command suite does better than `cabal`, such as the ability to load multiple modules, and move to using that instead?

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Run.hs
+++ b/lib/command/src/Obelisk/Command/Run.hs
@@ -46,13 +46,23 @@ import Debug.Trace (trace)
 import Distribution.Compiler (CompilerFlavor(..))
 import Distribution.PackageDescription.Parsec (parseGenericPackageDescription)
 import Distribution.Parsec.ParseResult (runParseResult)
+import Distribution.Pretty (prettyShow)
+import Distribution.Simple.Compiler (PackageDB (GlobalPackageDB))
+import Distribution.Simple.Configure (configCompilerEx, getInstalledPackages)
+import Distribution.Simple.PackageIndex (InstalledPackageIndex, lookupDependency)
+import Distribution.Simple.Program.Db (defaultProgramDb)
 import qualified Distribution.System as Dist
-import Distribution.Types.BuildInfo (buildable, cppOptions, defaultExtensions, defaultLanguage, hsSourceDirs, options)
+import Distribution.Types.BuildInfo (buildable, cppOptions, defaultExtensions, defaultLanguage, hsSourceDirs, options, targetBuildDepends)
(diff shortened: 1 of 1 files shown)


3noch:
Awesome!

rprije:
I tried this out against my personal project. It fixed the problem I was having with my `GHCJS.DOM.Document` module import and additionally notified me of some packages I'd forgotten to add to my `.cabal` file that were previously being concealed by this problem. I haven't noticed any new problems introduced. This PR is now good to review.

Ericson2314:
> `ob` seems to overlap with `cabal` in a lot of its scope. I wonder if a better approach than further work akin to what I'm doing here might be to backport to `cabal` the things the `ob` command suite does better than `cabal`, such as the ability to load multiple modules, and move to using that instead?

There was a GSOC this summer (which I helped mentor) for getting support for multiple components in GHC itself: https://summerofcode.withgoogle.com/projects/#5269390116782080. Folow-up work for Cabal and HIE has been started, after which I suppose we can drop what we have in `ob`.

rprije:
Updated this PR with a fix and rebased to the latest `develop` branch.

Details: After doing some updates of NixOS and Obelisk I discovered there were some problems with this PR. It was failing to load the GHC package DB when the system GHC was for a different version of GHC to the project package DB. This has now been fixed by cleaning up the code that handles loading the package db to be more consistent in getting the correct `ghc` and `ghc-pkg` programs and package db for the project from Nix.

rprije:
Is there anything I can do to help get this merged? I'd like to avoid resolving merge conflicts a 4th time.

ali-abrar:
Thanks! Apologies for the delay.

PR #807, opened 2020-09-18
WIP: Multiple domains

tomsmalley:
I have:


lib/route/src/Obelisk/Route.hs
```diff
@@ -526,6 +546,134 @@ checkEnum1EncoderFunc f = do
...
+      , ""
+      , "  Or for apps with multiple custom domains:"
+      , ""
+      , "MyDomain_A https://mydomain-a.com"
+      , "MyDomain_B https://mydomain-b.com"
+      , ""
+      , "  Or for local dev work, you can specify different ports:"
+      , ""
+      , "MyDomain_A http://localhost:8000"
+      , "MyDomain_B http://localhost:8001"
```
3noch, line 638: Does this support a variable number of domains (e.g. database driven domains)?

tomsmalley: No, that was specifically excluded for simplicity. It’s worth think about how we would extend this to support that though.

3noch: I probably haven't ingested what this is doing well enough to properly comment, but it seems like the domain doesn't need to be a separate type variable in the route type and instead could be reduced to nothing more than an encoder where the URL type is enriched to include this information.

3noch: In particular that seems like it would support variable number of domains by default.

tomsmalley: We do need to check that there is no overlap, which is currently done by universe.

3noch: I see. What's the benefit of adding the type parameter? I would imagine that an encoder could choose to map domains to any arbitrary part of the encoding.

ryantrinkle: I think it would be good if we can reduce the assumptions that Obelisk makes here.  For instance, what if someone wanted to just "decode" the domain and port using `id`, so that their app receives (Text, Word16) and can do whatever it likes with it?

tomsmalley: @3noch it's currently used so that the Universe instance for ObeliskRoute_Resource can generate routes for each domain, which is later used in overlap checks.

tomsmalley: @ryantrinkle I think that's still possible, but you'd need to use the DomainConfig to get that information. I'm not sure of a way to get rid of that...

ryantrinkle: @tomsmalley Can't ObeliskRoute_Resource just apply to all domains?  It doesn't seem like it should need to enumerate the domains, just have an appropriate Encoder.

tomsmalley: @ryantrinkle Ah, I thought you were against that based on one of our conversations, but I must've misunderstood.

lib/backend/src/Obelisk/Backend.hs
```diff
@@ -154,14 +171,52 @@ runSnapWithCommandLineArgs s = liftIO (commandLineConfig defaultConfig) >>= \c -
...
+
+requestPageName :: Request -> PageName
+requestPageName r = byteStringsToPageName p q
+  where
+    p = rqPathInfo r
+    q = rqQueryString r
+
+-- | Turn a request into a protocol relative domain
+requestDomain :: Request -> Domain
+requestDomain req = Domain $ "//" <> T.decodeUtf8 (rqHostName req)
```
alexfmpe, line 190: So very unsafe

lib/run/src/Obelisk/Run.hs
```diff
@@ -123,54 +189,62 @@ getConfigRoute configs = case Map.lookup "common/route" configs of
...
-  bracket
-    (bindPortTCPRetry settings (logPortBindErr port) (_runConfig_retryTimeout conf))
-    close
-    (\skt -> do
-        man <- newManager defaultManagerSettings
-        app <- obeliskApp configs defaultConnectionOptions frontend validFullEncoder uri $ fallbackProxy redirectHost redirectPort man
-        runner settings skt app)
+requestDomainWai :: W.Request -> Domain
+requestDomainWai req = Domain $ "//" <> hostName
+  where hostName = maybe "" T.decodeUtf8 $ W.requestHeaderHost req
```
alexfmpe, line 247: So very unsafe

PR #810, opened 2020-09-25
Put processed all.js files into an .assets directory

tomsmalley:
The consequence of this bug is all.js always 404s on full builds

I have:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -78,7 +78,7 @@ in rec {
     cd '${haskellLib.justStaticExecutables frontend}'
     shopt -s globstar
     for f in **/all.js; do
-      dir="$out/$(basename "$(dirname "$f")")"
+      dir="$out/$(basename "$(dirname "$f")").assets"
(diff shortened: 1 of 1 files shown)


3noch:
This bug was introduced by https://github.com/obsidiansystems/obelisk/pull/778

3noch (approved):
This fixes the bug.

ryantrinkle:
@tomsmalley Why does this work?

3noch:
Possibly a better change would be to have `serverExe` rename them all while copying/symlinking as that would *also* preserve certain properties of the behavior before #778, but I'm not sure those properties matter.

3noch:
I changed it do just that because `compressedJs` is a publicly accessible function in `default.nix`. If anyone is relying on the naming scheme of that derivation (as `serverExe` was doing) then it would break their code. As unlikely as that is, it's easy enough to preserve behavior.

Issue #762, opened 2020-06-10
Expose `ob shell` shell via nix


kmicklas:
`ob shell` creates a custom shell derivation which is less sensitive to changes than what you get from `default.nix`'s `shells.ghc`. It would be nice to expose this somehow via nix (either with IFD or by reimplementing it in nix) to be able to use it with editor integrations that use a nix shell, like [nix-buffer](https://github.com/shlevy/nix-buffer).

PR #817, opened 2020-10-28
Resurrect IDE support (VSCode + HLS)

srid:
With these two files in place, the skeleton works seemlessly on VSCode
with the Haskell plugin that uses haskell-language-server (the extension
downloads it as a static binary).

Just start vscode as `nix-shell --run "code ."`, or use the Nix Env Selector VScode extension (we can automate this by adding a [.vscode folder](https://github.com/srid/neuron/tree/master/.vscode)).

I have:

### DIFF
--- a/skeleton/hie.yaml
+++ b/skeleton/hie.yaml
@@ -1,0 +1,8 @@
+cradle:
+  cabal:
+    - path: "./frontend"
+      component: "lib:frontend"
+    - path: "./common"
+      component: "lib:common"
+    - path: "./backend"
+      component: "lib:backend"
--- a/skeleton/shell.nix
+++ b/skeleton/shell.nix
@@ -1,0 +1,1 @@
+(import ./. {}).shells.ghc


kmicklas:
The old attempt at this used `ob shell` to extract the ghci flags, which requires far fewer nix shell rebuilds than this way.

3noch:
At this point there is no official support for IDEs but if it works, great. Like @kmicklas said there are better ways to obtain a proper shell. And that also happens to be why I deleted `shell.nix` because there isn't a single correct answer to what it should contain. `ob shell` should be sufficient for the average case and you often want `nix-shell -A ghc.package` to work but that doesn't work if there is a `shell.nix` present.

srid:
The problem with `ob shell` is that it wouldn't work with [VScode remote](https://marketplace.visualstudio.com/items?itemName=arrterian.nix-env-selector) (ssh), unless we fork the [Nix Env Selector](https://marketplace.visualstudio.com/items?itemName=arrterian.nix-env-selector) extension and have it run `ob shell` manually.

3noch:
#762 would solve the issue of `ob shell` since it would become pure nix. Running `ob shell -- code .` should work no?

ryantrinkle:
@srid In general: we're happy to take PRs that add clean, general-purpose features, and if they help with IDE support, that's great. However, we don't have the resources right now to keep integration with a *specific* IDE working.  For instance, how would this PR get CIed?

srid:
@ryantrinkle For CI, you would run `nix-shell --run haskell-language-server` and expect it to succeed, after having put the static binary from [here](https://github.com/haskell/haskell-language-server/releases) in path (or you can get HLS from latest nixpkgs). It will process all Haskell source files, and fail if any of them fail to process.  

I'll close the PR though; as it looks like the goal here is to use `ob shell` rather than shells.ghc. If #762 gets resolved, I'll consider opening a PR with CI testing.

PR #819, opened 2020-11-06
Add catMaybesEncoder

kmicklas:
This is useful for working with, for example, optional query parameters.

I have:

### DIFF
--- a/lib/route/obelisk-route.cabal
+++ b/lib/route/obelisk-route.cabal
@@ -35,7 +35,8 @@ library
                  th-extras,
                  transformers,
                  universe,
-                 universe-dependent-sum
+                 universe-dependent-sum,
+                 witherable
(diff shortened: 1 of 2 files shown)


kmicklas:
On second thought, I don't think this is a valid encoder. It satisfies the algebraic round-trip law listed on `EncoderImpl` but `_encoderImpl_encode` isn't injective, which a (non-Haddock) comment on the field specifies.

ryantrinkle:
Yeah, I think in principle this can't roundtrip, because it is inherently losing the number (and spacing) of Nothing values.

Ericson2314:
Perhaps the opposite thing that does `catMaybes` for decode is a good idea? I would view that lossiness as analogous to not preserving white-space on round trip?

PR #824, opened 2020-11-28
Turn off hole fits to speed up compilation (take 2)

mightybyte:
Any occurrance of a type hole or undefined identifier that starts with
an underscore causes substantial slowdowns in compilation without this
GHC option.

I have:

### DIFF
--- a/skeleton/backend/backend.cabal
+++ b/skeleton/backend/backend.cabal
@@ -14,12 +14,12 @@ library
                , obelisk-route
   exposed-modules:
     Backend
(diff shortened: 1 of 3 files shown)


ryantrinkle (approved):

kmicklas:
Wouldn't it be better to put this in ob run, instead of the skeleton cabal files? That would automatically support previously created projects, and IMO Obelisk should avoid making too many choices about the GHCi experience if you're using plain cabal, since you can put whatever preferences you like in your `.ghci` then.

PR #835, opened 2020-12-31
Don't wait for nix show-derivation when deciding whether to provide UI feedback

ali-abrar:
Before this patch, we diffed the output of `nix show-derivation` to determine whether to attempt to rebuild the `staticFiles` derivation. Unfortunately, `show-derivation` often takes a significant amount of time and seems to do much of the same work involved in building the derivation. While we're waiting for the results of `show-derivation` to decide whether to embark upon and notify the user of an asset rebuild, the user gets no indication that we've noticed their change.

The tradeoff with this patch is that we'll be a little noisier: changes that don't actually impact the derivation but are within the project directory will cause no-op rebuilds for which the user will receive notifications. However, this is probably superior to a long, brooding silence while we determine whether to rebuild or not. To mitigate the problem, we ignore some common temp files that editors leave lying around, and limit our file-watching to the frontend, common, backend, and static folders.

I have:


ali-abrar:
Currently experimenting with using hnix to parse and evaluate the static asset derivation, with the hope of only watching the files that the evaluation cares about. That'll get rid of all this `ignoredFiles` stuff.

ali-abrar:
Related: https://github.com/haskell-nix/hnix/pull/804

ali-abrar:
> Related: [haskell-nix/hnix#804](https://github.com/haskell-nix/hnix/pull/804)
> 
> Once that PR is done, we can try to use that functionality to implement a more nuanced nix watcher.

Not doing this.

cgibbard (approved):

PR #843, opened 2021-03-06
Fix project shells

alexfmpe:
Search for dependencies inside `ob shell` as a tentative solution for fixing ob run/repl/watch.
I'm not really sure what the existing behavior is supposed to be, but `nixShellRunConfig` ends up failing on linux with
```
error: iOS builds are not supported on this platform.
ob: fd:11: hGetLine: end of file
```
and likewise on mac for Android. So does running `nix-shell` when there's only a `default.nix`. This is probably not a problem if one has remote builders setup for the other platform.

It seems like having a `shell.nix` that uses `shells.ghc` gets ob run/etc working *if* the project contains backend/frontend/common and no other packages. In particular, in obelisk/skeleton we get 
```
ob: Couldn't resolve dependency for logging-effect -any
CallStack (from HasCallStack):
  error, called at src/Obelisk/Command/Run.hs:514:14 in obelisk-command-0.9.0.1-GmAsx8iwxzhBQBzbMewBj8:Obelisk.Command.Run
```
`logging-effect` is an indirect dependency of (at least) `obelisk-command`

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command/Project.hs
+++ b/lib/command/src/Obelisk/Command/Project.hs
@@ -12,9 +12,9 @@ module Obelisk.Command.Project
   , initProject
   , nixShellRunConfig
   , nixShellRunProc
-  , obShellRunProc
   , nixShellWithHoogle
   , nixShellWithoutPkgs
+  , mkObNixShellProc
   , obeliskDirName
   , toImplDir
   , toObeliskDir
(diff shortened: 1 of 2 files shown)


danbornside:
lg2m

danbornside (approved):

ryantrinkle:
Is there a way we can invoke `ob shell` in haskell rather than by shelling out?  We're in the obelisk binary after all

alexfmpe:
Oh hmm yeah that makes more sense

PR #891, opened 2022-01-16
Add FieldHKD

alexfmpe:
EDIT: Moved explanation to haddocks

I have:

cc @endgame


lib/tabulation/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Revision history for tabulation
 
+* Add FieldHKD - a wrapper to allow mixing/reusing a non-HDK associated GADT with the higher kinded data record
```
endgame, line 3:
```suggestion
* Add FieldHKD - a wrapper to allow mixing/reusing a non-HKD associated GADT with the higher kinded data record
```

lib/tabulation/src/Data/Tabulation.hs
```diff
@@ -18,4 +19,8 @@ class HasFields a where
   traverseWithField t r = tabulateFieldsA (\f -> t f (indexField r f))
   indexField :: a -> Field a x -> x
   indexField a f = a ^. fieldLens f
-  {-# MINIMAL fieldLens, tabulateFieldsA #-}
\ No newline at end of file
+  {-# MINIMAL fieldLens, tabulateFieldsA #-}
+
+-- | A wrapper to allow mixing/reusing a non-HDK associated GADT with the higher kinded data record
```
endgame, line 24:
```suggestion
-- | A wrapper to allow mixing/reusing a non-HKD associated GADT with the higher kinded data record
```
Although seeing HDK in two places makes me think I might be wrong and this is actually a term I'm missing.

alexfmpe: Oops, typo and copy-pasta

Ericson2314:
This makes sense to me, but we better take the usage example (if not the examples of the alternatives) from the PR description and put in the haddocks first.

endgame:
Yeah, I think the usage example really needs to go into the haddocks.

(I still think rank2 `Representable` will likely be cleaner overall, but I don't really have a dog in this fight. Besides, AFAICT there's no package providing that class, let alone a canonical one, and I checked `rank2classes`, `conkin`, and `barbies`.)

alexfmpe:
While adding haddocks I realized since `FieldHKD` relates the associated GADTs, we can leverage it to go back and forth on the records via tabulate/index. Not sure how useful that is in pratice, but this seems like the module for it.

Let the name bike-shedding begin.

lib/tabulation/src/Data/Tabulation.hs
```diff
@@ -18,4 +19,134 @@ class HasFields a where
...
+unliftKind f r = 'tabulateFields' $ \field -> f ('indexField' r (FieldHKD field))
+
+xyPure :: 'Applicative' f => XY -> XYHKD f
+xyPure = 'liftKind' 'pure'
+
+xyRunIdentity :: XYHKD 'Identity' -> XY
+xyRunIdentity = 'unliftKind' 'runIdentity'
+@
+-}
+data FieldHKD field f x where
```
endgame, line 141: Can this be a newtype? If so, can lift/unliftkind be coercions?

cgibbard (approved):

lib/tabulation/src/Data/Tabulation.hs
Ericson2314: Yeah this has no existential so it should be fine as a newtype.

alexfmpe: Think I'm a bit lost here
1. how would I enforce the contents/index relationship if a newtype [doesn't allow GADTs](https://gitlab.haskell.org/ghc/ghc/-/issues/4009)? 
   ```haskell
   newtype FieldHKD field f (f x) = FieldHKD (field x)
   ```
2. liftKind/unliftKind don't convert between fields, they convert between the records, which are not newtypes of eachother. I can only imagine that being possible with [Unsaturated Type Families](https://github.com/ghc-proposals/ghc-proposals/pull/242) and something like
   ```haskell
   type family Id a where
     Id a = a

   data XYHKD f = XYHKD
     { _x' :: f ()
     , _y' :: f Bool
     }

   newtype XY = XY (XYHKD Id)
   ```

Ericson2314: Ah indeed
```haskell
newtype FieldHKD field f x where
  FieldHKD :: field x -> FieldHKD field f (f x)
```
doesn't quite work because there is a secret coersion for the `(f x)`

alexfmpe: Trying to use `GADTSyntax` with `newtype` confirms this isn't doable:

```• A newtype constructor cannot have existential type variables```

PR #892, opened 2022-01-21
Try removing hack for android

Ericson2314:
The cabal2nix patch we needed has long since landed.


madeline-os:
@Ericson2314 Does this build failure mean that the fix did not take, or did you remove too much overriding?

```
Setup: Encountered missing dependencies:
android-activity -any
```

madeline-os:
Ok, I think this needs to wait until we drop GHC 8.6 support.

Ericson2314:
That is what I recall too, because 8.6 ships with too old a Cabal version.

PR #901, opened 2022-02-20
Update FAQ.md

tdimiduk:
I'm not sure how many people have this problem, but this error message was pretty confusing for me until I realized what was going on, so it might be worth documenting.


ali-abrar (approved):

ali-abrar:
Thanks!

PR #903, opened 2022-03-10
Added externs file for Kadena Chainweaver

ohri-anurag:
I have:

  - [X] Followed the [contribution guide](https://github.com/obsidiansystems/obelisk/blob/develop/CONTRIBUTING.md#submitting-changes)
  - [X] Looked for lint in my changes with `hlint .` (lint found code you did not write can be left alone) - Did not change Haskell code.
  - [X] [Updated the changelog](https://github.com/obsidiansystems/obelisk/blob/develop/CONTRIBUTING.md#in-the-changelog)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,10 @@
 
 This project's release branch is `master`. This log is written from the perspective of the release branch: when changes hit `master`, they are considered released.
 
+## Unreleased
+* Added support for a static/externs.js file. This file allows users to specify global variables and namespaces in JS, that should not be used by the Google Closure Compiler
+during minification of the GHCJS produced JS.
+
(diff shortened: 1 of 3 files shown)


ohri-anurag:
Need to update the docs, will do it next week on Monday.

alexfmpe:
I wonder if we should have a sample externs.js with commented contents used in the skeleton so it's more discoverable. Then the FAQ could just link to it.

madeline-os (changes requested):
This looks good. I agree that we should add an example to the skeleton. Once we have that, I will approve :)

madeline-os (approved):
After some discussion, we have decided that the best course of action is to fix this with automation before we bless a solution in the skeleton.

PR #930, opened 2022-05-23
assets: add a warning to ob run when file doesn't exist

ali-abrar:
I have:


ChangeLog.md
```diff
@@ -21,6 +21,7 @@ This project's release branch is `master`. This log is written from the perspect
   * [#903](https://github.com/obsidiansystems/obelisk/pull/903): Added support for a file which allows users to specify global variables and namespaces in JS, that should not be used by the Google Closure Compiler during minification of the GHCJS produced JS. See the [FAQ](FAQ.md).
 * Static Assets
   * [#922](https://github.com/obsidiansystems/obelisk/pull/922): Serve .wasm files with the correct MIME type
+  * Add a warning to ob run when `static` is called with a path to a file that doesn't exist
```
alexfmpe, line 24: no longer a warning

madeline-os: fixed

lib/asset/manifest/src/Obelisk/Asset/TH.hs
```diff
@@ -31,7 +33,11 @@ staticPrefix :: FilePath
 staticPrefix = "/static"
 
 staticAssetRaw :: FilePath -> Q Exp
-staticAssetRaw fp = returnQ $ LitE $ StringL $ staticPrefix </> fp
+staticAssetRaw fp = do
+  exists <- runIO $ doesFileExist $ "static.out/" <> fp
```
madeline-os, line 37: I'm confused. What is `staticPrefix` and what is `static.out/` ?

madeline-os: This is sensitive to how the nix part of static assets works, right? Why is `static.out/` hardcoded?

ali-abrar: static.out is a symlink created in the project root directory. Whether you use a static asset derivation or a plain folder of static assets, it gets symlinked to static.out and `ob run` loads assets via that symlink. When the static asset derivation is rebuilt, the symlink is updated to the new nix store path.

The code for that behavior is [here](https://github.com/obsidiansystems/obelisk/blob/9127921ab7cd1e67f6869ee814491a5a7aafb698/lib/command/src/Obelisk/Command/Project.hs#L379-L403).

There's a note [here](https://github.com/obsidiansystems/obelisk/blob/develop/ChangeLog.md#v0920---2021-12-28) about static.out:

> * Feature: When staticFiles is a derivation, as opposed to a regular directory, produce a symlink to the result of that derivation at static.out and have ob run serve static assets from that symlink. This makes it possible for the static asset derivation to be rebuilt and the new results served without restarting ob run.
> * Feature: Rebuild static asset derivations while ob run is active as long as the change to the derivation is within the project folder. ob run now displays a message ("Static assets rebuilt and symlinked to static.out") whenever static assets have been rebuilt and the new static assets are being served.

As for `staticPrefix` vs `static.out`: the former is the URL prefix of static assets and the latter is the filepath prefix. `static` is generating a URL, so the `staticPrefix` gets appended. (Clearly this could all use better docs!)

madeline-os: inshallah, we will improve the docs

madeline-os (approved):

PR #948, opened 2022-07-20
Invoke shells in a more pure way; escape ghcid commands differently

madeline-os:
System.Process invokes `shell` commands using "/bin/sh". Uh-oh! That's
not what we want at all. So, instead, we manually invoke our shells
explicitly using "/usr/bin/env" so that the "sh" alias that nix-shell
installs is picked up.

However, we cannot control the executables we depend on, like ghcid,
which also use System.Process. Since that is the case, we need to fall
back to more generic, but less "surefire" escaping, that works for all
bourne-style shells.

I have:

  - [x] (Optional) Run CI tests locally: `nix-build release.nix -A build.x86_64-linux --no-out-link` (or `x86_64-darwin` on macOS)

### DIFF
--- a/lib/cliapp/src/Obelisk/CliApp/Process.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Process.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TupleSections #-}
 
 -- | An extension of `System.Process` that integrates with logging (`Obelisk.CLI.Logging`)
@@ -57,6 +58,7 @@ import qualified System.IO.Streams as Streams
 import System.IO.Streams.Concurrent (concurrentMerge)
 import System.Process (CreateProcess, ProcessHandle, StdStream (CreatePipe), std_err, std_out)
 import qualified System.Process as Process
+import System.Which (staticWhich)
 import Text.ShellEscape (bash, bytes)
 import qualified Data.Aeson as Aeson
 import Control.Monad.Log (Severity (..))
@@ -71,8 +73,17 @@ data ProcessSpec = ProcessSpec
 proc :: FilePath -> [String] -> ProcessSpec
 proc cmd args = ProcessSpec (Process.proc cmd args) Nothing
 
+-- | When we shell out, we want to know exactly which shell we're using.
+bashPath :: FilePath
+bashPath = $(staticWhich "bash")
+
+-- | We do not use Process.shell because it invokes "/bin/sh", which
+-- is a huge impurity for us. Unfortunately, we cannot guarantee that
+-- our dependent executables, such as ghcid, do the same.
+--
+-- Instead, we invoke `bash' specifically
 shell :: String -> ProcessSpec
-shell cmd = ProcessSpec (Process.shell cmd) Nothing
+shell cmd = ProcessSpec (Process.proc bashPath ["-c", cmd]) Nothing
 
 setEnvOverride :: (Map String String -> Map String String) -> ProcessSpec -> ProcessSpec
 setEnvOverride f p = p { _processSpec_overrideEnv = Just f }
--- a/lib/cliapp/src/Obelisk/CliApp/Process.hs
+++ b/lib/cliapp/src/Obelisk/CliApp/Process.hs
@@ -77,7 +77,7 @@ proc cmd args = ProcessSpec (Process.proc cmd args) Nothing
 bashPath :: FilePath
 bashPath = $(staticWhich "bash")
 
--- | We do not use Process.shell because it invokes "/bin/sh", which
+-- | We do not use Process.shell because it invokes @/bin/sh@, which
 -- is a huge impurity for us. Unfortunately, we cannot guarantee that
 -- our dependent executables, such as ghcid, do the same.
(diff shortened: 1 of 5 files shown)


lib/cliapp/src/Obelisk/CliApp/Process.hs
```diff
@@ -71,8 +71,13 @@ data ProcessSpec = ProcessSpec
...
 proc cmd args = ProcessSpec (Process.proc cmd args) Nothing
 
+-- | We do not use Process.shell because it invokes "/bin/sh", which
+-- is a huge impurity for us. Unfortunately, we cannot guarantee that
+-- our dependent executables, such as ghcid, do the same.
+--
+-- Instead, we invoke "/usr/bin/env sh -c <cmd>"
 shell :: String -> ProcessSpec
-shell cmd = ProcessSpec (Process.shell cmd) Nothing
+shell cmd = ProcessSpec (Process.proc "/usr/bin/env" ["sh", "-c", cmd]) Nothing
```
plt-amy, line 80: I don't understand this change. `/usr/bin/env sh` will be the first `sh` in the user's `$PATH`, which is just as of an unknown as whatever `/bin/sh` is — maybe even _more_ of an unknown. If we want this to be a shell _with guaranteed known behaviour_, shouldn't we [do something like this](https://github.com/obsidiansystems/cli-nix/pull/2/commits/3977c891720d9438c3050528cda7a96ec4227f44)?

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -312,9 +313,22 @@ nixShellRunConfig root isPure command = do
...
       ])
 
+-- | Escape using ANSI C-style quotes $''
+-- This does not work with all shells! Ideally, we would control exactly which shell is used,
+-- down to its sourced configuration, throughout the obelisk environment. At this time, this
+-- is not feasible.
 bashEscape :: String -> String
 bashEscape = BSU.toString . bytes . bash . BSU.fromString
 
+-- | Escape using bourne style shell escaping
```
plt-amy, line 323: Bourne is a proper noun

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -312,9 +313,22 @@ nixShellRunConfig root isPure command = do
...
+-- This is not as robust, but is necessary if we are passing to a shell we don't control.
```
plt-amy, line 324: How is it less robust?

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -312,9 +313,22 @@ nixShellRunConfig root isPure command = do
...
+-- The most prominent issue is that System.Process executes shell commands by invoking
```
plt-amy, line 325: `'System.Process'`

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -312,9 +313,22 @@ nixShellRunConfig root isPure command = do
...
+-- /bin/sh instead of /usr/bin/env sh. While we can avoid this by specifying a shell manually,
```
plt-amy, line 326: `@/bin/sh@`

lib/command/src/Obelisk/Command/Project.hs
plt-amy: Also: period to end the sentence

plt-amy (changes requested):

lib/cliapp/src/Obelisk/CliApp/Process.hs
madeline-os: :woman_facepalming: Yea

madeline-os: I tested it "on my machine" and thought it worked, but it didn't... because I have bash as my global sh.

lib/command/src/Obelisk/Command/Project.hs
madeline-os: Not every byte or control sequence can be escaped in that way. $'' can escape literally anything.

lib/command/src/Obelisk/Command/Project.hs
madeline-os: you get no punctuation but Bourne will get his name

lib/command/src/Obelisk/Command/Project.hs
cgibbard, line 325:
```suggestion
-- The most prominent issue is that 'System.Process' executes shell commands by invoking
```

lib/command/src/Obelisk/Command/Project.hs
cgibbard, line 326:
```suggestion
-- @/bin/sh@ instead of @/usr/bin/env sh@. While we can avoid this by specifying a shell manually,
```

lib/command/src/Obelisk/Command/Project.hs
```diff
@@ -312,9 +313,22 @@ nixShellRunConfig root isPure command = do
       , [cs]
       ])
 
+-- | Escape using ANSI C-style quotes $''
```
plt-amy, line 316: Does this Haddock compile? I'd assume you need `@$''@`

plt-amy (approved):
I concede on the punctuation because the bugfix looks good.

lib/command/src/Obelisk/Command/Project.hs
madeline-os: :fish: :fist_left: i'll check

PR #951, opened 2022-07-22
ob run: allow overriding the port

### LINKED ISSUE #950: Allow override for port while using ob-run
For most use cases, it suffices to specify the port in `config/common/route`, this is not the case in a situation where one wants to run a reverse proxy in front of `ob run`. In such a situation, the configured route will be for the externally accessible URI, and the port on which `ob run` listens is configured in the reverse proxy set up.

plt-amy:
Adds a switch `--port` to `ob run` to override, from the command line, the port specified in the route. Fixes #950. 

I have:

  - [X] Followed the [contribution guide](https://github.com/obsidiansystems/obelisk/blob/develop/CONTRIBUTING.md#submitting-changes)
  - [X] [Updated the changelog](https://github.com/obsidiansystems/obelisk/blob/develop/CONTRIBUTING.md#in-the-changelog)
  - [X] (Optional) Run CI tests locally: `nix-build release.nix -A build.x86_64-linux --no-out-link` (or `x86_64-darwin` on macOS)

### DIFF
--- a/lib/run/src/Obelisk/Run.hs
+++ b/lib/run/src/Obelisk/Run.hs
@@ -226,7 +226,6 @@ bindPortTCPRetry :: Settings
                  -> Int
                  -> IO Socket
 bindPortTCPRetry settings m n = catch (bindPortTCP (settingsPort settings) (settingsHost settings)) $ \(e :: IOError) -> do
-  print (settingsHost settings, settingsPort settings)
   m e
   threadDelay $ 1000000 * n
   bindPortTCPRetry settings (\_ -> pure ()) n
--- a/lib/run/src/Obelisk/Run.hs
+++ b/lib/run/src/Obelisk/Run.hs
@@ -22,7 +22,7 @@ import Prelude hiding ((.), id)
 import Control.Category
 import Control.Concurrent
 import Control.Exception
-import Control.Lens ((%~), (^?), _Just, _Right)
+import Control.Lens ((%~), (^?), (?~), _Just, _Right)
 import qualified Data.Attoparsec.ByteString.Char8 as A
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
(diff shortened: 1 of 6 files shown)


lib/run/src/Obelisk/Run.hs
```diff
@@ -130,11 +131,11 @@ runWidget
   -> IO ()
 runWidget conf configs frontend validFullEncoder = do
   uri <- either (fail . T.unpack) pure $ getConfigRoute configs
-  let port = fromIntegral $ fromMaybe 80 $ uri ^? uriAuthority . _Right . authPort . _Just
+  let port = fromMaybe (fromIntegral $ fromMaybe 80 $ uri ^? uriAuthority . _Right . authPort . _Just) (_runConfig_portOverride conf)
```
madeline-os, line 134: Can this be simplified to use like `<|>` or something? Lots of maybe stuff just being thrown around

lib/run/src/Obelisk/Run.hs
```diff
@@ -130,11 +131,11 @@ runWidget
...
       redirectHost = _runConfig_redirectHost conf
       redirectPort = _runConfig_redirectPort conf
       beforeMainLoop = do
-        putStrLn $ "Frontend running on " <> T.unpack (URI.render uri)
+        putStrLn $ "Frontend running on " ++ T.unpack (URI.render (uri & uriAuthority . _Right . authPort ?~ (fromInteger (fromIntegral port))))
```
madeline-os, line 138: It doesn't sit right with me that we have to edit both the log message and the actual logic to make them coherent together. I'm not sure what the solution is tho.

lib/run/src/Obelisk/Run.hs
```diff
@@ -79,13 +79,14 @@ import qualified System.Which
 
 run
   :: Int -- ^ Port to run the backend
+  -> Maybe Int
```
madeline-os, line 82: Haddock

lib/run/src/Obelisk/Run.hs
```diff
@@ -79,13 +79,14 @@ import qualified System.Which
...
   -> Maybe FilePath -- ^ Optional directory in which to find "cert.pem", "chain.pem" and "privkey.pem" to be used for TLS.
                     -- If this is Nothing and TLS is enabled, we'll generate a self-signed cert.
   -> ([Text] -> Snap ()) -- ^ Static asset handler
   -> Backend backendRoute frontendRoute -- ^ Backend
   -> Frontend (R frontendRoute) -- ^ Frontend
   -> IO ()
-run port certDir serveStaticAsset backend frontend = do
+run port frontendPort certDir serveStaticAsset backend frontend = do
```
madeline-os, line 89: I'm confused about why there's a port for the backend and a port for the frontend, now.

lib/run/src/Obelisk/Run.hs
```diff
@@ -79,13 +79,14 @@ import qualified System.Which
 
 run
```
madeline-os, line 80: Can we turn this function's args into a record? it's huge now.

madeline-os (changes requested):
Mostly just clarification requested.

madeline-os (approved):

PR #961, opened 2022-08-10
Add cabal.project file for local Haskell development

fendor:
I have:

This is purely an improvement for the Devs. It makes it much easier to use HLS on the code-base.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,25 @@
-# Contribution Guide
+# Contribution Guide <!-- omit in toc -->
 
 Contributions and issue reports are encouraged and appreciated!
 
 - [Opening Issues](#opening-issues)
 - [Submitting Changes](#submitting-changes)
   - [Guidelines for Commit Messages](#guidelines-for-commit-messages)
+    - [Summary Line](#summary-line)
+      - [Note on bumping dependencies](#note-on-bumping-dependencies)
+    - [Body](#body)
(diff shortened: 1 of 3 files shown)


lib/cabal.project
```diff
@@ -0,0 +1,15 @@
...
+    ./command
+    ./executable-config/inject
+    ./executable-config/lookup
+    ./frontend
+    ./route
+    ./run
+    ./selftest
+    ./snap-extras
+    ./tabulation
+
```
fendor, line 15: Adding something like:
```
package *
  ghc-options: -Wwarn
```
is necessary for newer GHC versions, since we have `-Werror` for unneeded imports, and cabal fails for that reason if you try to open, let's say 'selftest'.

madeline-os (approved):

lib/cabal.project
fendor: Added it to the cabal.project for good, since otherwise, it is not useful.

fendor:
Added some docs. 

I am not sure how to make sure this doesn't bitrot, can I add a test like `cabal repl all` somewhere? This would work good enough, imo.

CONTRIBUTING.md
```diff
@@ -1,14 +1,25 @@
-# Contribution Guide
+# Contribution Guide <!-- omit in toc -->
```
fendor, line 1: Otherwise, my VSCode markdown extension adds it to the TOC.

PR #967, opened 2022-08-15
Make --port compatible with TLS terminating reverse proxies

plt-amy:
Disables TLS handling on the server if the route is TLS **but** --port was given, to make it so that `ob run --port` can be used with reverse proxies that terminate TLS.

I have:


lib/run/src/Obelisk/Run.hs
```diff
@@ -191,26 +192,26 @@ runWidget toRun configs validFullEncoder = do
...
-      actualUri = uri & uriAuthority . _Right . authPort ?~ fromInteger (fromIntegral port)
 
       -- This is the server that will handle the backend requests. We
       -- support shuttling them off to any host:port pair.
       redirectHost = _runApp_backendHost toRun
       redirectPort = _runApp_backendPort toRun
 
       beforeMainLoop = do
-        putStrLn $ "Frontend running on " ++ T.unpack (URI.render actualUri)
+        putStrLn $ "Frontend running on http://localhost:" ++ show port ++ "/"
```
madeline-os, line 202: So before, this was logging the public-facing route, and now it's logging where `ob run` is listening on the local machine?

lib/run/src/Obelisk/Run.hs
```diff
@@ -191,26 +192,26 @@ runWidget toRun configs validFullEncoder = do
...
+
       settings = setBeforeMainLoop beforeMainLoop (setPort port (setTimeout 3600 defaultSettings))
 
-      -- Providing TLS here will also incidentally provide it to proxied requests to the backend.
-      prepareRunner = case uri ^? uriScheme . _Just . unRText of
-        Just "https" -> do
+      -- Providing TLS here will also incidentally provide it to proxied
+      -- requests to the backend.
+      prepareRunner =
+        if routeIsTLS && isNothing (_runApp_forceFrontendPort toRun) then
```
madeline-os, line 214: It's a little fraught that we are testing for TLS twice, here and in `beforeMainLoop`, once for logging and once for the actual logic. Is it easy to simplify that?

madeline-os (approved):

madeline-os:
Thanks @plt-amy !!

PR #978, opened 2022-08-30
Don't redefine mapSome

alexfmpe:
```haskell
import Obelisk.Route
import Data.Some
```

results in
```
    Ambiguous occurrence ‘mapSome’
    It could refer to either ‘Obelisk.Route.mapSome’,
                             imported from ‘Obelisk.Route’ at /path-to-obelisk/skeleton/frontend/src/Frontend.hs:16:1-20
                             (and originally defined
                                at /path-to-obelisk/lib/route/src/Obelisk/Route.hs:259:1-7)
                          or ‘Data.Some.mapSome’,
                             imported from ‘Data.Some’ at /path-to-obelisk/skeleton/frontend/src/Frontend.hs:24:1-16
                             (and originally defined in ‘some-1.0.2:Data.Some.Newtype’)
```

re-exporting seems strictly better than re-defining, though maybe it shouldn't be exported at all?

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -188,7 +188,7 @@ import Data.Maybe
 import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Semigroupoid
-import Data.Some (Some(Some))
+import Data.Some (Some(Some), mapSome)
 import Data.Tabulation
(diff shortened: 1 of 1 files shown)


madeline-os (approved):
Removing the re-export will be better, but for now I would prefer the non-breaking change.

PR #983, opened 2022-09-09
staticFileContent like embedFile

rubenmoor:
added function `staticFileContent`, allows embedding static assets as `ByteString` during compile time like `embedFile` does

I have:


fendor:
Hi! 
Changes look good to me, I am just wondering what's the benefit here over using `file-embed` directly?

rubenmoor:
> Hi! Changes look good to me, I am just wondering what's the benefit here over using `file-embed` directly?

I can't use `file-embed`. `embedFile :: FilePath -> Q Exp` requires a `FilePath` which is, in case of a static asset, the result of template haskell code, too. I couldn't combine `embedFile` with `staticFilePath` because of the GHC stage restriction and figured, I can simply re-implement `embedFile` for static assets like I did in this PR. 

I hope this makes sense. Maybe there is an alternative I overlooked.

ali-abrar:
You should be able to produce the filepath without template haskell. It's just `root </> fp`, I believe.

ali-abrar:
If you put a file in config/common or config/frontend, the contents are made available on the frontend. We do this by injecting the data as a base64-encoded  string into the head of the page. You can use the functions in `Obelisk.Configs` to retrieve those configs. The decoding is handled for you via this interface.

Would this functionality serve your purposes here?

cgibbard:
If I understand this change properly, it seems scary if the app has any large static files, no? If you were to drop, say, a 300MB video file into the static directory, emitting the entire contents of the video into your code at compile time might not work out too well. So if we want to do this, it should probably be controllable somehow. (But Ali's idea about using the existing config mechanism might also turn out to be good for your case here.)

rubenmoor:
> You should be able to produce the filepath without template haskell. It's just `root </> fp`, I believe.

There is the "asset pass through" mechanism, though. `root </> someFilePath` isn't aware of "static.out".

> If you put a file in config/common or config/frontend, the contents are made available on the frontend. We do this by injecting the data as a base64-encoded string into the head of the page. You can use the functions in `Obelisk.Configs` to retrieve those configs. The decoding is handled for you via this interface.
> 
> Would this functionality serve your purposes here?

 In principle, the config mechanism works. Only that my file isn't part of configuration in a meaningful way. I have a 770 kb file in json format that is the output of another haskell program, which I run prior to deployment. I.e. it contains expensively computed data that is then presented in the obelisk app.
 
Using the config mechanism just implies copying this file twice: to 'development-dir/config/frontend/' and to 'deployment-dir/config/frontend'. The internal base64 encoding and decoding is unnecessary overhead. (That is assuming that the 
 
 While feasible, I would consider this an abuse of the config mechanism.

rubenmoor:
> If I understand this change properly, it seems scary if the app has any large static files, no? If you were to drop, say, a 300MB video file into the static directory, emitting the entire contents of the video into your code at compile time might not work out too well. So if we want to do this, it should probably be controllable somehow. (But Ali's idea about using the existing config mechanism might also turn out to be good for your case here.)

It is just as scary as `embedFile` and given that, with obelisk and with this pull request, the file is shipped as part of the static file mechanism it is "controlled" like any other static file (images, videos, sounds, ...). You can check whether your app is too heavy on static files by checking the directory size. This is independent of how these files are embedded (in case of `embedFile`/this pull request that is as bytestrings).

PR #984, opened 2022-09-11
Add decoding utils

alexfmpe:
Original motivation was supporting some kind of canonicalization to allow each route to be decoded to from several urls. A lot of semantic agony later, this is what I ended up with.

I have:


lib/route/src/Obelisk/Route.hs
```diff
@@ -927,19 +930,45 @@ pageNameEncoder = bimap
...
+  -> Encoder check parse a b
+fallbackEncoder fallback = retryEncoder $ \e p -> p . fallback e
+
+-- | When parsing fails, try again with the provided function.
+retryEncoder
+  :: (Functor check, MonadError err parse)
+  => (err -> (b -> parse a) -> (b -> parse a))
+  -> Encoder check parse a b
+  -> Encoder check parse a b
+retryEncoder retry = unsafeWrapDecoding $ \p b -> catchError (p b) $ \e -> retry e p b
```
alexfmpe, line 959: Note this only satisfies `unsafeWrapDecoding`'s law if we can trust MonadError to preserve success. Which I wager we should, but the documentation has no laws whatsoever.
Currently being discussed in the mailing list: https://mail.haskell.org/pipermail/libraries/2022-September/031621.html

lib/route/src/Obelisk/Route.hs
```diff
@@ -927,19 +930,45 @@ pageNameEncoder = bimap
...
+
+-- | Completely override the decoding function.
+-- Since the only 'Encoder' law is round-tripping, values that fail to parse are unconstrained.
+-- Thus, to produce a law-abiding encoder, we need only preserve parse success.
+unsafeWrapDecoding
+  :: Functor check
+  => ((b -> parse a) -> (b -> parse' a)) -- ^ Required law for @wrap@ argument: @forall a b. p b = pure a => wrap p b = pure a@
```
alexfmpe, line 966: It seems `hoistParse` also implicitly relies on this law.
For instance
```haskell
hoistParse $ const $ throwError "bla"
```
will break roundtrips when we're encoding from inhabited types.
Should it be renamed to have an `unsafe` prefix?

ryantrinkle: Yes, with deprecation

lib/route/src/Obelisk/Route.hs
```diff
@@ -937,19 +940,45 @@ pageNameEncoder = bimap
...
+retryEncoder
+  :: (Functor check, MonadError err parse)
+  => (err -> (b -> parse a) -> (b -> parse a))
+  -> Encoder check parse a b
+  -> Encoder check parse a b
+retryEncoder retry = unsafeWrapDecoding $ \p b -> catchError (p b) $ \e -> retry e p b
+
+-- | Completely override the decoding function.
+-- Since the only 'Encoder' law is round-tripping, values that fail to parse are unconstrained.
+-- Thus, to produce a law-abiding encoder, we need only preserve parse success.
```
ryantrinkle, line 978: "preserve the *value* of successful parses", not just their successfulness

lib/route/src/Obelisk/Route.hs
ryantrinkle: Good enough

ryantrinkle (approved):

PR #987, opened 2022-09-24
Fix overlap logic in shadowEncoder

alexfmpe:
`shadowEncoder`'s overlap check is neither sound nor complete. To avoid the "n * m" performance of checking everything against everything, it iterates one axis only. The idea is that if there's a collision, decoding with the other encoder will give us the collision - but decoding to something alone does not mean that a collision exists!

Consider 

```haskell
data A = A deriving (Bounded, Enum, Eq, Ord, Show, Universe)
data B = B deriving (Bounded, Enum, Eq, Ord, Show, Universe)
data C = C1 | C2 deriving (Bounded, Enum, Eq, Ord, Show, Universe)

ac :: (MonadError Text check, MonadError Text parse) => Encoder check parse A C
ac = unsafeMkEncoder $ EncoderImpl
  { _encoderImpl_encode = \A -> C1
  , _encoderImpl_decode = \_ -> pure A
  }

bc :: (MonadError Text check, MonadError Text parse) => Encoder check parse B C
bc = enumEncoder $ \B -> C2
```

where the `A`s and `B`s do not overlap when encoding. This is what happens on develop:
```haskell
> ffor (checkEncoder $ shadowEncoder ac bc) $ \enc -> ffor [Left A, Right B] $ tryDecode enc . encode @(Either Text) enc
Right [Right (Left A),Right (Left A)]

> ffor (checkEncoder $ shadowEncoder bc ac) $ \enc -> ffor [Right A, Left B] $ tryDecode enc . encode @(Either Text) enc
Left "shadowEncoder: overlap detected: first encoder encodes B as C2, which second encoder decodes as A
```

In the first case the roundtrip law is broken as the decoding to `A` shadows the decoding to `B`.
In the second case we have a false positive for overlaps.

With this PR we instead get
```haskell
> ffor (checkEncoder $ shadowEncoder ac bc) $ \enc -> ffor [Left A, Right B] $ tryDecode enc . encode @(Either Text) enc  
Right [Right (Left A),Right (Right B)]

> ffor (checkEncoder $ shadowEncoder bc ac) $ \enc -> ffor [Right A, Left B] $ tryDecode enc . encode @(Either Text) enc
Right [Right (Right A),Right (Left B)]
```

Though I wonder whether `shadowEncoder` is a good name at all, since we can't allow shadowing because that breaks round-trips? Unless this "Overlaps" / "Gets overlapped" business is about unreachable encodings, for which we can do anything we want? See #988.

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -711,12 +711,12 @@ shadowEncoder f g = Encoder $ do
         Left a -> _encoderImpl_encode vf a
         Right b -> _encoderImpl_encode vg b
     , _encoderImpl_decode = \c ->
-        let mb = Right <$> _encoderImpl_decode vg c
-        in flip catchError (\_ -> mb) $ do
-          a <- _encoderImpl_decode vf c
-          case c == _encoderImpl_encode vf a of
-            False -> mb
-            True -> pure $ Left a
+        let ma = Left <$> _encoderImpl_decode vf c
+        in flip catchError (\_ -> ma) $ do
+          b <- _encoderImpl_decode vg c
+          case c == _encoderImpl_encode vg b of
+            False -> ma
+            True -> pure $ Right b
     }
 
 enum1Encoder
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -166,7 +166,7 @@ import Control.Monad.Trans (lift)
 import Data.Monoid ((<>))
 #endif
 #if __GLASGOW_HASKELL__ >= 906
-import Control.Monad (forM, (<=<))
+import Control.Monad (forM, guard, (<=<))
 import Control.Monad.Trans (lift)
 #endif
 #endif
@@ -532,6 +532,7 @@ maybeToEitherEncoder = unsafeMkEncoder $ EncoderImpl
 
 maybeEncoder
   :: ( MonadError Text check
+     , Eq b
      , Show a
      , Show b
      , check ~ parse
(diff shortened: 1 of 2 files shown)


cgibbard:
I think the encoder:

```
ac :: (MonadError Text check, MonadError Text parse) => Encoder check parse A C
ac = unsafeMkEncoder $ EncoderImpl
  { _encoderImpl_encode = \A -> C1
  , _encoderImpl_decode = \_ -> pure A
  }
```

is incorrect. _encoderImpl_decode should check that its argument is C1, and if not, fail. Had it done so, I think shadowEncoder would have done the right thing. Using unsafeMkEncoder directly is dangerous for reasons like this.

alexfmpe:
> _encoderImpl_decode should check that its argument is C1, and if not, fail

That would be the case if we required succesful decodes to roundtrip back when encoding and maybe we could, but the docs require only the opposite direction.
https://github.com/obsidiansystems/obelisk/blob/e885031efdd89169b41ef12190992794a3aa6ae6/lib/route/src/Obelisk/Route.hs#L329-L332

Wouldn't `handleEncoder` be invalid by the same reasoning given we could use it to derive the above `ac` in terms of the one you described with `handleEncoder (const A)` ?
https://github.com/obsidiansystems/obelisk/blob/e885031efdd89169b41ef12190992794a3aa6ae6/lib/route/src/Obelisk/Route.hs#L941-L952

alexfmpe:
Confirmed this fixes the issues caught by the test suite and uncommented those cases. 
This PR is the main reason I added the tests in the first place.

lib/route/src/Obelisk/Route.hs
```diff
@@ -692,7 +694,9 @@ shadowEncoder f g = Encoder $ do
   overlaps <- fmap catMaybes $ forM universe $ \a -> do
     let c = _encoderImpl_encode vf a
     mb <- gCanParse c
-    pure $ fmap (\b -> (a, b, c)) mb
+    pure $ mb >>= \b -> do
+      guard $ _encoderImpl_encode vg b == c
```
alexfmpe, line 703: This will stop working when #1075 is merged, but if I add `import Control.Monad` here the `-Werror` will cause this to fail on 8.10.

ryantrinkle: We could/should disable -Werror for redundant imports. Ideally it would be an error to have imports that are redundant for *all supported versions*, but I'm not sure to achieve that.

ryantrinkle:
Looks great!

Noting here from our discussion in matrix that maybe we want `a`'s noncannonical encodings to take precedence over `b`'s.

PR #988, opened 2022-09-24
Add unsafeSumEncoder

alexfmpe:
On top of #987 since I ran into it while figuring out this PR and need the same fix for `unsafeSumEncoder`.

`shadowEncoder` requires at least one of its arguments to be finite and small in practice, but that's not always possible. 
This PR allows callers to tailor their overlap checks to known structure of their encoders.
The original motivation is that I've ran into a situation myself where both summands have infinite domain but it's trivial to determine that they don't overlap since they differ on a finite prefix. So I ended up inlining the (current, bugged) `shadowEncoder` and hacking off the overlap check. That now becomes
```haskell
unsafeSumEncoder (\_ _ -> pure ()) -- I have discovered a truly remarkable proof of non-overlap which this comment is too small to contain
```

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -711,12 +711,12 @@ shadowEncoder f g = Encoder $ do
         Left a -> _encoderImpl_encode vf a
         Right b -> _encoderImpl_encode vg b
     , _encoderImpl_decode = \c ->
-        let mb = Right <$> _encoderImpl_decode vg c
-        in flip catchError (\_ -> mb) $ do
-          a <- _encoderImpl_decode vf c
-          case c == _encoderImpl_encode vf a of
-            False -> mb
-            True -> pure $ Left a
+        let ma = Left <$> _encoderImpl_decode vf c
+        in flip catchError (\_ -> ma) $ do
+          b <- _encoderImpl_decode vg c
+          case c == _encoderImpl_encode vg b of
+            False -> ma
+            True -> pure $ Right b
     }
 
 enum1Encoder
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -166,7 +166,7 @@ import Control.Monad.Trans (lift)
 import Data.Monoid ((<>))
 #endif
 #if __GLASGOW_HASKELL__ >= 906
-import Control.Monad (forM, (<=<))
+import Control.Monad (forM, guard, (<=<))
 import Control.Monad.Trans (lift)
 #endif
 #endif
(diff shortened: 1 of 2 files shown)


alexfmpe:
I went along with the `maybeEncoder` precedent for the name, but that's problematic since it suggests we're encoding *from* a sum, given the convention in place for, say, `dmapEncoder`. Maybe we should flip the names around to `maybeDecoder`, etc?

lib/route/src/Obelisk/Route.hs
```diff
@@ -527,14 +529,15 @@ maybeToEitherEncoder = unsafeMkEncoder $ EncoderImpl
...
+     , Eq b
      , Show a
      , Show b
      , check ~ parse
      )
   => Encoder check parse () b
   -> Encoder check parse a b
   -> Encoder check parse (Maybe a) b
-maybeEncoder f g = shadowEncoder f g . maybeToEitherEncoder
+maybeEncoder f g = unsafeSumEncoder exhaustiveOverlapCheck f g . maybeToEitherEncoder
```
ryantrinkle: Seems better to define shadow encoder in terms of these two then use it here so we don't have to use "unsafe"* here.

lib/route/src/Obelisk/Route.hs
```diff
@@ -674,36 +677,70 @@ consEncoder = unsafeMkEncoder $ EncoderImpl
...
+    , _encoderImpl_decode = \c ->
+        let mb = Right <$> _encoderImpl_decode vg c
+        in flip catchError (\_ -> mb) $ do
+          a <- _encoderImpl_decode vf c
+          case c == _encoderImpl_encode vf a of
+            False -> mb
+            True -> pure $ Left a
     }
 
+{-# DEPRECATED shadowEncoder "Use 'unsafeSumEncoder exhaustiveOverlapCheck' instead. This binding will be removed in a future release as it does no actual shadowing since that would break roundtrips." #-}
```
ryantrinkle: It shadows noncanonical encodings of b with noncanonical encodings of a, no?

ryantrinkle:
Adding `unsafeSumEncoder` looks good to me, but in your specific use case wouldn't it be better to encode the prefix and contents separately and then combine? And in decoding split the prefix off, decode it, then use that to select the decoder for the rest

lib/route/src/Obelisk/Route.hs
alexfmpe: Fair enough

alexfmpe:
> Adding unsafeSumEncoder looks good to me, but in your specific use case wouldn't it be better to encode the prefix and contents separately and then combine? And in decoding split the prefix off, decode it, then use that to select the decoder for the rest

Hmm probably yeah

PR #990, opened 2022-10-06
assets: staticFilePath should return the unhashed path

ali-abrar:
One of the recent changes to `staticFilePath` resulted in it returning a hashed filename instead of the actual filename. The hashed filename is used for URLs (for cache and cache-busting purposes) and obelisk-asset-serve-snap knows how to handle them. However, we don't need that hashed filename when looking things up on the local disk since we'll already be referring to an asset by its path in the nix store.

Prior to this change, this code: `Graphics.Text.TrueType.loadFontFile $(staticFilePath "fonts/inter/Inter-ExtraBold.ttf")` would work under `ob run` but fail when deployed to the server.

The error message is revealing: `/nix/store/mm2qvdw7j1lwzpbal6i84cxjxj1y716x-static/fonts/inter/0ig14m80pwsxv27k1nr8fjzd3a9dqqjj7fmryginih3mzswp692f-Inter-ExtraBold.ttf: openBinaryFile: does not exist (No such file or directory)`. Inspecting that folder in the nix store reveals (as expected) a file called Inter-ExtraBold.ttf but no version of that file with a hashed prefix.

I have:

### DIFF
--- a/lib/asset/manifest/src/Obelisk/Asset/TH.hs
+++ b/lib/asset/manifest/src/Obelisk/Asset/TH.hs
@@ -62,8 +62,10 @@ staticAssetFilePathRaw
 staticAssetFilePathRaw root = staticAssetWorker root staticOutPath
 
 staticAssetFilePath :: FilePath -> FilePath -> Q Exp
-staticAssetFilePath root fp = do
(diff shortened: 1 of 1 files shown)


ali-abrar:
This is an unreleased regression, hence no changelog.

madeline-os (approved):

PR #1000, opened 2022-11-12
Extract 'dsumEncoder' from chainEncoder

alexfmpe:
I wanted to call it `dependentEncoder` since it exhibits a bit of context dependent parsing, but then there's the dependent sum aspect and the name would get confusing. While a dependent sum encoder must be ctx dependent, the converse is not true. We can get a simpler
```haskell
  => Encoder check parse a a'
  -> (a -> Encoder Identity parse b b')
  -> Encoder check parse (a,b) (a', b')
```
by wrapping/unwraping via `Const`. I'm tempted to add such a util for convenience but am not coming up with a good name. Maybe `contextDependentEncoder` on account of it being the simpler such encoder?

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -124,6 +124,7 @@ module Obelisk.Route
   , pathSegmentEncoder
   , queryOnlyEncoder
   , Decoder(..)
+  , dsumEncoder
   , dmapEncoder
   , fieldMapEncoder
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):
Looks good to me!

ryantrinkle (approved):
This looks great; thanks!

alexfmpe:
Should I bundle a `contextDependentEncoder` with this?

PR #1036, opened 2023-06-06
Make rebooting a flag for "ob deploy push"

cidkidnix:
We might not always want to reboot the remote server, this should be the users decision to *actually* reboot

I have:

### DIFF
--- a/lib/command/src/Obelisk/Command.hs
+++ b/lib/command/src/Obelisk/Command.hs
@@ -2,9 +2,10 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
-{-# LANGUAGE PackageImports #-}
+{- ORMOLU_DISABLE -}
 module Obelisk.Command where
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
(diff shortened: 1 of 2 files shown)


lib/command/src/Obelisk/Command/Deploy.hs
```diff
@@ -258,7 +261,8 @@ nix-env -p /nix/var/nix/profiles/system --set "${bashEscape outPath}"
 /nix/var/nix/profiles/system/bin/switch-to-configuration boot
 booted="$(readlink /run/booted-system/{initrd,kernel,kernel-modules})"
 built="$(readlink /nix/var/nix/profiles/system/{initrd,kernel,kernel-modules})"
-if [ "$booted" = "$built" ]; then
+echo "${bashEscape (show reboot)}"
+if [[ "$booted" = "$built" && "${bashEscape (show reboot)}" == "${bashEscape (show False)}" ]]; then
```
tdimiduk, line 265: It seems to me like it might be better to just template in the reboot if the flag is set rather than doing logic in bash

lib/command/src/Obelisk/Command.hs
```diff
@@ -154,6 +155,9 @@ deployCommand cfg = hsubparser $ mconcat
       , command "ios" $ info ((,) <$> pure IOS <*> fmap pure (strArgument (metavar "TEAMID" <> help "Your Team ID - found in the Apple developer portal"))) mempty
       ]
 
+    reboot :: Parser Bool
+    reboot = flag False True (long "reboot")
```
mankyKitty, line 159: Rebooting on deploy is currently the default behaviour, this shouldn't change that. Otherwise users may be surprised to see that their running apps aren't the latest version following a deployment.

cidkidnix: I'd argue that it should've never been turned on by default in the first place, This should've always been the users choice to *actually* reboot

lib/command/src/Obelisk/Command.hs
```diff
@@ -2,9 +2,10 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
-{-# LANGUAGE PackageImports #-}
+{- ORMOLU_DISABLE -}
```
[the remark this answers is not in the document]
ali-abrar: Agree

lib/command/src/Obelisk/Command.hs
ali-abrar: You're talking about restarting the backend service, right?

mankyKitty: I was mistaken this is referring to the restart of the machine the backend service is running on. The backend service will still be restarted as part of a deploy that did not include the `--reboot` flag.

tdimiduk:
Any thoughts on whether this will be merged? I'm currently holding off on updating to latest obelisk and using a branch with this patch.

PR #989, opened 2022-09-26
Allow sharing query parameters at route branches

alexfmpe:
Edit: Rebased on top of #1079 

Allows turning 
```haskell
data ParentRoute :: * -> * where
  ParentRoute_ChildA :: ParentRoute ()
  ParentRoute_ChildB :: ParentRoute ()
  ParentRoute_ChildC :: ParentRoute ()

  FrontendRoute_Parent -> PathSegment "parent" $ pathComponentEncoder $ \case
    ParentRoute_ChildA -> PathEnd $ unitEncoderLenient mempty
    ParentRoute_ChildB -> PathSegment "b" $ pathOnlyEncoderIgnoringQuery . unitEncoder mempty
    ParentRoute_ChildC -> PathSegment "c" $ pathOnlyEncoderIgnoringQuery . unitEncoder mempty
```
into
```haskell
  FrontendRoute_Parent -> PathSegment "parent" $ pathComponentEncoderIgnoringQuery $ \case
    ParentRoute_ChildA -> PathEnd id
    ParentRoute_ChildB -> PathSegment "b" $ unitEncoder mempty
    ParentRoute_ChildC -> PathSegment "c" $ unitEncoder mempty
```

and

```haskell
  FrontendRoute_Parent :: FrontendRoute (R ParentRoute)

data ParentRoute :: * -> * where
  ParentRoute_ChildA :: ParentRoute (Map Text (Maybe Text))
  ParentRoute_ChildB :: ParentRoute (Map Text (Maybe Text))
  ParentRoute_ChildC :: ParentRoute (Map Text (Maybe Text))

  FrontendRoute_Parent -> PathSegment "parent" $ pathComponentEncoder $ \case
    ParentRoute_ChildA -> PathEnd id
    ParentRoute_ChildB -> PathSegment "b" queryOnlyEncoder
    ParentRoute_ChildC -> PathSegment "c" queryOnlyEncoder
```
into

```haskell
  FrontendRoute_Parent :: FrontendRoute (R ParentRoute, Map Text (Maybe Text))

  FrontendRoute_Parent -> PathSegment "parent" $ pathComponentEncoderSharingQuery id $ \case
    ParentRoute_ChildA -> PathEnd id
    ParentRoute_ChildB -> PathSegment "b" $ unitEncoder mempty
    ParentRoute_ChildC -> PathSegment "c" $ unitEncoder mempty
```

Motivation
[...]

PR #1079, opened 2024-05-15
Generalise SegmentResult

alexfmpe:
Ran into a non-urls usecase where I wanted to use `([Text], ByteString)` instead of `([Text], Map Text (Maybe Text))` and realized the groundwork for #989 allowed me to use `([Text], q)` as long as I relaxed the type signatures so I pulled that bit into this separate, more conservative PR.

The tutorial describes the `PageName` version of `pathComponentEncoder`. Should that be updated with this PR or do we keep it simple?

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route.hs
+++ b/lib/route/src/Obelisk/Route.hs
@@ -592,11 +592,11 @@ checkEnum1EncoderFunc f = do
 
 -- | This type is used by pathComponentEncoder to allow the user to indicate how to treat
 -- various cases when encoding a dependent sum of type `(R p)`.
-data SegmentResult check parse a =
-    PathEnd (Encoder check parse a (Map Text (Maybe Text)))
+data SegmentResult check parse a b =
+    PathEnd (Encoder check parse a b)
     -- ^ Indicate that the path is finished, with an Encoder that translates the
     -- corresponding value into query parameters
-  | PathSegment Text (Encoder check parse a PageName)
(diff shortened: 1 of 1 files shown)


lib/route/src/Obelisk/Route.hs
```diff
@@ -587,11 +587,11 @@ checkEnum1EncoderFunc f = do
 
 -- | This type is used by pathComponentEncoder to allow the user to indicate how to treat
 -- various cases when encoding a dependent sum of type `(R p)`.
-data SegmentResult check parse a =
-    PathEnd (Encoder check parse a (Map Text (Maybe Text)))
+data SegmentResult check parse a b =
```
alexfmpe, line 595: Breaking change - migration while keeping old meaning is to use `(Map Text (Maybe Text))` for b, but sometimes this can be generalized, as with `indexOnlyRouteSegment`. 

Do we want to do a proper warn-first-break-later migration cycle at the cost of polluting the namespace with a `SegmentResult'` or so for a few releases?

alexfmpe:
Ended up also locally defining a variant of `queryOnlyEncoder` with the more general type.

I wonder if this should be more of a generalise-query-type thing.

In theory we could do the same for path, so instead of `PageName` everywhere, unless url-specific we'd instead have `([fragment], query)` but I don't actually have a use-case for that so hard to tell what exactly to generalise to. What even is a generic path afterall?

At least allowing `()` locally as the query type seems pretty convenient and common

PR #1084, opened 2024-06-05
Remove Nix store path from source maps

isaac-uptrust:
Obelisk successfully serves minified JS, the corresponding unminified JS, and the source map that relates the two. But the source map's `sources` field contains the Nix store path of the unminified JS, not a URL. Browsers try to fetch the source code from a bogus URL:
`https://example.com/nix/store/HASH-compressedJs/frontend.jsexe/all.unminified.js`.

This change improves the situation by tweaking the arguments to `closure-compiler`; using relative paths to the JS files instead of absolute paths.

Works on Chromium, but Firefox still has trouble.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ This project's release branch is `master`. This log is written from the perspect
 
 * Obelisk.Route: add pathQueryEncoder and generalizeIdentity
 * [#1071](https://github.com/obsidiansystems/obelisk/pull/1071): Support deployment information repository sub-directories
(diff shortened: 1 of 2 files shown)


tdimiduk:
I just ran into this issue when i was profiling my code last week. Anything I can to do help get this merged?

alexfmpe:
> Works on Chromium, but Firefox still has trouble.

Does this mean Firefox is kept the same as before the PR or does something get worse? 
Do we know what's left for Firefox to work?
Do we know why there is a difference?

PR #1087, opened 2024-06-15
Run checkEncoder with template haskell

alexfmpe:
Got the idea of using TH for encoders long ago from @ryantrinkle. 

~This has the side effect of speeding up reloads by doing the check only when `Common.Route` is modified instead of every reload. Likely not to be noticed unless relying on large `Universe` instances.~ 
EDIT: Hmm no it doesn't. It's pretty unsatisfying that the lack of a `Lift` instance (at least when using `->` and `Kleisli m` categories) means we need to check twice. All we get is upgrading fails-at-startup to fails-at-compile-time.

I marked this as draft because everytime `Common.Route.Checked` is recompiled in ghci (and thus `ob` commands) I start getting
```
<interactive>: ^^ Could not load 'CommonziRouteziChecked_validFullEncoder_closure', dependency unresolved. See top entry above.


ByteCodeLink.lookupCE
During interactive linking, GHCi couldn't find the following symbol:
  CommonziRouteziChecked_validFullEncoder_closure
This may be due to you not asking GHCi to load extra object files,
archives or DLLs needed by your current session.  Restart GHCi, specifying
the missing library using the -L/path/to/object/dir and -lmissinglibname
flags, or simply by naming the relevant files on the GHCi command line.
Alternatively, this link failure might indicate a bug in GHCi.
If you suspect the latter, please report this as a GHC bug:
  https://www.haskell.org/ghc/reportabug


...done
```
until I recompile `Backend.hs` again. Not using `validFullEncoder` in `Backend.hs` also stops it from triggering

I have:

### DIFF
--- a/skeleton/backend/src/Backend.hs
+++ b/skeleton/backend/src/Backend.hs
@@ -1,10 +1,13 @@
 module Backend where
 
 import Common.Route
+import Common.Route.Checked
+import Data.Functor.Identity
 import Obelisk.Backend
+import Obelisk.Route
 
 backend :: Backend BackendRoute FrontendRoute
 backend = Backend
   { _backend_run = \serve -> serve $ const $ return ()
-  , _backend_routeEncoder = fullRouteEncoder
+  , _backend_routeEncoder = hoistCheck (pure . runIdentity) validFullEncoder
   }
--- a/skeleton/common/src/Common/Route/Checked.hs
+++ b/skeleton/common/src/Common/Route/Checked.hs
@@ -1,0 +1,20 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeApplications #-}
+
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
+module Common.Route.Checked where
+
+import Data.Functor.Identity
+import qualified Data.Text as Text
+import Obelisk.Route
+
+import Common.Route
+
+case fullRouteEncoder of
+  Left err -> error $ Text.unpack err
+  Right _ ->
+    [d|
+      validFullEncoder :: Encoder Identity Identity (R (FullRoute BackendRoute FrontendRoute)) PageName
+      Right validFullEncoder = fullRouteEncoder
+    |]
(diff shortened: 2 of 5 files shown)


skeleton/backend/src/Backend.hs
```diff
@@ -1,10 +1,13 @@
...
+import Common.Route.Checked
+import Data.Functor.Identity
 import Obelisk.Backend
+import Obelisk.Route
 
 backend :: Backend BackendRoute FrontendRoute
 backend = Backend
   { _backend_run = \serve -> serve $ const $ return ()
-  , _backend_routeEncoder = fullRouteEncoder
+  , _backend_routeEncoder = hoistCheck (pure . runIdentity) validFullEncoder
```
alexfmpe, line 12: This could just be `_backend_routeEncoder = validFullEncoder` but that's a breaking change, probably best to get adoption first (assuming we merge this)

skeleton/common/src/Common/Route/Checked.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Common.Route
+
+case fullRouteEncoder of
+  Left err -> error $ Text.unpack err
+  Right _ ->
+    [d|
+      validFullEncoder :: Encoder Identity Identity (R (FullRoute BackendRoute FrontendRoute)) PageName
+      Right validFullEncoder = fullRouteEncoder
+    |]
```
alexfmpe, line 20: Can't do this directly in `Common.Route` due to stage restriction.
I would also like to have some sort of ```thisIsTheOnlyBindingYouShouldEdit = fullRouteEncoder``` before the `case` so that renames to `fullRouteEncoder` in "userland" only require a single rename in this weird place (and thus unlikely to get out of sync) but that requires yet another module due to stage restriction.

alexfmpe: Wasn't able to pull this into a DRY util as there's no way to get a `Lift` instance for `Encoder` to splice back in.
At most we could have a util taking in `Encoder check parse a b` for the check and `Name`/`Dec` for use in the new binding and pass both "forms" of `fullRouteEncoder`. Maybe it'd be worth putting that in `Obelisk.Route` ?

PR #1092, opened 2024-08-13
Less `<nixpkgs>`

Ericson2314:
I have:


default.nix
```diff
@@ -13,8 +13,6 @@ let
   inherit (reflex-platform) hackGet nixpkgs;
   pkgs = nixpkgs;
 
-  inherit (import dep/gitignore.nix { inherit (nixpkgs) lib; }) gitignoreSource;
-
```
Ericson2314: dead code, I think? Unless this is for downstream consumers.

cgibbard: If this isn't needed, do we still need the thunk? Is it imported elsewhere?

Ericson2314: Yes, it is imported elsewhere (can't find in this diff, but I saw it via grepping for `gitignoreSource`)

Ericson2314: I put it back for now.

PR #1099, opened 2024-10-01
Add `pairRoute`

Ericson2314:
`subPairRoute` / `subPairRoute_` promoted a bad style, this function should be used instead.

I have:

### DIFF
--- a/lib/route/src/Obelisk/Route/Frontend.hs
+++ b/lib/route/src/Obelisk/Route/Frontend.hs
@@ -31,6 +31,7 @@ module Obelisk.Route.Frontend
   , mapRoutedT
   , subRoute
   , subRoute_
+  , pairRoute
   , subPairRoute
   , subPairRoute_
   , maybeRoute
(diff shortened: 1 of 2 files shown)


lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -216,14 +217,36 @@ subRoute_ :: (MonadFix m, MonadHold t m, GEq r, Adjustable t m) => (forall a. r
 subRoute_ f = factorRouted $ strictDynWidget_ $ \(c :=> r') -> do
   runRoutedT (f c) r'
 
--- | Like 'subRoute_', but with a pair rather than an R
-subPairRoute_ :: (MonadFix m, MonadHold t m, Eq a, Adjustable t m) => (a -> RoutedT t b m ()) -> RoutedT t (a, b) m ()
-subPairRoute_ f = withRoutedT (fmap (\(a, b) -> Const2 a :/ b)) $ subRoute_ (\(Const2 a) -> f a)
-
```
Ericson2314, line 222: Just moved this below, so the `subPairRoute*` are together (and the deprecation notice is easier to follow).

lib/route/src/Obelisk/Route/Frontend.hs
```diff
@@ -216,14 +217,31 @@ subRoute_ :: (MonadFix m, MonadHold t m, GEq r, Adjustable t m) => (forall a. r
...
+-- routes over the right side.
+pairRoute
+  :: (Monad m, Reflex t)
+  => (Dynamic t rl -> RoutedT t rr m a)
+  -> RoutedT t (rl, rr) m a
+pairRoute f = do
+  r <- askRoute
+  withRoutedT
+    (fmap snd)
+    (f $ fmap fst r)
```
alexfmpe, line 234: It's a bit odd that both these sub-dynamics trigger at the same time when we're treating this as a simplified `a :/ b`. 

Since we're "peeling" off the left bit first, one could expect a hierarchical relationship, that is, for
```pairRoute $ \leftDyn -> do
  dyn_ $ ffor leftDyn leftWidget
  rightDyn <- askRoute
  dyn_ $ ffor rightDyn rightWidget
```
to act like `subRoute`, and for `leftWidget` to survive occurences of `rightDyn`. Or possibly also vice-versa if not thinking of it as hierarchical but as a RoutedT equivalent of `factorDyn` for products

Ericson2314: @alexfmpe and I talked about the bit. It is true the lack of a `holdUniqDyn` making the sides change independently will make this not work as well as one might expect.

(Of course the fact that we need to `holdUniqDyn` and `factorDyn`, vs starting with the nested structure, is a bit suspicious, but that's a separate problem for a separate day.)

PR #1102, opened 2024-11-18
Prevent accidentally getting dependencies with cabal

alexfmpe:
This make sure all dependencies are provided by nix and `cabal` can't won't download any, preventing one from forgetting to re-launch the shell after adding a not-currently-available package.

For instance, when doing `cabal build all` inside `nix-shell -A shells.ghcjs` 

I have:


ymeister:
I have concerns regarding how this will play out with `haskell.nix` branch which is expected to become default somewhere in the future. `haskell.nix` uses `cabal.project` and `*.cabal` files to get the build plan, so setting `active-repositories` might interfere with that (though it needs to be checked).

PR #1108, opened 2024-12-19
Fix mobile builds

alexfmpe:
I have:

### DIFF
--- a/all-builds.nix
+++ b/all-builds.nix
@@ -41,6 +41,19 @@ let
     else if lib.isList v then lib.concatMap collect v
     else [];
 
+  # A simple derivation that just creates a file with the names of all
+  # of its inputs. If built, it will have a runtime dependency on all
+  # of the given build inputs.
+  pinBuildInputs = name: buildInputs: (local-self.nixpkgs.releaseTools.aggregate {
+    inherit name;
+    constituents = buildInputs;
+  }).overrideAttrs (old: {
+    buildCommand = old.buildCommand + ''
+      echo "$propagatedBuildInputs $buildInputs $nativeBuildInputs $propagatedNativeBuildInputs" > "$out/deps"
+    '';
+    inherit buildInputs;
+  });
+
   perPlatform = lib.genAttrs cacheBuildSystems (system: let
     reflex-platform = import ./dep/mars { inherit system __useNewerCompiler; };
 
@@ -90,7 +103,7 @@ let
         inherit iosSkeleton;
       };
     in packages // {
-      cache = reflex-platform.pinBuildInputs
+      cache = pinBuildInputs
         "obelisk-${system}-${nameSuffix}"
         (collect packages);
     };
@@ -100,12 +113,12 @@ let
       unprofiled = mkPerProfiling false;
     };
   in perProfiling // {
-    cache = reflex-platform.pinBuildInputs
+    cache = pinBuildInputs
       "obelisk-${system}"
       (map (p: p.cache) (builtins.attrValues perProfiling));
   });
 
-  metaCache = local-self.reflex-platform.pinBuildInputs
+  metaCache = pinBuildInputs
     "obelisk-everywhere"
     (map (a: a.cache) (builtins.attrValues perPlatform));
 
--- a/all-builds.nix
+++ b/all-builds.nix
@@ -4,7 +4,6 @@
   }
 , local-self ? import ./. self-args
 , supportedSystems ? [ builtins.currentSystem ]
-, __useNewerCompiler  ? true # false if one wants to use ghc 8.6.5
 }:
 
 let
@@ -55,7 +54,7 @@ let
   });
 
   perPlatform = lib.genAttrs cacheBuildSystems (system: let
-    reflex-platform = import ./dep/mars { inherit system __useNewerCompiler; };
+    reflex-platform = import ./dep/mars { inherit system; };
 
     mkPerProfiling = profiling: let
       obelisk = import ./. (self-args // { inherit system profiling; });
--- a/all-builds.nix
+++ b/all-builds.nix
@@ -81,8 +81,6 @@ let
         echo "return" >> "$out"
         cat "${skeleton.shells.ghc}" >> "$out"
       '';
-      androidSkeleton = skeleton.android.frontend;
-      iosSkeleton = skeleton.ios.frontend;
       nameSuffix = if profiling then "profiled" else "unprofiled";
       packages = {
         skeletonProfiledObRun = rawSkeleton.__unstable__.profiledObRun;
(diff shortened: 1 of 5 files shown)


skeleton/default.nix
```diff
@@ -15,14 +15,38 @@
...
+  web = {
+    executable = ps: exes: (exes (ps.frontend)).frontend;
+  };
   extraArgs = {
     staticFiles = ./static;
   };
   src = ./.;
   inputThunks = pkgs.obsidianCompilers.thunkSets.common ++ [
     pkgs._dep.source.aeson-1541
+    pkgs._dep.source.android-activity
```
alexfmpe, line 49: Version on hackage is too old and a release wouldn't fix it until we bump hackage index

all-builds.nix
```diff
@@ -41,6 +41,19 @@ let
     else if lib.isList v then lib.concatMap collect v
     else [];
 
+  # A simple derivation that just creates a file with the names of all
+  # of its inputs. If built, it will have a runtime dependency on all
+  # of the given build inputs.
+  pinBuildInputs = name: buildInputs: (local-self.nixpkgs.releaseTools.aggregate {
```
alexfmpe, line 46: This was removed from reflex-platform. Not sure if intentional

all-builds.nix
```diff
@@ -55,7 +54,7 @@ let
   });
 
   perPlatform = lib.genAttrs cacheBuildSystems (system: let
-    reflex-platform = import ./dep/mars { inherit system __useNewerCompiler; };
```
alexfmpe, line 45: Gives me eval errors

all-builds.nix
```diff
@@ -96,10 +94,10 @@ let
           ghcjs
           serverSkeletonExe
           ;
-      } // lib.optionalAttrs reflex-platform.androidSupport {
-        inherit androidSkeleton;
-      } // lib.optionalAttrs reflex-platform.iosSupport {
-        inherit iosSkeleton;
+      } // lib.optionalAttrs (system == "x86_64-linux") {
+        android-app = rawSkeleton.android.app.aarch64;
+      } // lib.optionalAttrs (system == "x86_64-darwin") {
```
alexfmpe, line 99: `androidSupport/iosSupport` are gone:
 https://github.com/reflex-frp/reflex-platform/blob/20a0a01ec32a747106cb22605832cb57f6fb62ca/default.nix#L106-L107

For now I mirrored what's in 
https://github.com/reflex-frp/reflex-platform/blob/20a0a01ec32a747106cb22605832cb57f6fb62ca/release.nix#L15-L20 
but it's strange that the logic for whether a platform is supported be pushed on users

dep/mars/github.json
```diff
@@ -1,8 +1,8 @@
 {
   "owner": "reflex-frp",
   "repo": "reflex-platform",
-  "branch": "haskell.nix",
+  "branch": "haskell.nix-platforms",
   "private": false,
-  "rev": "d5f1a1eb2ab91faa2fdb162085c40e417b0e5afe",
-  "sha256": "1vb69xjlb0c0f0hbrvbhw9na392sl7nn464d3kinpbjdyd9kdw6n"
+  "rev": "4e3bf11e34a4958d8850b1ac60e654fe12eac59b",
```
alexfmpe: https://github.com/reflex-frp/reflex-platform/pull/843

skeleton/default.nix
```diff
@@ -15,23 +15,72 @@
...
+          cp -r obelisk-skeleton $out
+          cp $out/bin/obelisk-skeleton $out/obelisk-skeleton.app
+        '';
+      };
+      packages.jsaddle-wkwebview.src = (thunkSource ../dep/jsaddle) + "/jsaddle-wkwebview";
+      packages.jsaddle-wkwebview.components.library = {
+        frameworks =
+          if (pkgs.stdenv.targetPlatform.isiOS) then lib.mkForce [ pkgs.darwin.iosSdkPkgs.sdk pkgs.darwin.apple_sdk.frameworks.CoreFoundation ]
+          else [ pkgs.darwin.apple_sdk.frameworks.CoreFoundation ];
+      };
```
alexfmpe, line 83: Was getting webkit2gtk eval failures on android so cargo-culted this section from https://github.com/reflex-frp/reflex-platform/blob/4fccfb196ac724695a5006a456b2125eeff331e5/example/default.nix#L62

We definitely don't want this re-done on every project. 
Something more like, shoved into a trivially importable overlay on reflex-platform?
Could maybe have it implicitly applied by obelisk, but then what happens on module conflicts?

skeleton/default.nix
```diff
@@ -15,23 +15,72 @@
   }
 }:
 with obelisk;
-project {} ({ pkgs, ... }: {
+let args = {
+  # You must accept the Android Software Development Kit License Agreement at
+  # https://developer.android.com/studio/terms in order to build Android apps.
+  # Uncomment this to indicate your acceptance:
+  android_sdk_accept_license = true;
+  allowUnfree = true;
```
alexfmpe, line 23: Should be set to false before merging. 
How do we make sure CI caches it though? 
Add the accept/allow as args to this file defaulting to false and overriden by CI?

skeleton/default.nix
cidkidnix: I think the linked code should *always* be applied in obelisk (not so much in reflex-platform). If folks need to overwrite it they can use "lib.mkOverride" for a higher value than "lib.mkForce". I'd much rather this be a *default* instead of some documentation or comments anywhere

all-builds.nix
cidkidnix: hm the `ghcShell` failure on darwin is interesting, I think we'll be okay disabling it (for now at least) but maybe we should be disabling webkit2gtk for darwin..

all-builds.nix
cidkidnix: If you could post those errors that'd be great!

all-builds.nix
cidkidnix: Yeah, I wasn't needing them while doing the haskell.nix work. I don't think we need them in reflex-platform anymore but pulling it into the obelisk nix is fine by me

skeleton/default.nix
cidkidnix: makes sense!

skeleton/default.nix
alexfmpe: Well yes but a default on obelisk or on the obelisk skeleton?

all-builds.nix
alexfmpe: I mean, the flag is gone on the haskell.nix branch. 
Compare
https://github.com/reflex-frp/reflex-platform/blob/2e0d31bc1b565eed02a3848cde0e0dbc4f33a2d4/default.nix#L12
with
https://github.com/reflex-frp/reflex-platform/blob/4fccfb196ac724695a5006a456b2125eeff331e5/default.nix#L2

PR #1114, opened 2025-02-27
Document need for nix-command in setup

augyg:
**only changes documentation

Obelisk uses nix eval in a number of places which will fail and be unknown to the user if they don't have the -v option.

In any case, it would be easiest to notify users of the need for nix-command in order for Obelisk to function as expected.

I have:


alexfmpe:
Makes sense, indentation seems off though?

Ericson2314:
I suppose this works for now, but really, the `ob` command should not be using unstable features.

augyg:
Sorry, I dont know why the indentation was all weird but should be fixed now

alexfmpe (approved):

